PR #32, opened 2024-03-19
Missing pieces from #26 and closes #29

### LINKED ISSUE #29: Remove #! from non-executable files
I see that a number of the `.py` files start with  `#!/usr/bin/env/python`, as in:

https://github.com/CDCgov/multisignal-epi-inference/blob/main/model/src/pyrenew/basic.py#L1

These are only needed when you will be executing the file directly, because their purpose is to tell the shell which interpreter to run. I.e., so you can do `./foo.py` rather than `python foo.py`. However, if the python code is just a module that isn't intended to be run but just imported, then you don't need this.

In addition, this is actually the wrong line, because what's after the `#!` needs to be something executable, but this isn't usually a valid path, because `/usr/bin/env` is a program, not a directory. Rather, you want `/usr/bin/env python`, which runs the `env` command to run `python`.

gvegayoncdc:
- Removing `#!` from files.
- Renaming RandomProcess to RandomVariable.
- Renaming modules to singular not plural.


model/src/pyrenew/convolve.py
```diff
@@ -1,6 +1,3 @@
-#!/usr/bin/env/python
-# -*- coding: utf-8 -*-
```
dylanhmorris, line 2:
```suggestion
# -*- coding: utf-8 -*-
```
Can keep encoding notes; just delete shebangs

model/src/pyrenew/model/hospitalizations.py
```diff
@@ -1,11 +1,8 @@
-#!/usr/bin/env/python
-# -*- coding: utf-8 -*-
-
 from collections import namedtuple
 
-from pyrenew.metaclasses import Model, RandomProcess, _assert_sample_and_rtype
-from pyrenew.models.rtinfectionsrenewal import RtInfectionsRenewalModel
-from pyrenew.processes import RtRandomWalkProcess
+from pyrenew.metaclasses import Model, RandomVariable, _assert_sample_and_rtype
```
dylanhmorris, line 3: if we're renaming `pyrenew.models` --> `pyrenew.model`, `pyrenew.processes` --> `pyrenew.process` etc, rename `pyrenew.metaclasses`--> `pyrenew.metaclass`?

dylanhmorris (commented):
1. One typo in the docs
2. Add text encoding notes back in back in?
3. Name of `metaclass` module?

dylanhmorris (approved):
Looks good. We can do the `metaclasses` --> `metaclass` renaming in a separate PR.

gvegayoncdc:
💨

PR #35, opened 2024-03-21
34 port notebooks under modeldocs to quarto

gvegayoncdc:
- Added the `qmd` versions of the jupyter notebooks (and their md versions).
- Added a couple of makefile entries to make building easier.
- Under model/docs. pre-commit will only check `*_cache`, if needed.


model/Makefile
```diff
@@ -4,4 +4,12 @@ install:
...
 
+docs: docs/pyrenew_demo.md docs/getting-started.md
+
+docs/pyrenew_demo.md: docs/pyrenew_demo.qmd
+	quarto render docs/pyrenew_demo.qmd
+
+docs/getting-started.md: docs/getting-started.qmd
+	quarto render docs/getting-started.qmd
+
 .PHONY: install test
```
gvegayoncdc, line 15:
```suggestion
.PHONY: install test docs
```

gvegayoncdc:
This is ready for review, @dylanhmorris, @natemcintosh, @AFg6K7h4fhy2, or @SamuelBrand1 :)

model/docs/getting-started.md
```diff
@@ -0,0 +1,224 @@
...
+axs[1].plot(range(0, 31), sim_data[1])
+axs[1].set_ylabel('Infections')
+
+fig.suptitle('Basic renewal model')
+fig.supxlabel('Time')
+plt.tight_layout()
+plt.show()
+```
+
+![](getting-started_files/figure-commonmark/basic-fig-1.png)
```
dylanhmorris, line 164: this does not seem to be getting committed. Might need to create a `.gitignore` exemption.

model/docs/getting-started.md
```diff
@@ -0,0 +1,224 @@
...
+    warmup:  35%|###4      | 1039/3000 [00:02<00:01, 1101.95it/s, 127 steps of size 3.26e-02. acc. prob=0.79]
+    warmup:  43%|####2     | 1280/3000 [00:02<00:01, 1374.32it/s, 63 steps of size 4.97e-02. acc. prob=0.79]
+    warmup:  51%|#####     | 1523/3000 [00:02<00:00, 1615.94it/s, 63 steps of size 4.24e-02. acc. prob=0.79]
+    warmup:  59%|#####8    | 1761/3000 [00:02<00:00, 1804.39it/s, 37 steps of size 3.88e-02. acc. prob=0.79]
+    warmup:  66%|######6   | 1990/3000 [00:02<00:00, 1920.65it/s, 127 steps of size 4.13e-02. acc. prob=0.79]
+    sample:  74%|#######3  | 2217/3000 [00:02<00:00, 1948.52it/s, 127 steps of size 3.43e-02. acc. prob=0.93]
+    sample:  81%|########1 | 2437/3000 [00:02<00:00, 1944.18it/s, 127 steps of size 3.43e-02. acc. prob=0.93]
+    sample:  88%|########8 | 2649/3000 [00:02<00:00, 1926.81it/s, 127 steps of size 3.43e-02. acc. prob=0.93]
+    sample:  95%|#########5| 2854/3000 [00:02<00:00, 1926.79it/s, 127 steps of size 3.43e-02. acc. prob=0.94]
+    sample: 100%|##########| 3000/3000 [00:03<00:00, 995.42it/s, 127 steps of size 3.43e-02. acc. prob=0.93]
```
dylanhmorris, line 198: Some way to prevent multi-line progress bar output in the rendered `.md`?

model/docs/getting-started.qmd
```diff
@@ -0,0 +1,185 @@
+---
+title: "Getting started with pyrenew"
+format: gfm
+engine: knitr
```
dylanhmorris, line 4: Do we definitely want a `knitr` rather than `jupyter` engine binding? As far as I can tell from [the docs](https://quarto.org/docs/computations/execution-options.html#engine-binding), using `knitr` as our engine will mean the Python execution happens via `reticulate`. If we don't specify an engine, Quarto should [default to `jupyter` given the presence of only Python code in the `.qmd`](https://quarto.org/docs/computations/execution-options.html#engine-binding). We could also explicitly specify `engine: jupyter`.

dylanhmorris (commented):
Some comments @gvegayoncdc

model/docs/getting-started.md
dylanhmorris: Update, yes:
https://github.com/CDCgov/multisignal-epi-inference/blob/2af2b0696db1285c61ffef0a8fd52112565c7049/.gitignore#L28

Could allow `.png` within `docs`?

dylanhmorris: I think that if we are going to keep the rendered `.md` docs in the repo under version control, we should also include any needed images. The alternative is not to keep the rendered `.md` under version control, but rather just the `.qmd` source files

model/docs/getting-started.qmd
gvegayoncdc: I think it makes sense using jupyter. I'll switch to that.

gvegayoncdc: Yes, Jupyter is faster than knitr.

dylanhmorris (dismissed):
LGTM. Merge when you're ready.

model/docs/getting-started.qmd
```diff
@@ -0,0 +1,186 @@
+---
+title: "Getting started with pyrenew"
+format: gfm
+engine: jupyter
+---
+
+This document illustrates two features of `pyrenew`: (a) the set of included `RandomVariable`es, and (b) model composition.
```
dylanhmorris, line 7: Typo 

```suggestion
This document illustrates two features of `pyrenew`: (a) the set of included `RandomVariable`s, and (b) model composition.

dylanhmorris (commented):
Spotted one typo

PR #39, opened 2024-03-21
Formally addresses implementing the observed hospitalizations module

ghost:
This PR contributes to the support of the hospital admissions signal. It includes the following:

- Adds `process.DeterministicProcess` and `observation.DeterministicObs` to the corresponding modules. This can be used to wrap fixed quantities when needed.
- Updates `latent.HospitalAdmisions` latent class to treat IHR, weekday effect, and p hosp as `RandomVariable`. Again, with the `DeterministicObs` these can be passed as fixed quantities (or inferred).
- Adds more details to the hospitalization bit of the `equations.md` file, mostly pointing to the wastewater model.
- Re-organizes the `getting-started.qmd` documentation to clarify.
- `latent.Infections` now receives a `RandomVariable` for the generation interval. Before it was a fixed array.
- Extends testing of `model.Hospitalizations` assessing that the new bits that can either do inference or reuse some of the new components (e.g., generation weekday effect).


model/src/pyrenew/process/deterministic.py
```diff
@@ -0,0 +1,73 @@
+from pyrenew.distutil import validate_discrete_dist_vector
+from pyrenew.metaclass import RandomVariable
+
+
+class DeterministicProcess(RandomVariable):
```
dylanhmorris, line 5: Noting that we plan to discuss a bit more whether and if so how to use this construct.

model/src/pyrenew/process/deterministic.py
```diff
@@ -0,0 +1,73 @@
...
+    """A wrapper instance of RandomVariable that allows pass deterministic
+    quantities"""
+
+    def __init__(
+        self,
+        vars: tuple,
+        label: str = "a_random_variable",
+        validate_pmf: bool = False,
```
dylanhmorris, line 13: I wouldn't declare this construct argument for the abstract base class. Should it be declared only for subclasses designed to represent discrete pmfs?

Indeed, one could have 
```python 
class DiscretePMF(DeterministicProcess)
```

model/src/pyrenew/process/deterministic.py
dylanhmorris: Another thing to consider if we do: the name. `DeterministicProcess` or `DeterministicVariable`?

model/src/pyrenew/process/deterministic.py
```diff
@@ -0,0 +1,73 @@
...
+
+        Parameters
+        ----------
+        vars : tuple
+            A tuple with arraylike objects.
+        label : str
+            A label to assign to the process.
+        validate_pmf : bool
+            When True, it will create of copy of vars and call
+            pyrenew.distutil.validate_discrete_dist_vector on each one of its
```
dylanhmorris, line 25: See above comment.

model/src/pyrenew/process/deterministic.py
```diff
@@ -0,0 +1,73 @@
...
+
+        self.validate(vars)
+
+        if validate_pmf:
+            vars2 = list(vars)
+            for i in range(0, len(vars2)):
+                vars2[i] = validate_discrete_dist_vector(vars2[i])
+            self.vars = tuple(vars2)
+        else:
+            self.vars = vars
```
dylanhmorris, line 41: See above comment

model/src/test/test_model_hospitalizations.py
```diff
@@ -4,25 +4,72 @@
...
+    @staticmethod
+    def validate(self):
+        return None
+
+    def sample(self, random_variables, constants):
+        return (
+            npro.sample(name=self.name, fn=dist.Uniform(high=0.99, low=0.01)),
+        )
+
+
```
dylanhmorris, line 31: Even just for a test case, I think I wouldn't declare this class with such a general name. Maybe something like:
```python 
class UniformProbForTest
```

dylanhmorris: Also I think it can be on the full unit interval.

dylanhmorris (commented):
Some comments @gvegayon

model/Makefile
```diff
@@ -12,4 +12,9 @@ docs/pyrenew_demo.md: docs/pyrenew_demo.qmd
 docs/getting-started.md: docs/getting-started.qmd
 	quarto render docs/getting-started.qmd
 
-.PHONY: install test docs
+clean:
+	rm -rf docs/*_files/
+	rm -f docs/getting-started.ipynb
+	rm -f docs/pyrenew_demo.ipynb
```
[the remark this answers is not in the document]
gvegayon: The thing is, when using jupyter as quarto engine, it generates the notebooks, which are only deleted once the document is successfully created. Otherwise, if it fails during compilation, the notebook is kept in there. The problem is that it is sort of used as cache, which is something we don't want it if failed! That said, I'll see if quarto has further options for the jupyter engine to avoid that behavior.

model/src/pyrenew/process/deterministic.py
gvegayon: Since it is under process, I voted for process :P. One thing worth discussing is that both process/deterministic.py and observation/deterministic.py are 99% the same, only difference is the name, so it makes me wonder if it should be moved to pyrenew under metaclass or something.

model/Makefile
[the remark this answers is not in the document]
dylanhmorris: Ah I see. That makes sense.

model/src/pyrenew/process/deterministic.py
gvegayon: This is not intended to be a metaclass. I actually use it in the code. I'll show you som examples later.

dylanhmorris: Even if you envision this as concrete, I'd still suggest making a child class for PMFs.

gvegayon: I see your point. Added to the discussion!

gvegayon: So perhaps, have a `deterministic` submodule? Sounds like a option, but probably overkill?

model/Makefile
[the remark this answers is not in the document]
gvegayon: Just checked, there aren't many options to play with in [quarto+jupyter](https://quarto.org/docs/computations/execution-options.html#jupyter-options). So I'll mark this as resolved.

model/src/pyrenew/process/deterministic.py
gvegayon: Added in the latest commit.

gvegayon:
OK, @dylanhmorris. I've included the discussed changes. More improvements will be addressed/discussed during the next sprint!

model/docs/getting-started.qmd
```diff
@@ -154,42 +161,49 @@ ax.set_yscale("log")
...
 
-Where the first three are collections of instances of `RandomVariable` and the last one a collection of instances of `Model`. The following diagram shows a detailed view of how meta classes, modules, and classes interact to create the `RtInfectionsRenewalModel` instantiated in the previous section:
+Where the first four are collections of instances of `RandomVariable` and the last one a collection of instances of `Model`. The following diagram shows a detailed view of how meta classes, modules, and classes interact to create the `RtInfectionsRenewalModel` instantiated in the previous section:
 
 ```{mermaid}
 %%| label: overview-of-RtInfectionsRenewalModel
-flowchart TB
+flowchart LR
     rand((RandomVariable\nmetaclass))
     models((Model\nmetaclass))
```
dylanhmorris, line 176: Suggest putting the model metaclass above the random variable metaclass in the diagram, so that the metaclasses form a single column

dylanhmorris (commented):
High-level comment: we need to standardize language on "Hospital admissions" versus "Hospitalizations". I prefer the former ("Hospital admissions"), because there's less chance of confusion between incidence and prevalence.

gvegayon:
> High-level comment: we need to standardize language on "Hospital admissions" versus "Hospitalizations". I prefer the former ("Hospital admissions"), because there's less chance of confusion between incidence and prevalence.

Yes, I think it is worthwhile having a single issue to standardize language across the package. And have a document to summarize all.

model/docs/getting-started.qmd
gvegayon: I know, but is so wide! 

```mermaid
flowchart TB
    rand((RandomVariable\nmetaclass))
    models((Model\nmetaclass))

    subgraph observations[Observations module]
        obs["observed_infections\n(PoissonObservation)"]
    end

    subgraph latent[Latent module]
        inf["latent_infections\n(Infections)"]
        i0["I0\n(Infections0)"]
    end

    subgraph process[Process module]
        rt["rt_proc\n(RtRandomWalkProcess)"]
    end

    subgraph deterministic[Deterministic module]
        detpmf["gen_int\n(DeterministicPMF)"]
    end

    subgraph model[Model module]
        model1["model1\n(RtInfectionsRenewalModel)"]
    end

    rand-->|Inherited by|observations
    rand-->|Inherited by|process
    rand-->|Inherited by|latent
    rand-->|Inherited by|deterministic
    models-->|Inherited by|model

    detpmf-->|Composes|model1
    i0-->|Composes|model1
    rt-->|Composes|model1
    obs-->|Composes|model1
    inf-->|Composes|model1
```

```mermaid
flowchart LR
    rand((RandomVariable\nmetaclass))
    models((Model\nmetaclass))

    subgraph observations[Observations module]
        obs["observed_infections\n(PoissonObservation)"]
    end

    subgraph latent[Latent module]
        inf["latent_infections\n(Infections)"]
        i0["I0\n(Infections0)"]
    end

    subgraph process[Process module]
        rt["rt_proc\n(RtRandomWalkProcess)"]
    end

    subgraph deterministic[Deterministic module]
        detpmf["gen_int\n(DeterministicPMF)"]
    end

    subgraph model[Model module]
        model1["model1\n(RtInfectionsRenewalModel)"]
    end

    rand-->|Inherited by|observations
    rand-->|Inherited by|process
    rand-->|Inherited by|latent
    rand-->|Inherited by|deterministic
    models-->|Inherited by|model

    detpmf-->|Composes|model1
    i0-->|Composes|model1
    rt-->|Composes|model1
    obs-->|Composes|model1
    inf-->|Composes|model1
```

But if you feel strongly, I can change it!

dylanhmorris: My suggestion was still to use the portrait layout (second of the two here), but just with the circles all on the left (which to avoid overlapping arrows will require `Model metaclass` to go above or below `RandomVariable metaclass`. On reflection, I think above would be best, since its child `Model module` is labeled at the top.

gvegayon: For some reason your suggestion was not up to date. But we are on the same page. LR it is!

model/src/pyrenew/latent/hospitaladmissions.py
```diff
@@ -104,7 +104,7 @@ class HospitalAdmissions(RandomVariable):
 
     def __init__(
         self,
-        inf_hosp_int: RandomVariable,
+        inform_hosp: RandomVariable,
```
dylanhmorris, line 107: `infection_to_hospitalization_interval`

gvegayon: Is that the same as the delay? In the equation we have:

$$
H(t) = \omega(t) p_\mathrm{hosp}(t) \sum_{\\tau = 0}^{T_d} d(\tau) I(t-\tau
$$

where:

- $\omega(t)$: `weekday_effect_dist`
- $p_\mathrm{hosp}(t)$: `p_hosp_dist`
- $I(t)$: IHR (`infect_hosp_rate_dist`) * `random_variables.get(self.infections_varname)`
- $d(\tau)$: `inf_hosp_int` (delay)

Here is the code:

https://github.com/CDCgov/multisignal-epi-inference/blob/517fce1ab66c8d589577f027f6dcaf4eebd3733f/model/src/pyrenew/latent/hospitaladmissions.py#L194-L208

dylanhmorris: Yes; it's the delay distribution from infection to (reported) admission, represented as a discrete PMF

gvegayon: OK, renamed in [ddac288](https://github.com/CDCgov/multisignal-epi-inference/pull/39/commits/ddac28844aba2e66c6d5ce1c1b7369965f569c77)

model/src/pyrenew/latent/hospitaladmissions.py
```diff
@@ -16,54 +15,153 @@
...
+)
+
+
+class InfectHospRate(RandomVariable):
+    """Infection to Hospitalization Rate"""
+
+    def __init__(
+        self,
+        varname: str = "IHR",
+        dist: dist.Distribution = dist.LogNormal(jnp.log(0.05), 0.05),
```
dylanhmorris, line 31: Shouldn't preserve the default now that this is modular; default was reasonable in the `basic.py` example, but this should have no default.

model/src/pyrenew/latent/hospitaladmissions.py
```diff
@@ -16,54 +15,153 @@
...
+        infect_hosp_rate_varname : str
+            Name of the entry in random_variables that holds the observed
+            infection-hospitalization rate (IHR).
             (if available).
         hospitalizations_predicted_varname : str
             Name to assign to the deterministic component in numpyro of
             predicted hospitalizations.
-        IHR_dist : dist.Distribution, optional
+        infect_hosp_rate_dist : RandomVariable, optional
             Infection to hospitalization rate pmf.
```
dylanhmorris, line 132: Not a PMF but rather a distribution from which to sample the IHR

model/src/pyrenew/latent/infections.py
```diff
@@ -25,18 +19,18 @@ class Infections(RandomVariable):
...
     ) -> None:
         """Default constructor
 
         Parameters
         ----------
-        gen_int: ArrayLike
-            A vector representing the pmf of the generation interval
+        gen_int_varname : str.
+            Name of the element in `random_variables` that will hold the value
+            of 'generation interval'.
```
dylanhmorris, line 33: I think that if we find ourselves doing things like this it's a good argument for the `sample(named_arg1, named_arg2, **kwargs)` pattern instead.

dylanhmorris (commented):
Some comments

model/src/pyrenew/latent/infections.py
gvegayon: Agreed. This will involve a no minor lift, so I'll create a separate issue for this.

model/src/pyrenew/latent/hospitaladmissions.py
gvegayon: Addressed in [d0851f9](https://github.com/CDCgov/multisignal-epi-inference/pull/39/commits/d0851f9f3c3289950f4fca8c855a67fe36ad8ed9)

model/src/pyrenew/latent/hospitaladmissions.py
gvegayon: Yes, agree. I think this should be part of the language issue #50. I'm sure there's more of this around it. I'll mark it as resolve, but add an explicit reference of this to the issue.

model/src/pyrenew/latent/infections.py
gvegayon: Furthermore, we should talk about this tomorrow. I'll put it on the agenda.

gvegayon:
@dylanhmorris, this is ready for review.

model/src/pyrenew/latent/hospitaladmissions.py
```diff
@@ -16,54 +15,153 @@
...
 
         Parameters
         ----------
-        inf_hosp_int : ArrayLike
-            pmf for reporting (informing) hospitalizations.
+        infection_to_admission_interval : RandomVariable
+            pmf for reporting (informing) hospitalizations (see
+            pyrenew.observations.Deterministic).
+        infect_hosp_rate_dist : RandomVariable
+            Infection to hospitalization rate pmf.
```
dylanhmorris, line 122:
```suggestion
            Infection to hospitalization rate distribution.

File: model/src/pyrenew/latent/hospitaladmissions.py [hunk 13]
```diff
@@ -16,54 +15,153 @@
...
         self,
-        inf_hosp_int: ArrayLike,
+        infection_to_admission_interval: RandomVariable,
+        infect_hosp_rate_dist: RandomVariable,
         infections_varname: str = "infections",
-        IHR_obs_varname: str = "IHR",
         hospitalizations_predicted_varname: str = "predicted_hospitalizations",
-        IHR_dist: dist.Distribution = dist.LogNormal(jnp.log(0.05), 0.05),
+        weekday_effect_dist: RandomVariable = DeterministicVariable((1,)),
+        p_hosp_dist: RandomVariable = DeterministicVariable((1,)),
```
dylanhmorris, line 112: Since this represents a reporting rate/probability, suggest renaming it to `p_report_dist` throughout.

model/src/pyrenew/latent/i0.py
```diff
@@ -0,0 +1,77 @@
...
+    """Initial infections helper class.
+
+    It creates a random variable for the initial infections with a prior
+    distribution.
+    """
+
+    def __init__(
+        self,
+        name: str = "I0",
+        I0_dist: dist.Distribution = dist.LogNormal(0, 1),
```
dylanhmorris, line 16: I do not think this should have a default prior.

dylanhmorris (changes requested):
Minor changes

model/src/pyrenew/latent/i0.py
gvegayon: Addressed in 33d9ed0

model/src/pyrenew/latent/hospitaladmissions.py
gvegayon: Addressed in 33d9ed0

gvegayon:
Implemented requested changes in 33d9ed0 and 39c54c4. The two existing models do not have any default values now.

model/README.md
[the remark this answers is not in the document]
gvegayon: I like [chicago style](https://en.wikipedia.org/wiki/Title_case#:~:text=Chicago%20Manual%20of%20Style%5B,subtitles%20(overrides%20the%20rules%20above).), so I would go with alt opt. 2. I don't think the current default falls under any style.

model/docs/getting-started.md
```diff
@@ -1,127 +1,142 @@
...
+## The fundamentals
+
+`pyrenew`’s core components are the metaclasses `RandomVariable` and
+`Model`. From the package’s perspective, a `RandomVariable` is a
+quantity models can sample and estimate, **including deterministic
+quantities**. Mainly, sampling from a `RandomVariable` involves calling
+the `sample()` method. The benefit of this design is the definition of
+the sample function can be arbitrary, allowing the user to either sample
+from a distribution using `numpyro.sample()`, compute fixed quantities
+(like a mechanistic equation,) or return a fixed value (like a
```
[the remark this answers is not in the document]
gvegayon: [(Oxford comma fan here.)](https://www.chicagomanualofstyle.org/16/ch06/ch06_sec096.html) The comma is right, the only thing that I would change is where the comma goes. Although in the past it was a common practice to include a comma right before closing a parenthesis (like [here](https://www.bartleby.com/lit-hub/prose-works/2-democratic-vistas-paras-90119/#:~:text=Not%20but%20that,they%20last%20forever.)), the [Chicago manual does not like that](https://www.chicagomanualofstyle.org/16/ch06/ch06_sec053.html).

model/README.md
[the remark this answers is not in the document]
gvegayon: Also, found this [title case converter](https://titlecaseconverter.com/)... very nice!

model/docs/getting-started.md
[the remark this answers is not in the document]
gvegayon: Addressed in 25fa66a

model/README.md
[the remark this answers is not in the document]
gvegayon: Addressed in 25fa66a

model/src/pyrenew/latent/hospitaladmissions.py
```diff
@@ -132,7 +132,7 @@ def __init__(
             predicted hospitalizations.
         weekday_effect_dist : RandomVariable, optional
             Weekday effect.
-        p_hosp_dist : RandomVariable, optional
+        p_report_dist  : RandomVariable, optional
             Hospitalization probability.
```
dylanhmorris, line 136:
```suggestion
            Reporting probability for hospital admissions. Defaults to 1 (full reporting).

File: model/src/test/test_model_hospitalizations.py [hunk 17]
```diff
@@ -367,7 +367,7 @@ def test_model_hosp_with_obs_model_weekday_phosp():
         infection_to_admission_interval=inf_hosp,
         infections_varname="infections",
         weekday_effect_dist=weekday,
-        p_hosp_dist=p_hosp,
+        p_report_dist=p_hosp,
```
dylanhmorris, line 370: I'd rename the input variable as well as the argument name, for maximum clarity.

dylanhmorris (changes requested):
A few final tweaks needed.

model/src/test/test_model_hospitalizations.py
gvegayon: Addressed in 134f786

model/src/pyrenew/latent/hospitaladmissions.py
```diff
@@ -132,8 +132,9 @@ def __init__(
             predicted hospitalizations.
         weekday_effect_dist : RandomVariable, optional
             Weekday effect.
-        p_report_dist  : RandomVariable, optional
-            Hospitalization probability.
+        hosp_report_prob_dist  : RandomVariable, optional
+            Reporting probability for hospital admissions. Defaults to 1 (full
```
dylanhmorris, line 136:
```suggestion
            Distribution or fixed value for the hospital admission reporting probability. Defaults to 1 (full

File: model/src/test/test_model_hospitalizations.py [hunk 19]
```diff
@@ -266,7 +266,7 @@ def test_model_hosp_with_obs_model_weekday_phosp_2():
         infection_to_admission_interval=inf_hosp,
         infections_varname="infections",
         weekday_effect_dist=weekday,
-        p_report_dist=p_hosp,
+        hosp_report_prob_dist=p_hosp,
```
dylanhmorris, line 269: rename `p_hosp` in line 262 and here

model/src/test/test_model_hospitalizations.py
```diff
@@ -367,7 +367,7 @@ def test_model_hosp_with_obs_model_weekday_phosp():
...
```
dylanhmorris, line 370: rename `p_hosp` to `p_report` in line 363 and here. Also, shouldn't it be `p_report = DeterministicVariable((<some scalar>,))`, not `p_report = DeterministicVariable((weekday,))`?

dylanhmorris (changes requested):

model/src/test/test_model_hospitalizations.py
dylanhmorris: If we're testing a vector-valued `p_report`, then for robustness shouldn't it be a different vector from `weekday`?

dylanhmorris (dismissed):
LGTM

model/src/test/test_model_hospitalizations.py
gvegayon: Sure, I'll make it separate before merging. Thanks!

gvegayon (approved):
All suggestions have been addressed. This is a bit of a bottle neck for other issues, so I'll go ahead and merge it. Thanks for your input, @dylanhmorris!

PR #48, opened 2024-03-28
My patch (playground) to the sphinx patch

gvegayon:
Edits to the sphinx workflow checking how to deploy it. It works, but we need to make sure we are publishing where it is supposed to be. Right now, it is using `github-pages-test` environment.

### DIFF
--- a/.github/workflows/sphinx.yaml
+++ b/.github/workflows/sphinx.yaml
@@ -9,31 +9,50 @@ jobs:
       contents: write
     steps:
     - uses: actions/checkout@v4
-    # - name: Build HTML
-    #   uses: ammaraskar/sphinx-action@master
+
+    # Setting up the environment
+    - name: Setup Pages
+      uses: actions/configure-pages@v4
+
     - name: Set up Python
       uses: actions/setup-python@v2
       with:
         python-version: '3.11'
-    # - uses: abatilo/actions-poetry@v3.0.0
+
     - name: Install Poetry
       run: pip install poetry
+
     - name: Install dependencies with Poetry
       run: |
         cd docs
         poetry install
+
+    # Building
     - name: Build HTML using Poetry
       run: |
         cd docs
         poetry run sphinx-build -M html source/ build/
+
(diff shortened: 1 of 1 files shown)


gvegayon:
Hey, @AFg6K7h4fhy2, this workflow file is now working! https://cdcgov.github.io/multisignal-epi-inference/

We just need to set preview to false and make sure where are we releasing the website (environments).

cc @dylanhmorris

.github/workflows/sphinx.yaml
```diff
@@ -1,39 +1,66 @@
...
+    - name: Upload artifacts
+      uses: actions/upload-pages-artifact@v3
+      with:
+        name: github-pages-test
+        path: docs/build/html/
+
+  deploy:
+    # Deploy to the github-pages environment
+    environment:
+      name: github-pages-test
```
gvegayon, line 52: May need change

.github/workflows/sphinx.yaml
```diff
@@ -1,39 +1,66 @@
...
+
+    # Specify runner + deployment step
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Deploy to GitHub Pages
+      id: deployment
+      uses: actions/deploy-pages@v4
+      with:
+        artifact_name: github-pages-test
```
gvegayon, line 65: May need change

.github/workflows/sphinx.yaml
```diff
@@ -1,39 +1,66 @@
...
-#         path: docs/build/html/
-#     - name: Deploy
-#       uses: peaceiris/actions-gh-pages@v3
-#       if: github.ref == 'refs/heads/main'
-#       with:
-#         github_token: ${{ secrets.GITHUB_TOKEN }}
-#         publish_dir: docs/build/html
+name: "Sphinx: Render docs"
+
+on: push
```
gvegayon, line 3: Will need to specify which branches it supports. Not sure if it should only be main.

PR #49, opened 2024-03-29
Sphinx(it) deploy of doc

gvegayon:
@AFg6K7h4fhy2, I've opened the PR for you. I think we are almost there to merge. I'll check on the github-pages environemnt to ensure all works OK (and delete the github-pages-test environment.)


gvegayon:
This is ready for review. We can figure out the details later on. The only thing I would double-check is that we are not adding any file that is not human-generated. @AFg6K7h4fhy2, can you double-check that?

.github/workflows/sphinx.yaml
```diff
@@ -0,0 +1,62 @@
...
+      contents: write
+    steps:
+    - uses: actions/checkout@v4
+
+    # Setting up the environment
+    - name: Setup Pages
+      uses: actions/configure-pages@v4
+
+    - name: Set up Python
+      uses: actions/setup-python@v2
```
dylanhmorris, line 24: Use latest version 
```suggestion
      uses: actions/setup-python@v5
```

dylanhmorris (commented):
Minor suggestion

PR #59, opened 2024-04-03
Setup of quarto docs

gvegayon:
It sets up a GHA that:

- Installs poetry.
- Installs pyrenew.
- Sets up quarto.
- Renders all documents under `model/docs/*.qmd`

Eventually, this action could be used to generate the `rst` documents needed for the website. For the moment, is just a check that the documentation can properly render.


.github/workflows/docs.yaml
```diff
@@ -0,0 +1,40 @@
+name: Quarto documents
+
+on:
+  pull_request:
```
dylanhmorris, line 4: Should probably run only on PRs to main, no?

```suggestion
  pull_request:
     branches: [main]

File: .github/workflows/docs.yaml [hunk 2]
```diff
@@ -0,0 +1,40 @@
...
+  push:
+    branches: [main]
+
+jobs:
+  install-and-test:
```
dylanhmorris, line 9: rename job

.github/workflows/docs.yaml
```diff
@@ -0,0 +1,40 @@
...
+
+      - name: installing other dependencies
+        run: poetry run -C model pip install pyyaml nbclient nbformat
+
+      - name: Render documents
+        run: |
+            cd model
+            for file in $(find docs -name '*.qmd'); do
+                poetry run quarto render $file
+            done
```
dylanhmorris, line 40: I don't think this will actually commit the newly rendered docs to the repo, but maybe I've misunderstood the goal here.

.github/workflows/docs.yaml
gvegayon: What about production/tagging whatever we end up calling the future branches? We can leave it to the main only, but I just wanted to comment on that so we have it on the back of our minds!

.github/workflows/docs.yaml
gvegayon: Yes, this is not commiting the rendered docs, so it only checks that they render. In the future, I hope to use this as a reference for us to create the rst documents. Perhaps we can do that now?

.github/workflows/docs.yaml
gvegayon: done in d434cc9

.github/workflows/docs.yaml
dylanhmorris: I have no issue with just checking that the docs rendering successfully if that was the goal.

My two thoughts:

1. If we're still intending to have github-flavored Markdown versions of the rendered tutorials tracked in the repo, why not commit the versions rendered in CI?
2.  That said, there's an argument that we shouldn't have rendered tutorials in the repo, namely the Python adage that ["There should be one-- and preferably only one --obvious way to do it."](https://peps.python.org/pep-0020/). With that in mind, perhaps the rendered tutorials should exist _only_ on the documentation website.

.github/workflows/docs.yaml
dylanhmorris: Yes, I think if in the future we have other key persistent branches those should be added here.

dylanhmorris (approved):
I think this is good to go, but we should continue discussing https://github.com/CDCgov/multisignal-epi-inference/pull/59/files#r1552090213

.github/workflows/docs.yaml
gvegayon: 1. I was thinking about that. The one problem I see is when you are working on that file, you'd like to compile it to ensure it is running properly.
2. I see the point, perhaps have `model/docs/README.md` pointing to the website with the compiled docs and remove the `md` files from the tree (just keep qmd)?

I'll open an issue for this.

PR #62, opened 2024-04-03
Do not attempt to deploy Sphinx site in PRs

### LINKED ISSUE #61: Only try to deploy docs from pushes to `main`, not PRs
> Also: the `deploy` step of the Sphinx build workflow should only run on `main`, otherwise we'll get failures like the one we have here.

_Originally posted by @dylanhmorris in https://github.com/CDCgov/multisignal-epi-inference/pull/39#pullrequestreview-1977716168_

More precisely, only on pushes to `main` rather than PRs into `main`.

dylanhmorris:
Fixes #61

### DIFF
--- a/.github/workflows/sphinx.yaml
+++ b/.github/workflows/sphinx.yaml
@@ -1,6 +1,12 @@
 name: "Sphinx: Render docs"
 
-on: push
+on:
+  push:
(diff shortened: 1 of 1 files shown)


gvegayon (approved):
LGTM

Issue #47, opened 2024-03-28
First pass on documentation


gvegayon:
## Goal

Give a first pass on the documentation, ensuring all the basics are there.

## Context

Agreed by the team, we want to have good documentation from the beginning of the project. 

## Required features

1. Each python function under `model/src/pyrenew` should:

  - Have a docstring featuring (i) a description, (2) parameters, and (3) outputs following [numpy's style](https://numpydoc.readthedocs.io/en/latest/format.html).
  - All functions should have type hints ([link](https://docs.python.org/3/library/typing.html))
    
    For the latter, here is an example of a good typed hinted function:
    
    ```python
    import jax.numpy as jnp
    from pyrenew.metaclass import RandomVariable
    
    def myfun(
      x: str,
      y: jnp.array,
      z: RandomVariable
      ) -> None
    ```

2. In the case of the demos under `model/docs`, these should also be checked.

3. `__init__.py` files should have a description of the module/submodule (currently missing).

## Specifications

The following files under `model/src/pyrenew` should be checked:

@AFg6K7h4fhy2:
- [ ] Top-level directory.
- [x]  `latent/`
- [x] `model/`
- [x] `observation/`

@cshelley 

- [x] `process`
- [x] Quarto files under `model/docs`.

Create a single PR per checkbox. PRs should use #16 as a baseline (so either wait until it is merged or use it as a baseline when creating the branch).

gvegayon:
@cshelley, I would start 1 PR for checking the stuff under `docs`, but first, wait for #39 to be merged, as I've added lots of docs there. @AFg6K7h4fhy2, let's meet with @cshelley to see how we distribute this.

Issue #38, opened 2024-03-21
pre-commit numpydoc validation


ghost:
## Goal

To enforce the usage of `numpydoc` for documenting functions, we can leverage a pre-commit hook designed for this: https://numpydoc.readthedocs.io/en/latest/validation.html. This could be a heavy lift!

## Context

All issues tagged with `documentation`.

## Specifications

- [ ] Setup the pre-commit hook to validate numpydoc
- [ ] Update the model package once the pre-commit hook is running.

dylanhmorris:
@AFg6K7h4fhy2 are you up for doing this?

gvegayoncdc:
This will be a nice tool, but as I said earlier, it is heavy lifting! I suggest giving priority to the sphinx website before working more in this.

gvegayon:
Have you asked AI to help you with that?

AFg6K7h4fhy2:
Pasting this here so we don't have to look at the site as often

```
ERROR_MSGS = {
    "GL01": "Docstring text (summary) should start in the line immediately "
    "after the opening quotes (not in the same line, or leaving a "
    "blank line in between)",
    "GL02": "Closing quotes should be placed in the line after the last text "
    "in the docstring (do not close the quotes in the same line as "
    "the text, or leave a blank line between the last text and the "
    "quotes)",
    "GL03": "Double line break found; please use only one blank line to "
    "separate sections or paragraphs, and do not leave blank lines "
    "at the end of docstrings",
    "GL05": 'Tabs found at the start of line "{line_with_tabs}", please use '
    "whitespace only",
    "GL06": 'Found unknown section "{section}". Allowed sections are: '
    "{allowed_sections}",
    "GL07": "Sections are in the wrong order. Correct order is: {correct_sections}",
    "GL08": "The object does not have a docstring",
    "GL09": "Deprecation warning should precede extended summary",
    "GL10": "reST directives {directives} must be followed by two colons",
    "SS01": "No summary found (a short summary in a single line should be "
    "present at the beginning of the docstring)",
    "SS02": "Summary does not start with a capital letter",
    "SS03": "Summary does not end with a period",
    "SS04": "Summary contains heading whitespaces",
    "SS05": "Summary must start with infinitive verb, not third person "
    '(e.g. use "Generate" instead of "Generates")',
    "SS06": "Summary should fit in a single line",
    "ES01": "No extended summary found",
    "PR01": "Parameters {missing_params} not documented",
[...]

PR #63, opened 2024-04-03
Refactoring sample from dicts to named arguments + **kwargs

gvegayon:
This PR includes the following:

- Updates the metaclasses `RandomVariable` and `Model` to use `**kwargs` for `sample()`.
- Updates all instances of these under the submodules.
- Within the `model` module, it ensures that arguments are via keywords instead of dictionaries.
- Updates the tests to reflect the change.
- Updates the demos under docs to reflect the change.

**For a new discussion**: Now that I've done this refactoring, the need to standardize the submodules became apparent. 

- When specifying the observation process for any of the models, it makes sense to have the `RandomVariable.sample()` method of `observed` class feature `mean` and `obs`. This way the `HospitalizationModel` would always expect the observation process to receive `mean` for the latent hospitalizations and `obs` for the observed hospitalizations.
- Other arguments could still be passed as now `**kwargs` are unpacked whenever calling a `RandomVariable.sample()` method.
- Like the `observed`, `process` could also have defaults for `RandomVariable.sample()`, e.g., `n_timesteps`.

### DIFF
--- a/model/docs/pyrenew_demo.md
+++ b/model/docs/pyrenew_demo.md
@@ -21,8 +21,6 @@ import numpyro.distributions as dist
 from pyrenew.process import SimpleRandomWalkProcess
 ```
 
-    An NVIDIA GPU may be present on this machine, but a CUDA-enabled jaxlib is not installed. Falling back to cpu.
-
 ``` python
 np.random.seed(3312)
 q = SimpleRandomWalkProcess(dist.Normal(0, 0.001))
@@ -71,10 +69,7 @@ latent_hospitalizations = HospitalAdmissions(
     )
 
 # And observation process for the hospitalizations
-observed_hospitalizations = PoissonObservation(
-    rate_varname='latent',
-    counts_varname='observed_hospitalizations',
-    )
+observed_hospitalizations = PoissonObservation()
 
 # And a random walk process (it could be deterministic using
 # pyrenew.process.DeterministicProcess())
@@ -93,7 +88,7 @@ hospmodel = HospitalizationsModel(
 
 ``` python
 with seed(rng_seed=np.random.randint(1, 60)):
-    x = hospmodel.sample(constants=dict(n_timepoints=30))
+    x = hospmodel.sample(n_timepoints=30)
 x
 ```
 
@@ -130,15 +125,12 @@ for axis in ax[:-1]:
 ![](pyrenew_demo_files/figure-commonmark/fig-hosp-output-1.png)
 
 ``` python
-sim_dat={"observed_hospitalizations": x.sampled}
-constants = {"n_timepoints":len(x.sampled)-1}
-
 # from numpyro.infer import MCMC, NUTS
 hospmodel.run(
     num_warmup=1000,
     num_samples=1000,
-    random_variables=sim_dat,
-    constants=constants,
+    observed_hospitalizations=x.sampled,
+    n_timepoints = len(x.sampled)-1,
     rng_key=jax.random.PRNGKey(54),
     mcmc_args=dict(progress_bar=False),
     )
--- a/model/src/pyrenew/deterministic/deterministic.py
+++ b/model/src/pyrenew/deterministic/deterministic.py
@@ -47,7 +47,7 @@ class DeterministicVariable(RandomVariable):
 
         Parameters
         ----------
-        kwargs : dict
+        **kwargs : dict, optional
             Ignored.
 
         Returns
--- a/model/src/pyrenew/deterministic/deterministic.py
+++ b/model/src/pyrenew/deterministic/deterministic.py
@@ -41,18 +41,14 @@ class DeterministicVariable(RandomVariable):
 
     def sample(
         self,
-        random_variables: dict = None,
-        constants: dict = None,
+        **kwargs,
     ) -> tuple:
         """Retrieve the value of the deterministic Rv
 
         Parameters
         ----------
-
-        random_variables : dict
-            Ignored. Default None.
-        constants : dict
-            Ignored. Default None.
+        kwargs : dict
+            Ignored.
 
         Returns
         -------
--- a/model/src/pyrenew/model/hospitalizations.py
+++ b/model/src/pyrenew/model/hospitalizations.py
@@ -128,7 +128,8 @@ class HospitalizationsModel(Model):
         n_timepoints : int
             Number of timepoints to sample (passed to the basic renewal model).
         **kwargs : dict, optional
-            Keyword arguments passed to the sampling methods.
+            Additional keyword arguments passed through to internal `sample()`
+            calls, if any
 
         Returns
         -------
--- a/model/src/pyrenew/model/hospitalizations.py
+++ b/model/src/pyrenew/model/hospitalizations.py
@@ -127,7 +127,7 @@ class HospitalizationsModel(Model):
         ----------
         n_timepoints : int
             Number of timepoints to sample (passed to the basic renewal model).
-        kwargs : dict
+        **kwargs : dict, optional
             Keyword arguments passed to the sampling methods.
 
         Returns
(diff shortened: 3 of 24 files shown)


model/src/pyrenew/deterministic/deterministic.py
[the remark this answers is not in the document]
gvegayon: Great point, @AFg6K7h4fhy2! I'll make a note on the metaclass. The idea is to have it as a standard. All methods should have `**kwargs`. This will ensure developers (us) that sample functions of `RandomVariable` and `Model` will always accept `**kwargs` (and thus must be passed through).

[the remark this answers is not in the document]
dylanhmorris: Tip: you can comment on individual lines of a file by clicking the blue plus button on mouseover

model/src/pyrenew/deterministic/deterministic.py
```diff
@@ -41,18 +41,14 @@ def validate(vars: tuple) -> None:
...
 
         Parameters
         ----------
-
-        random_variables : dict
-            Ignored. Default None.
-        constants : dict
-            Ignored. Default None.
+        kwargs : dict
+            Ignored.
```
dylanhmorris, line 51: Agree with @AFg6K7h4fhy2 that I think it would be best to say something like 
```Python
"""
**kwargs : dict, optional
    Additional keyword arguments passed through to internal `sample()` calls, if any
"""
```

model/src/pyrenew/model/hospitalizations.py
```diff
@@ -62,7 +64,7 @@ def __init__(
             gen_int=gen_int,
             I0=I0,
             latent_infections=latent_infections,
-            observed_infections=None,
+            observed_infections=DeterministicVariable((0,)),
```
dylanhmorris, line 67: [Mutable default](https://docs.python-guide.org/writing/gotchas/)

model/src/pyrenew/model/hospitalizations.py
```diff
@@ -76,23 +78,24 @@ def __init__(
...
+        **kwargs,
     ) -> tuple:
         return self.latent_hospitalizations.sample(
-            random_variables=random_variables,
-            constants=constants,
+            latent=infections,
+            **kwargs,
         )
 
     def sample_hospitalizations_obs(
```
dylanhmorris, line 94: Needs updated docstring (still refers to `constants` and `random_variables`)

dylanhmorris (changes requested):
Some tweaks.

RE: `Observation` classes: as discussed in person, we should standardize the name of the predicted value and observed value keyword arguments to an `Observation`'s `sample()` method. I think `obs` for observed (to be consistent with `numpyro.sample()`. For the prediction, how about `pred`, which is nicely consistent with `obs`?

model/src/pyrenew/model/hospitalizations.py
gvegayon: But that's within the function's body, not the definition, so it should be OK, right? Also, is it everything a mutable default?  How about strings?

model/src/pyrenew/deterministic/deterministic.py
gvegayon: In some cases, the kwargs are explicitly ignored, i.e., not passed to anything. But we keep them for consistency. Should I still say these could be passed? That doesn't seem right.

model/src/pyrenew/model/hospitalizations.py
gvegayon: You are commenting on the deleted code.

model/src/pyrenew/model/hospitalizations.py
gvegayon: I think the code is OK, @dylanhmorris

gvegayon: Also, I changed a couple of instances where a `RandomVariable` had a default value to instead have None. Those cases with default strings I left them as strings are inmutable.

gvegayon:
@AFg6K7h4fhy2 and @dylanhmorris, all done with your requests:

- Improved doc on `**kwargs`: I still left a couple where it says `Ignored`, but that's because it is literally that. 
- Renamed `mean` to `predicted` for the observation processes.
- Replaced a default value `RandomVariable` to none to avoid mutable defaults (but left a few strings, as strings are inmutable).

model/src/pyrenew/deterministic/deterministic.py
dylanhmorris: Clarification. As I understand it, the pattern we're planning to use is the following:

```Python
class Example(RandomVariable):
   
    def __init__(self):
           self.internal_rv_ = OtherRandomVariable()

    def sample(self, mandatory_named_arg: ArrayLike = None, **kwargs):
           self.do_something_with_named_arg(mandatory_named_arg)
           self.interval_rv_.sample(**kwargs) 
```

So any `RandomVariable.sample()` method only refers to named arguments within its own function body, but it also passes any additional `**kwargs` it receives onward to any calls to other `RandomVariable.sample() methods` that happen under the hood.

model/src/pyrenew/model/hospitalizations.py
dylanhmorris: Oops, sorry!

model/docs/pyrenew_demo.md
```diff
@@ -93,7 +88,7 @@ hospmodel = HospitalizationsModel(
 
 ``` python
 with seed(rng_seed=np.random.randint(1, 60)):
-    x = hospmodel.sample(constants=dict(n_timepoints=30))
+    x = hospmodel.sample(n_timepoints=30)
```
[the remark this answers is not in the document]
dylanhmorris: Good point. No reason not to say something like `Suppose our timestep size is one day; here's how we'd simulate about a month (30 days) worth of data` in the docs.

model/src/pyrenew/observation/poisson.py
```diff
@@ -23,40 +22,36 @@ def __init__(
...
         self.eps = eps
 
         return None
 
     def sample(
         self,
-        random_variables: dict,
-        constants: dict = None,
+        predicted: ArrayLike,
+        obs: ArrayLike = None,
```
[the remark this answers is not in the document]
dylanhmorris: Update: #65 and #66

model/docs/pyrenew_demo.md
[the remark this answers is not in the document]
gvegayon: This is something that will be relevant for #47 (cc @cshelley). Tagging to be addressed then.

model/src/pyrenew/process/simplerandomwalk.py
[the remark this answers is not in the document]
gvegayon: Keeping that for now, as it is really ignored.

model/docs/pyrenew_demo.md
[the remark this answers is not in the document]
dylanhmorris: Sounds good. @AFg6K7h4fhy2 make an issue and assign yourself / @cshelley ?

model/src/pyrenew/process/rtrandomwalk.py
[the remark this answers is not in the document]
gvegayon: Can you point to a specific case? I think most are `<arg> : <type>` (which is what numpydoc does).

model/src/pyrenew/model/hospitalizations.py
gvegayon: This was addressed in 44cb3d9

model/src/pyrenew/process/rtrandomwalk.py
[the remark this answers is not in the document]
gvegayon: Also, I think this kind of thing will be addressed once #38 is up and running!

model/src/pyrenew/model/hospitalizations.py
dylanhmorris: Oops, I was tricked by how GH collapsed the diffs.

PR #73, opened 2024-04-12
First Pass Documentation (docstrings) For Latent Folder

AFg6K7h4fhy2:
This pull request includes a first pass 

* Correction and or modifications of existing docstrings in `./model/src/pyrenew/latent`
* Addition of new docstrings in `./model/src/pyrenew/latent`

Note: You will find some TODO placeholders, mostly in the Notes section of some docstrings. These occur where the author was uncertain about standard procedure or the content.


model/pyproject.toml
```diff
@@ -15,8 +15,26 @@ polars = "^0.20.13"
 matplotlib = "^3.8.3"
 pillow = "^10.3.0" # See #56 on CDCgov/multisignal-epi-inference
 ipykernel = "^6.29.3"
+numpydoc = "^1.6.0"
```
natemcintosh, line 18: Should this go in the `[tool.poetry.group.test.dependencies]` section? Or is it needed to run the models?

model/src/pyrenew/latent/hospitaladmissions.py
```diff
@@ -14,30 +15,65 @@
...
+        Validates distribution is Numpyro distribution
+    sample(**kwargs)
+        Produces a sample of the IHR
+    """
 
     def __init__(
         self,
-        dist: dist.Distribution,
-        varname: str = "IHR",
+        dist: Optional[dist.Distribution] = dist.LogNormal(jnp.log(0.05), 0.05),
```
natemcintosh, line 66:
```suggestion
        dist: dist.Distribution | None = dist.LogNormal(jnp.log(0.05), 0.05),
```
This is not a necessary change, just a suggestion. In 3.10+, you can use this notation instead of `Optional[..]`. I personally find it slightly nicer to read, but it's really not of much consequence. See [docs here](https://docs.python.org/3/library/typing.html#typing.Optional)

natemcintosh (approved):

model/pyproject.toml
gvegayon: I wonder if this should be required for the project, @AFg6K7h4fhy2. Since the package can still run without `numpydoc`. The same applies to the `ipykernel` (which I added). What are your thoughts, @dylanhmorris?

dylanhmorris: Presumably `[tool.poetry.group.docs.dependencies]`, no?
https://python-poetry.org/docs/managing-dependencies/#optional-groups

model/src/pyrenew/latent/hospitaladmissions.py
dylanhmorris: `dist` should default to `None`, not to a specific instantiated distribution (the latter was an artifact of old example code)
``` suggestion
        dist: dist.Distribution | None = None
```

model/pyproject.toml
```diff
@@ -15,8 +15,26 @@ polars = "^0.20.13"
...
+    '\.undocumented_method$',
+    '\.__repr__$',
+    '\.__init__$',
+]
+override_SS05 = [  # override SS05 to allow docstrings starting with these words
+    '^Process ',
+    '^Assess ',
+    '^Access ',
+]
+
```
gvegayon, line 37: Probably part of the `tool.poetry.groups.docs.dependencies` as suggested by @dylanhmorris.

model/src/pyrenew/latent/hospitaladmissions.py
```diff
@@ -14,30 +15,67 @@
...
-        dist: dist.Distribution,
-        varname: str = "IHR",
+        dist: Optional[dist.Distribution] = dist.LogNormal(
+            jnp.log(0.05), 0.05
+        ),
+        varname: Optional[str] = "IHR",
     ) -> None:
-        """Default constructor
+        """
+        Default constructor
```
gvegayon, line 72: Just a question, I like the line breaks for starting docs, but I always see docstring starting right away (no like break after `"""`). What's the best practice here? cc @dylanhmorris and @natemcintosh

model/src/pyrenew/latent/hospitaladmissions.py
```diff
@@ -47,17 +85,43 @@ def __init__(
...
+        AssertionError
+            If the inputted distribution is not a Numpyro distribution.
+        """
         assert isinstance(distr, dist.Distribution)
 
     def sample(self, **kwargs) -> InfectHospRateSample:
+        """
+        Produces a sample of the IHR
+
+        Parameter
```
gvegayon, line 114: Make sure is `Parameters` 😄 

```suggestion
        Parameters
```

gvegayon (commented):
Overall it looks good. Ping me when the comments/discussions are addressed

model/src/pyrenew/latent/hospitaladmissions.py
natemcintosh: If there is a best practice, I'm not aware of it. I always like to have a line break at the start of a docstring, if only to make reading it slightly easier. And the line break is gone when docs are generated (at least when I look at it using the VSCode hover to get docs)

dylanhmorris: I like to start with a line break, and you do see it (although not universally) in public code. Here's an example from the source code for numpydoc itself 

https://github.com/numpy/numpydoc/blob/defebb1d5ffcef5f53c05a15091dbad2118ecd98/numpydoc/validate.py#L157-L174

gvegayon: Thanks for confirming that!

model/src/pyrenew/latent/hospitaladmissions.py
gvegayon: Solved in 898ac6c

model/pyproject.toml
gvegayon: Solved in 898ac6c

model/src/pyrenew/latent/hospitaladmissions.py
```diff
@@ -14,30 +15,67 @@
...
-    """Infection to Hospitalization Rate"""
+    """
+    Infection to Hospitalization Rate
+
+    Methods
+    -------
+    validate(distr)
+        Validates distribution is Numpyro distribution
+    sample(**kwargs)
+        Produces a sample of the IHR
```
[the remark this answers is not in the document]
gvegayon: New issue in #79. Will resolve then.

model/src/pyrenew/latent/hospitaladmissions.py
gvegayon: Addressed in 0710c3a

model/pyproject.toml
gvegayon: Addressed in 0710c3a

gvegayon (approved):
All looks good, @AFg6K7h4fhy2. I took care of a few things were requested and created an issue for another (#79). Will merge this and continue with the others.

PR #77, opened 2024-04-15
Issue 47 - pyrenew_demo.qmd

cshelley:
Completed documentation for pyrenew demo


gvegayon (approved):
Looks great, @cshelley! We should add more details later, but this is better than what we have right now.

cshelley:
Thanks! And agreed - I'd have put a lot more description of each function into this if I had had more time.

PR #74, opened 2024-04-12
First Pass Documentation (docstrings) For Model Folder

AFg6K7h4fhy2:
This pull request includes a first pass 

* Correction and or modifications of existing docstrings in `./model/src/pyrenew/model`
* Addition of new docstrings in `./model/src/pyrenew/model`

Note: You will find some TODO placeholders, mostly in the Notes section of some docstrings. These occur where the author was uncertain about standard procedure or the content.

PR #86, opened 2024-04-19
First Pass Documentation (docstrings) For Deterministic Folder

AFg6K7h4fhy2:
This pull request includes a first pass for:

* Correction and or modifications of existing docstrings in ./model/src/pyrenew/deterministic
* Addition of new docstrings in ./model/src/pyrenew/deterministic

Other relevant PRs: (closed: #73 , #74 , #76 )


model/poetry.lock
gvegayon, line 1: I wonder if `poetry.lock` should always change with PRs? Or better yet, shouldn't the `pyproject.toml` file be the only thing to distribute with the repo? [poetry doesn't have a black/white stance on it](https://python-poetry.org/docs/basic-usage/#committing-your-poetrylock-file-to-version-control). I think poetry.lock should only be distributed for development, but not for other users, furthermore, perhaps we should only have poetry.lock for the actual pipeline implementation (not the pipeline package).

What do you think @AFg6K7h4fhy2 @dylanhmorris @natemcintosh @zsusswein?

natemcintosh: Can you explain the difference between "the actual pipeline implementation (not the pipeline package)."?

gvegayon: Sorry, I meant deployment, the pipeline here is a package that, together with pyrenew, will be used for forecasting. That's what I have in my head.

gvegayon (approved):
Minor issue

gvegayon (approved):

PR #93, opened 2024-04-29
Adding Null RVs

gvegayon:
This PR adds the following three new classes (as instances of `DeterministicRV`):

- `NullVariable`
- `NullProcess`
- `NullObservation`

The PR also changes how observation processes are used in the Rt Infections and Hospitalization models. Instead of allowing to have `None` for the obs. process, it uses `NullObservation` as a default value.

It adds the corresponding tests.


model/src/pyrenew/model/hospitalizations.py
```diff
@@ -87,10 +88,13 @@ def __init__(
             gen_int=gen_int,
             I0=I0,
             latent_infections=latent_infections,
-            observation_process=None,
+            observation_process=NullObservation(),
```
dylanhmorris, line 91: Equivalent but clearer to pass `None` here as well, I think.

Or could allow for an infection observation process (default `None`) to be specified at `HospitalizationsModel` instantiation?

gvegayon: In principle, we could pass observed infections and thus use `observation_process` for the `RtInfectionsModel` within the `HospitalizationsModel`. Do you think we should?

gvegayon: OK, 6047e60 now passes infections during construction (observation process) and when sampling. Both optionals. This would be the first multisignal model :P.

gvegayon:
@AFg6K7h4fhy2 can you take a look at this? Thanks!

model/src/pyrenew/deterministic/nullrv.py
AFg6K7h4fhy2, line 1: For the following (where they occur) does this indicate that other arguments might, at some point, exist (so these `kwargs` statements are placeholders, somewhat)? 

```
**kwargs : dict, optional
    Ignored.
```

model/src/pyrenew/model/hospitalizations.py
AFg6K7h4fhy2, line 1: Flagging on this file that in other PRs, TODOs in docstrings and classless namedtuples have been corrected / modified.

model/src/pyrenew/deterministic/nullrv.py
gvegayon: Yes it does.

PR #94, opened 2024-04-29
Clean Model Documentation

AFg6K7h4fhy2:
This PR does several things for the MSR model documentation. Namely:

- Removes Methods from class docstrings
- Removes TODOs from Notes sections
- Creates classes for namedtuple objects
- Adjusts spelling and fixes grammatical errors
- Uses `any`, `tuple` instead of `Any`, `Tuple` from `typing`
- Removes `Optional`s from `typing`


gvegayon:
OK, @AFg6K7h4fhy2. I fixed it:

- We were using indexing of the tuples in some places. Since you were implementing classes from scratch, the indexing method was unavailable. I solved that using `typing.NamedTuple` as the base class.
- The `NamedTuple` class doesn't allow rewriting `__init__`, so instantiation is now made directly by declaring members and defaulting them to `None`.
- This also required how things were documented. Now, class members are declared as attributes in the docstring of the class.

PR #96, opened 2024-05-02
47 thl2 first pass on documentation

cshelley:
Completed first-pass documention on getting_started.qmd. Still think it would be good to at least point users toward a description of renewal models in general and ours specifically.


gvegayon:
I'll take a look at the conflicts.

gvegayon (approved):
All looks great! I fixed the conflicts (which were minor), and there's no changes to the codebase, so we are good to go (as soon as the CI checks pass)!

PR #107, opened 2024-05-08
Create a dockerfile for running the model code locally or in azure

gvegayon:
- [x] Adds a Dockerfile that builds `pyrenew`.
- [x] Adds a workflow that builds the image and pushes it to `ghcr.io/cdcgov/pyrenew:latest`.
- [x] Extends the `Makefile` by adding a helper function and targets related to the container.


AFg6K7h4fhy2 (approved):
Nothing I might add that could not be added later.

Issue #88, opened 2024-04-22
How to handle the poetry.lock file (should we include it?)


gvegayon:
> I wonder if `poetry.lock` should always change with PRs? Or better yet, shouldn't the `pyproject.toml` file be the only thing to distribute with the repo? [poetry doesn't have a black/white stance on it](https://python-poetry.org/docs/basic-usage/#committing-your-poetrylock-file-to-version-control). I think poetry.lock should only be distributed for development, but not for other users, furthermore, perhaps we should only have poetry.lock for the actual pipeline implementation (not the pipeline package).
> 
> What do you think @AFg6K7h4fhy2 @dylanhmorris @natemcintosh @zsusswein?

_Originally posted by @gvegayon in https://github.com/CDCgov/multisignal-epi-inference/pull/86#discussion_r1572595946_

natemcintosh:
I think using the `poetry.lock` for deploying the code would be good. I am ambivalent about keeping it otherwise

PR #121, opened 2024-05-16
Upx3 typos fix

gvegayon:


model/docs/example-with-datasets.qmd
```diff
@@ -313,15 +313,15 @@ class WeekdayEffect(metaclass.RandomVariable):
...
+        and = npro.sample(
             name="weekday_effect",
             fn=npro.distributions.TruncatedNormal(
                 loc=1.0, scale=.5, low=0.1, high=10.0
                 ),
             sample_shape=(7,)
         )
 
-        return jnp.tile(ans, self.nweeks)[:self.len]
+        return jnp.tile(and, self.nweeks)[:self.len]
```
gvegayon, line 324: I'm thinking, with this kind of "fix," the typos pre-commit hook may not be good enough. I'm afraid of it introducing bugs without us noticing. What do do you, @damonbayer and @dylanhmorris?

poetry.lock
gvegayon, line 1: Also, `poetry.lock` files are a bit of a pain here. Right now, this PR has conflicts with main because of it. Issue #88 deals with that, so we should probably restart the discussion. 

cc @damonbayer @dylanhmorris @natemcintosh

model/docs/example-with-datasets.qmd
gvegayon: Maybe, there's a way of correcting just the docstrings?

cc @natemcintosh

natemcintosh: There are a [couple of options](https://github.com/crate-ci/typos?tab=readme-ov-file#false-positives) for telling `typos` to ignore things. I've used file exclusion before to remove test data with false positives. 

But just docstrings might be harder. We could definitely hard code in `ans` as something to be ignored, but it's not the most ideal, and maybe one day we will want it to catch an `ans` that should have been a `and` in a README somewhere.

Overall, I'd say it might be worth considering moving typos from pre-commit to regular CI runs say once a week, or once a month? Then someone has to come and check the results though ...

dylanhmorris: Catching up here. I wasn't aware that there was a plan to run something that did autocorrect on spelling for source code. My knee-jerk reaction is to be against any such thing, as it seems likely to cause far more problems than it solves.

Why not simply use non-correcting (i.e. just flagging) spell checkers in our local editors, or a report in CI that flags _potential_ spelling mistakes but does not change them?

damonbayer: I feel similarly to Dylan. Big headache for minimal benefit. False negatives are not that bad, and false positives will be a frequency annoyance.

natemcintosh: We can simply change the input args for `typos` to not alter the code. See the bottom paragraph of [this page](https://github.com/crate-ci/typos/blob/master/docs/pre-commit.md).

In theory, we just need to add
`args: []` to the `typos` section of `.pre-commit-config.yaml`

gvegayon:
OK, so I was able to update the branch to main (it had lots of conflicts). Checks pass, except for typos, reporting a false positive (see [here](https://github.com/CDCgov/multisignal-epi-inference/actions/runs/9198864732/job/25302505391#step:4:155)). I don't think pre-commit should break if this happens. We can discuss further next sprint.

cc @dylanhmorris @AFg6K7h4fhy2 @natemcintosh @damonbayer

gvegayon:
Also, it is worthwhile mentioning we have to be mindful about keeping PR branches up to date with main. Having to manually deal with conflicts can be avoided.

gvegayon:
@AFg6K7h4fhy2, I just merged main into this to see how it behaves. Depending on that I'll approve and merge :)! Thanks for your patience with this!

gvegayon:
There are still some corrections to do: load should be old

image

model/pyproject.toml
```diff
@@ -42,6 +42,13 @@ checks = [
     "PR07",
     "RT01"
 ]
+ignore = [
+    "ES01",
+    "SA01",
+    "EX01",
+    "SS06",
+    "RT05"
+]
```
damonbayer, line 51: Why is this being added here?

gvegayon:
@AFg6K7h4fhy2, I tried to add regex to capture `jnp.arange` (which typos tries to correct to `jnp.arrange`), and it is not working properly. That's what's causing pre-commit to fail.

cc @damonbayer

AFg6K7h4fhy2:
Do you know if `arange` is the last failure with the typos PR?

gvegayon:
Thanks, @AFg6K7h4fhy2! But I think the "arange" should be flagged as a typo. I was trying to flag the more specific case of `jnp.arange` or `np.arange` as OK. IMO, adding `arange` to the list of OK words is fine for the moment, but typos should allow us to be more specific.

docs/pyproject.toml
damonbayer, line 1: Is this added automatically? If not, please remove.

docs/pyproject.toml
```diff
@@ -29,6 +29,7 @@ exclude = [  # don't report on objects that match any of these regex
 ]
 
 
+
```
damonbayer, line 32: Is this added automatically? If not, please remove.

damonbayer (approved):

Issue #116, opened 2024-05-10
Exponential growth model for initial infections


gvegayon:
## Goal

Implement the initialization of the infections using an exponential growth model as in the wastewater model (see [here](https://github.com/CDCgov/wastewater-informed-covid-forecasting/blob/61af2bbbd0cc179460b3b06c4fa209b3a50f6101/model_definition.md#renewal-process-for-incident-infections)):

> This process is initialized by estimating an initial exponential growth[^EpiNow2] of infections for 50 days prior to the calibration start time $t_0$:
> 
> $$ I(t) = I_0\exp(rt) $$
> 
> where $I_0$ is the initial per capita infection incident infections and $r$ is the exponential growth rate.

## Context

During discussions of feature prioritization. This particular component was not listed.

## Required features

The feature should be implemented as a `RandomVariable` object part of the `latent` submodule.

## Specifications

- [ ] Mathematical description of the component.
- [ ] Unit test evaluating the feature.
- [ ] Integration test with the basic model and hospital admissions model.
- [ ] Integration as part of a tutorial.

These specs may be moved to a different issue involving other components of the replication of the hospital admissions model in the wastewater project.

## Out of scope

- TBD

gvegayon:
Hey @damonbayer, as we discussed, this could be implemeted via a latent process either extending the current implementation of the infections module or adding a new one called something like `latent.InfectionsWithExpGrowth(RandomVariable)`.

damonbayer:
From my understanding, there are three key times related to this implementation:
* $t_0$: the time where the first latent infection(s) occur
* $t_{\text{seeded}}$: the time at which there is a sufficient infection history such that the renewal process can begin.
* $t_{\text{obs}}$: the time at which the first observation occurs.

I suppose there is no reason that $t_{\text{seeded}}$ couldn't be after $t_{\text{obs}}$, but I think it's appropriate to enforce $t_0 < t_{\text{seeded}} \leq t_{\text{obs}}$.

This issue relates to generating $I(t)$ for $t \in [t_0, t_{\text{seeded}}]$. In the current model, we have $I(t) = I_0$ for $t \in [t_0, t_{\text{seeded}}]$ and we only save $I(t)$ for $t \geq t_{\text{seeded}}$.

**When we have variation in $I(t)$ for $t \in [t_0, t_{\text{seeded}}]$, I assume we will want to keep track of all of these values. Is that correct?**

gvegayon:
[...]

PR #123, opened 2024-05-17
Rt with infection feedback

gvegayon:
This PR:

- [x] Creates the `InfectionsWithFeedback` latent module.
- [x] Adds the new module `pyrenew.datautils` for common data management operations.
- [x] Adds the tutorial `extending_pyrenew.qmd`
- [x] Corrects the order of outputs from the `sample_infections_with_feedback` function.
- [x] Adds tests modules for `InfectionsWithFeedback` and `pyrenew.datautils`.

Extra:

- [x] Uses `jax.numpy.atleast_1d()` for `DeterministicRV.sample()` (so scalars are coerced into 1d arrays)
- [x] Uses `jax.numpy.atleast_1d()` for `DistributionalRV.sample()` (so scalars are coerced into 1d arrays)
- [x] Fixes the reminder of the ghost outbreak.

PR #124, opened 2024-05-19
Rework Initial Infections and Support Exponential Growth Initializations

damonbayer:
Closes https://github.com/CDCgov/multisignal-epi-inference/issues/116.

Requesting early feedback on my very WIP changes, since this is my first contribution, and I'm pretty inexperienced at Python, object-oriented programming, and Numpyro/JAX.

Eventually, this PR will achieve several objectives:

1. Implementation of `InfectionSeedMethod` of which the current only behavior (padding with 0's) will be one method, and an exponential grown initialization will be another. Must meet all specifications documented in https://github.com/CDCgov/multisignal-epi-inference/issues/116.
2. Rewrite of the `Infections0` class to correspond to all of the "seed" infections, rather than just the initial infections (see [here](https://github.com/CDCgov/multisignal-epi-inference/issues/116#issuecomment-2113056295)). This work is currently being done in `model/src/pyrenew/latent/infection_seeding_method.py`. 
3. Update all existing tutorials to work with the new methods.
4. Maybe: write a new tutorial relating to the seeding methods.

Much of the current PR was implemented by pattern matching existing classes and not fully understanding what I was doing. I have littered the code with questions throughout and will self-review to ask more questions.


model/src/pyrenew/latent/infection_seeding_method.py
```diff
@@ -0,0 +1,51 @@
...
+    def validate(n_timepoints: int) -> None:
+        assert isinstance(n_timepoints, int), "n_timepoints must be an integer"
+        assert n_timepoints > 0, "n_timepoints must be positive"
+
+    @abstractmethod
+    def seed_infections(self, I0: ArrayLike) -> ArrayLike:
+        pass
+
+
+class SeedInfectionsZeroPad(InfectionSeedMethod):
```
damonbayer, line 28: This is written to produce the same result as `SeedInfectionsZeroHstack`, but it doesn't work within `sample`. See `scratch/InfectionSeedingProcess_demo.py` for example code.

model/src/pyrenew/latent/infection_seeding_method.py
```diff
@@ -0,0 +1,51 @@
...
+    def seed_infections(self, I0: ArrayLike):
+        return jnp.pad(I0, (self.n_timepoints - I0.size, 0))
```
damonbayer, line 30: `I0` should always be of size 1. Where should I enforce that? Does that mean I should have `I0: float` *or `I0: int` in the function definition?

In other functions that have `I0` as an argument, they expect `ArrayLike`, so maybe this is a NumPyro thing? ex: https://github.com/CDCgov/multisignal-epi-inference/blob/main/model/src/pyrenew/latent/infections.py#L79

model/src/pyrenew/latent/infection_seeding_method.py
```diff
@@ -0,0 +1,51 @@
...
+
+# -*- coding: utf-8 -*-
+# numpydoc ignore=GL08
+"""
+Classes for the infection seeding methods
+"""
+
+
+class InfectionSeedMethod(metaclass=ABCMeta):
+    def __init__(self, n_timepoints: int):
```
damonbayer, line 14: I'm not sure if `InfectionSeedMethod` should have an `I0` attribute or how to implement sampling, if it does.

model/src/pyrenew/latent/infection_seeding_method.py
```diff
@@ -0,0 +1,51 @@
...
+
+class SeedInfectionsRepeat(InfectionSeedMethod):
+    def seed_infections(self, I0: ArrayLike):
+        return jnp.tile(I0, self.n_timepoints)
+
+
+# Defined as below, I don't see how we can sample `rate`, which we probably want to do.
+# Not sure if rate should be ArrayLike or float
+class SeedInfectionsExponential(InfectionSeedMethod):
+    def __init__(self, n_timepoints: int, rate: ArrayLike):
```
damonbayer, line 46: Not sure if rate should be `ArrayLike` or `float`. See my above question about `I0`.

model/src/pyrenew/latent/infection_seeding_process.py
```diff
@@ -0,0 +1,38 @@
...
+        infection_seed_method: InfectionSeedMethod,
+    ) -> None:
+        assert isinstance(I0_dist, dist.Distribution)
+        assert isinstance(infection_seed_method, InfectionSeedMethod)
+
+    def sample(self) -> tuple:
+        I0 = npro.sample("I0", self.I0_dist)
+
+        infection_seeding = self.infection_seed_method.seed_infections(I0)
+        I0_det = npro.deterministic("I0", infection_seeding)
```
damonbayer, line 36: Copying from https://github.com/CDCgov/multisignal-epi-inference/blob/main/model/src/pyrenew/process/rtrandomwalk.py#L123

What is the point of declaring this `deterministic`? It confused me because `I0` is being sampled.

scratch/InfectionSeedingProcess_demo.py
```diff
@@ -0,0 +1,60 @@
...
+# Works:
+with npro.handlers.seed(rng_seed=rng_seed):
+    exp_dat = exp_model.sample()
+exp_dat
+
+
+# Doesn't work:
+with npro.handlers.seed(rng_seed=rng_seed):
+    zero_pad_dat = zero_pad_model.sample()
+zero_pad_dat
```
damonbayer, line 49: This returns a float of length 1, for some reason.

scratch/InfectionSeedingProcess_demo.py
```diff
@@ -0,0 +1,60 @@
...
+
+# Works:
+SeedInfectionsZeroPad(n_timepoints).seed_infections(jnp.array([11.0]))
```
damonbayer, line 52: Using it outside of sample, it works as expected.

damonbayer (commented):
Self-reviewing to ask questions for other reviewers (or future me) to answer.

model/docs/getting-started.qmd
damonbayer, line 1: `model1.run()` currently produces the error `AssertionError: all sites must have unique names but got `I0` duplicated`

model/src/pyrenew/latent/infection_seeding_method.py
damonbayer: We should make `rate` a random variable, so it can be sampled (or deterministic).

model/src/pyrenew/latent/infection_seeding_method.py
damonbayer: Don't do that.

damonbayer:
@gvegayon and I discussed all my questions over Teams. No need review the current state.

model/src/pyrenew/latent/infection_seeding_method.py
```diff
@@ -0,0 +1,54 @@
...
+    @abstractmethod
+    def seed_infections(self, I0: ArrayLike) -> ArrayLike:
+        pass
+
+
+class SeedInfectionsZeroPad(InfectionSeedMethod):
+    def seed_infections(self, I0: ArrayLike):
+        # alternative implementation:
+        # return jnp.hstack([jnp.zeros(self.n_timepoints - I0.size), I0])
+        return jnp.pad(I0, (self.n_timepoints - I0.size, 0))
```
damonbayer, line 35: In this class, `I0` has a different meaning than I generally expect it to have. Usually `I0` would be the first element of the array. In this example, `I0` is the last element.

dylanhmorris: This comes back to this question about nomenclature, I think: https://github.com/CDCgov/multisignal-epi-inference/issues/116#issuecomment-2113092808

model/src/pyrenew/latent/infection_seeding_method.py
dylanhmorris: Noting that we have changed design strategy here to require array-valued `I0` of at least the length of the largest backward looking PMF (generation interval, feedback strength, etc.)

model/src/pyrenew/latent/infection_seeding_process.py
dylanhmorris: `numpyro.deterministic` labels deterministic functions of other parameters for tracking/recording during MCMC. Similar to declaring a `transformed parameter` or deterministic `generated quantity` in Stan. As written we probably don't want this, as it gives the computed quantity the same tracing name ("`I0`") as the upstream directly-sampled quantity from which it is produced.

model/src/pyrenew/latent/infection_seeding_method.py
damonbayer: I have revised names to make things clearer, but the entire point of this `InfectionSeedMethod` class to create `I0`'s that fulfill this requirement. I am calling `I0`-like things that are inputs to `InfectionSeedMethod`s `I_pre_seed` and `I0`-like-things that are inputs to `sample_infections` (which must be longer than the largest backward-looking PMF) `I_seed`.

On the note of naming, sample_infections should probably be renames something like `compute_future_infections`, to avoid suggesting something stochastic is happening.

dylanhmorris: Good point, will make issue.

gvegayon: Addressed in #137

gvegayon:
Ready for you, @damonbayer. Missing docstrings are causing errors, but the tests are passing now.

damonbayer:
Really struggling to get anything remotely sensible without padding: 
image
With padding, things look okay: 
image
Latent infections look a lot more "natural" than with the zero padding we had before: 
image

damonbayer:
@gvegayon @dylanhmorris This is in a good place for a review. It still needs tests for the new functionality. These tests will probably look something like `scratch/InfectionSeedingProcess_demo.py`, which will need to be removed.

gvegayon:
> @gvegayon @dylanhmorris This is in a good place for a review. It still needs tests for the new functionality. These tests will probably look something like `scratch/InfectionSeedingProcess_demo.py`, which will need to be removed.

The website build error should be solved once #144 and #123 are merged. Should we remove the "Draft" status of the PR?

damonbayer:
Doesn't matter to me if it is a "draft" or not. It should not be merged without tests, but I assume there will be other issues that prevent the merging.

damonbayer:
@dylanhmorris @gvegayon Tests are added. This is ready for review.

damonbayer:
Working on resolving merge issues now.

damonbayer:
Noting to myself to address comment in `rtinfectionsrenewalmodel.py`
```python
        # Padding i0 to match gen_int
        # PADDING SHOULD BE REMOVED ONCE
        # https://github.com/CDCgov/multisignal-epi-inference/pull/124
        # is merged.
        # SEE ALSO:
        # https://github.com/CDCgov/multisignal-epi-inference/pull/123#discussion_r1612337288
```

gvegayon:
Houston, we have a problem. Per #143, we acquired the policy that sampling with a particular duration should return arrays of that duration. At the same time, per this PR, we also agreed that baseline infection should always be included in the latent infections (the `hstack`). What happens in the case of convolutions? Right now:

1. `compute_infections_from_rt` returns latent infections of the same length of `Rt_raw`, but those do not include the baseline infections `I0`.
2. The same is true for `compute_infections_from_rt_with_feedback`.

A naive solution would be trimming the tail of the resulting infections, and appending the baseline infections used to compute the time series. But that indirectly discards more recent `Rt` entries. Even then, if we followed that wrong approach, we would still have a problem: What happens when the initial infections have the same length of `Rt`/`Rt_raw`? 

Ideas @dylanhmorris, @damonbayer, @SamuelBrand1, @seabbs?

damonbayer:
> A naive solution would be trimming the tail of the resulting infections, and appending the baseline infections used to compute the time series. But that indirectly discards more recent `Rt` entries. Even then, if we followed that wrong approach, we would still have a problem: What happens when the initial infections have the same length of `Rt`/`Rt_raw`? 


Could we instead prepend the Rt vectors with missing values?

damonbayer:
`compute_infections_from_rt` and `compute_infections_from_rt_with_feedback` should return latent infections including the baseline infections `I0`. Can we prepend the Rt vectors with missing values? When we decide how to handle time more generally, we can come up with a cleaner solution.

dylanhmorris:
I agree with @damonbayer. The fact that we end up with a longer infection timeseries than the `Rt` timeseries is not a bug; it's a consequence of how the renewal process is initialized.

@gvegayon, is your concern that you'd like `infections[ind]` to correspond to the same timepoint as `Rt[ind]` for every index `ind` (at least until we have a general approach to handling model time)?

gvegayon:
Yes, I understand that that's not a bug, but I am not clear about how to handle the situation. Now, I have the following observation:

- The input dataset is of length `n_timepoints`.
- The output latent infections will be of length `len(I0) + n_timepoints`, because all Rt sampling functions under `process` return a vector of length `n_timepoints`
- When we finally do inference using latent infections of length `len(I0) + n_timepoints`, how do we do it to compare it to the original dataset which is of length `n_timepoints`? As I'm writing this, I would imagine that the latent functions should truncate the computed Rt by removing the tail. Since the calculations of latent infections are mechanistic, it should be OK. But I'm not an expert on this!

damonbayer:
Say we have input data of size `n_timepoints` and specify `n_seed_points` for `I0`.

We eventually should coerce everything to vectors of size `n_timepoints + n_seed_points`.

Rt will be `[rep(NA, n_seed_points) + inferred Rt]`.
Data will be `[rep(NA, n_seed_points) + original input data]`.
Infections will be `[I0 + I_computed_from_Rt]`

Let's assume that the seeding period ends when the observations begin, which isn't a necessary assumption, but should be fine for now.

damonbayer:
All tests are passing except `test_model_hosp_with_obs_model_weekday_phosp`. I will keep investigating, but wouldn't mind a lift from @dylanhmorris or @gvegayon, if you have any time.

It has something to do with https://github.com/CDCgov/multisignal-epi-inference/blob/aacd66c7c7b3e55f32e52c3ffb55d86c825dc458/model/src/pyrenew/latent/hospitaladmissions.py#L182-L189

Where `predicted_admissions` is changing sizes, so it cannot always match the size of `self.weekday_effect_dist.sample(**kwargs)[0]`.

damonbayer:
All tests are passing on my local machine. Not sure why they don't pass in CI.

This PR is already too big, but I think my experience here points to some much needed revisions to the Hospital Admissions code, tutorials, and tests. It's very hard to tell if things are working correctly because they don't produce realistic output. I don't know whether to attribute that to bugs or poorly-chosen inputs.

dylanhmorris:
Some argument we shouldn't have the plotting snapshot tests at all.

damonbayer:
> Some argument we shouldn't have the plotting snapshot tests at all.

The other weird issue was some numerical precision thing again. Two arrays are exactly equal on my machine, while they weren't even almost equal in CI.

model/src/pyrenew/model/admissionsmodel.py
```diff
@@ -250,28 +251,35 @@ def sample(
...
+                sampled_obs, *_ = self.sample_admissions_process(
+                    predicted=latent[i0_size + padding :],
+                    observed_admissions=observed_admissions[
+                        i0_size + padding :
+                    ],
                     **kwargs,
                 )
-        else:
-            sampled = None
+            # this is to accommodate the current version of test_model_hosp_no_obs_model. Not sure if we want this behavior
```
damonbayer, line 276: Just calling attention to this comment

dylanhmorris: In general seeing this many switches suggests to me that this can and should be refactored and modularized. But separate PR I think. Let me know if you disagree.

model/src/pyrenew/latent/infection_seeding_process.py
```diff
@@ -0,0 +1,75 @@
+# -*- coding: utf-8 -*-
+# numpydoc ignore=GL08
+import numpyro as npro
+from pyrenew.latent.infection_seeding_method import InfectionSeedMethod
+from pyrenew.metaclass import RandomVariable
+
+
+class InfectionSeedingProcess(RandomVariable):
+    """Generate an initial infectioon history"""
```
dylanhmorris, line 9: Typo
```suggestion
    """Generate an initial infection history"""
```

model/src/pyrenew/latent/infection_seeding_method.py
```diff
@@ -0,0 +1,175 @@
...
+        """
+        if I_pre_seed.size != self.n_timepoints:
+            raise ValueError(
+                f"I_pre_seed must have the same size as n_timepoints. Got I_pre_seed of size {I_pre_seed.size} and and n_timepoints of size {self.n_timepoints}."
+            )
+        return jnp.array(I_pre_seed)
+
+
+class SeedInfectionsExponential(InfectionSeedMethod):
+    """Generate seed infections according to exponential growth.
```
dylanhmorris, line 119: Needs to be a raw string or backslashes need to be escaped.
```suggestion
    r"""Generate seed infections according to exponential growth.
```

model/src/pyrenew/latent/infection_seeding_process.py
```diff
@@ -0,0 +1,75 @@
...
+        Returns
+        -------
+        None
+        """
+        InfectionSeedingProcess.validate(I_pre_seed_rv, infection_seed_method)
+
+        self.I_pre_seed_rv = I_pre_seed_rv
+        self.infection_seed_method = infection_seed_method
+
+        return None
```
dylanhmorris, line 34:
```suggestion
```

model/src/pyrenew/latent/infection_seeding_process.py
```diff
@@ -0,0 +1,75 @@
...
+        """Sample the infection seeding process.
+
+        Returns
+        -------
+        tuple
+            a tuple where the only element is an array with the number of seeded infections at each time point.
+        """
+        (I_pre_seed,) = self.I_pre_seed_rv.sample()
+        infection_seeding = self.infection_seed_method(I_pre_seed)
+        npro.deterministic("infection_seeding", infection_seeding)
```
dylanhmorris, line 73: I'd give this RV a `name` and pass that to `npro.deterministic` to avoid variable name clashes if the user instantiates two `InfectionSeedingProcess` RVs

model/src/pyrenew/latent/infections.py
```diff
@@ -103,6 +103,10 @@ def sample(
         InfectionsSample
             Named tuple with "infections".
         """
+        if I0.size < gen_int.size:
+            raise ValueError(
+                f"Initial infections must be at least as long as the generation interval. Got {I0.size} initial infections and {gen_int.size} generation interval."
+            )
```
dylanhmorris, line 109: This duplicates the lines right below (113-117), but I prefer doing the check here and removing the below.
Then  `recent_I0 = I0[-gen_int_rev.size :]` doesn't need to be wrapped in an `else`.

Also makes it more consistent with `InfectionsWithFeedback.sample()`. Which implies that for DRY purposes we should actually have a function that performes this bounds checking and subsetting. Separate PR though.

model/src/pyrenew/latent/infectionswithfeedback.py
```diff
@@ -37,6 +37,16 @@ class InfectionsWithFeedback(RandomVariable):
     This class computes infections, given Rt, initial infections, and generation
     interval.
 
+    Parameters
+    ----------
+    infection_feedback_strength : RandomVariable
+        Infection feedback strength.
+    infection_feedback_pmf : RandomVariable
+        Infection feedback pmf.
+    infections_mean_varname : str, optional
```
dylanhmorris, line 46: Making a note that these varnames should not have the word `mean` in them, but that's a separate PR.

model/src/pyrenew/latent/infectionswithfeedback.py
```diff
@@ -129,7 +139,7 @@ def sample(
             Reproduction number.
         I0 : ArrayLike
             Initial infections, as an array
-            at least as long as the
+            of the same length as the generation
```
dylanhmorris, line 142: It _should_ be  "at least as long", no, given that we subset `I0` appropriately below?

model/src/test/test_infection_seeding_method.py
```diff
@@ -0,0 +1,79 @@
...
+        rate * (np.arange(n_timepoints) - t_pre_seed)
+    )
+    testing.assert_array_almost_equal(
+        infections_custom_t_pre_seed,
+        infections_custom_t_pre_seed_manual,
+        decimal=5,
+    )
+
+    assert infections_custom_t_pre_seed[t_pre_seed] == I_pre_seed
+    return None
```
dylanhmorris, line 50:
```suggestion
```

model/src/test/test_infection_seeding_method.py
```diff
@@ -0,0 +1,79 @@
...
+    I_pre_seed_RV = DeterministicVariable(10.0, name="I_pre_seed_RV")
+    (I_pre_seed,) = I_pre_seed_RV.sample()
+
+    infections = SeedInfectionsZeroPad(n_timepoints).seed_infections(
+        I_pre_seed
+    )
+    testing.assert_array_equal(
+        infections, np.pad(I_pre_seed, (n_timepoints - I_pre_seed.size, 0))
+    )
+    return None
```
dylanhmorris, line 66:
```suggestion
```

model/src/test/test_infection_seeding_method.py
```diff
@@ -0,0 +1,79 @@
...
+    """Check that the SeedInfectionsFromVec class generates the correct number of infections at each time point."""
+    n_timepoints = 10
+    I_pre_seed_RV = DeterministicVariable(np.arange(10), name="I_pre_seed_RV")
+    (I_pre_seed,) = I_pre_seed_RV.sample()
+
+    infections = SeedInfectionsFromVec(n_timepoints).seed_infections(
+        I_pre_seed
+    )
+    testing.assert_array_equal(infections, I_pre_seed)
+    return None
```
dylanhmorris, line 79:
```suggestion
```

model/src/test/test_infection_seeding_method.py
dylanhmorris, line 1: No need for an explicit `return None` at the end of a `void`-like function (e.g. a pytest test)

model/src/test/test_infection_seeding_method.py
```diff
@@ -0,0 +1,79 @@
...
+        infections_custom_t_pre_seed,
+        infections_custom_t_pre_seed_manual,
+        decimal=5,
+    )
+
+    assert infections_custom_t_pre_seed[t_pre_seed] == I_pre_seed
+    return None
+
+
+def test_seed_infections_zero_pad():
```
dylanhmorris, line 53: add a case here where we zero-pad a non-scalar `I_pre_seed_RV`?

model/src/test/test_infection_seeding_method.py
```diff
@@ -0,0 +1,79 @@
...
+from pyrenew.deterministic import DeterministicVariable
+from pyrenew.latent import (
+    SeedInfectionsExponential,
+    SeedInfectionsFromVec,
+    SeedInfectionsZeroPad,
+)
+
+
+def test_seed_infections_exponential():
+    """Check that the SeedInfectionsExponential class generates the correct number of infections at each time point."""
```
dylanhmorris, line 13: have a test that an error is raised if `I_pre_seed_RV` is not a scalar as required by `SeedInfectionsExponential`?

model/src/test/test_infection_seeding_process.py
```diff
@@ -0,0 +1,40 @@
...
+    vec_model = InfectionSeedingProcess(
+        DeterministicVariable(jnp.arange(n_timepoints), name="I0"),
+        SeedInfectionsFromVec(n_timepoints),
+    )
+
+    for model in [zero_pad_model, exp_model, vec_model]:
+        with npro.handlers.seed(rng_seed=1):
+            model.sample()
+
+    return None
```
dylanhmorris, line 40:
```suggestion
```

dylanhmorris (changes requested):
Looks very good. A few changes, all small but some critical.

gvegayon:
To regenerate the image, I did the following:

```bash
cd model
poetry shell
make test_images
```

model/src/pyrenew/latent/infection_seeding_process.py
damonbayer: addressed in https://github.com/CDCgov/multisignal-epi-inference/pull/124/commits/ef3203c1ecd6534398e18fcb9e30091708c13ab4

model/src/pyrenew/latent/infections.py
damonbayer: addressed in https://github.com/CDCgov/multisignal-epi-inference/pull/124/commits/7a453a964466c0c27967a989e8fd7d673ad978ce

model/src/pyrenew/latent/infectionswithfeedback.py
damonbayer: addressed in https://github.com/CDCgov/multisignal-epi-inference/pull/124/commits/cf01560b46e8603bb161e995074b0fa64615cf0d

model/src/test/test_infection_seeding_method.py
damonbayer: addressed in https://github.com/CDCgov/multisignal-epi-inference/pull/124/commits/20e7b6748948c618f1a1162cea95cfb33e3cb400

model/src/test/test_infection_seeding_method.py
damonbayer: addressed in https://github.com/CDCgov/multisignal-epi-inference/pull/124/commits/50b91e4f54585f290888b686c6d9f53478cf11f8

AFg6K7h4fhy2 (approved):

model/src/pyrenew/datautils.py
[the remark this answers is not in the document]
damonbayer: addressed in https://github.com/CDCgov/multisignal-epi-inference/pull/124/commits/c65af09dc2f86c44b95dc83f23d40c322c20d7eb

PR #131, opened 2024-05-22
Weekly Rt with autoregressive difference

gvegayon:
This PR adds the following:

- [x] A new class for the `process` module: `RtWeeklyDiff`.
  - The `RandomVariable` samples from an autoregressive process of the following form: $\log Rt(t) \sim Normal(\log Rt(t - 1) + \beta (\log Rt(t-1) - \log Rt(t-2)), \sigma^2_r)$
  - The sampled Rt is then extended to match the length of the data.
  - The class allows starting from any day of the week.
  - $\beta$, $\sigma$, and the initialization of Rt (Rt(0) and Rt(1)) are provided as `RandomVariable`s.
- [x] Adds tests validating the behavior of the class.

### DIFF
--- a/model/src/pyrenew/latent/infections.py
+++ b/model/src/pyrenew/latent/infections.py
@@ -14,7 +14,7 @@ from pyrenew.metaclass import RandomVariable
 
 class InfectionsSample(NamedTuple):
     """
-    A container for holding the output from the InfectionsSample.
+    A container for holding the output from the `latent.Infections.sample()`.
 
     Attributes
     ----------
--- a/model/src/pyrenew/process/rtweeklydiff.py
+++ b/model/src/pyrenew/process/rtweeklydiff.py
@@ -1,0 +1,201 @@
+from typing import NamedTuple
+
+import jax.numpy as jnp
+import numpyro as npro
+import numpyro.distributions as dist
+from jax import lax
+from jax.typing import ArrayLike
+from pyrenew.metaclass import Model, RandomVariable, _assert_sample_and_rtype
+
+
+class RtWeeklyDiffSample(NamedTuple):
+    """
+    A container for holding the output from the `process.RtWeeklyDiff.sample()`.
+
+    Attributes
+    ----------
+    rt : ArrayLike
+        The sampled Rt.
+    """
+
+    rt: ArrayLike | None = None
+
+    def __repr__(self):
+        return f"RtWeeklyDiffSample(rt={self.rt})"
+
+
+class RtWeeklyDiff(Model):
+    r"""
+    Weekly Rt with autoregressive difference.
+
+    Notes
+    -----
+    This class samples a weekly Rt with autoregressive difference. The
+    mathematical formulation is given by:
+
+    .. math::
+        \log[\mathcal{R}^\mathrm{u}(t_3)] \sim \mathrm{Normal}\left(\log[\mathcal{R}^\mathrm{u}(t_2)] \
+            + \beta \left(\log[\mathcal{R}^\mathrm{u}(t_2)] - \
+             \log[\mathcal{R}^\mathrm{u}(t_1)]\right), \sigma_r \right)
+
+    where :math:`\mathcal{R}^\mathrm{u}(t)` is the weekly reproduction number
+    at week :math:`t`, :math:`\beta` is the autoregressive parameter, and
+    :math:`\sigma_r` is the standard deviation of the noise.
+    """
+
+    def __init__(
+        self,
+        n_obs: int,
+        weekday_data_starts: int,
+        log_rt_prior: RandomVariable,
+        autoreg: RandomVariable,
+        sigma_r: RandomVariable,
+        site_name: str = "rt_weekly_diff",
+    ) -> None:
+        """
+        Default constructor for RtWeeklyDiff class.
+
+        Parameters
+        ----------
+        n_obs : int
+            Number of observations.
+        weekday_data_starts : int
+            Weekday data starts, must be between 0 and 6, 0 beign Sunday.
+        log_rt_prior : RandomVariable
+            Log Rt prior for the first two observations.
+        autoreg : RandomVariable
+            Autoregressive parameter.
+        sigma_r : RandomVariable
+            Standard deviation of the noise.
+        site_name : str, optional
+            Name of the site. Defaults to "rt_weekly_diff".
+
+        Returns
+        -------
+        None
+        """
+
+        self.validate(
+            n_obs,
+            weekday_data_starts,
+            log_rt_prior,
+            autoreg,
+            sigma_r,
+        )
+
+        self.n_obs = n_obs
+        self.weekday_data_starts = weekday_data_starts
+        self.n_weeks = jnp.ceil(n_obs / 7).astype(int)
+        self.log_rt_prior = log_rt_prior
+        self.autoreg = autoreg
+        self.sigma_r = sigma_r
+        self.site_name = site_name
+
+        return None
+
+    @staticmethod
+    def validate(
+        n_obs: int,
+        weekday_data_starts: int,
+        prior: any,
+        autoreg: any,
+        sigma_r: any,
+    ) -> None:
+        """
+        Validate the input parameters.
+
+        Parameters
+        ----------
+        n_obs : int
+            Number of observations.
+        weekday_data_starts : int
+            Weekday data starts.
+        prior : any
+            Log Rt prior for the first two observations.
+        autoreg : any
+            Autoregressive parameter.
+        sigma_r : any
+            Standard deviation of the noise.
+
+        Returns
+        -------
+        None
+        """
+
+        # Nweeks should be a positive integer
+        assert n_obs > 0, f"n_obs should be a positive integer. It is {n_obs}."
+
+        # Weekday data starts should be a positive integer between 0 and 6
+        assert 0 <= weekday_data_starts <= 6, (
+            "weekday_data_starts should be a positive integer between 0 and 6."
+            + f"It is {weekday_data_starts}."
+        )
+
+        _assert_sample_and_rtype(prior)
+        _assert_sample_and_rtype(autoreg)
+        _assert_sample_and_rtype(sigma_r)
+
+        return None
+
+    def sample(
+        self,
+        duration: int | None = None,
+        **kwargs,
+    ) -> RtWeeklyDiffSample:
+        """
+        Samples the weekly Rt with autoregressive difference.
+
+        Parameters
+        ----------
+        duration : int, optional
+            Duration of the sequence. Defaults to None, in which case it is
+            set to the number of observations (`self.n_obs`).
+        **kwargs : dict, optional
+            Additional keyword arguments passed through to internal sample()
+            calls, should there be any.
+
+        Returns
+        -------
+        RtWeeklyDiffSample
+            Named tuple with "rt".
+        """
+
+        # Initial sample
+        log_rt_prior = self.log_rt_prior.sample(**kwargs)[0]
+        b = self.autoreg.sample(**kwargs)[0]
+        s_r = self.sigma_r.sample(**kwargs)[0]
+
+        # Sample noise
+        noise = npro.sample(
+            self.site_name + "_error",
+            dist.Normal(0, s_r),
+            sample_shape=(self.n_weeks,),
+        )
+
+        # Building the scanner
+        def _rt_scanner(log_rts, sigma):
+            next_log_rt = log_rts[1] + b * (log_rts[1] - log_rts[0]) + sigma
+            return jnp.hstack([log_rts[1:], next_log_rt]), next_log_rt
+
+        # Scanning
+        _, log_rt = lax.scan(
+            f=_rt_scanner,
+            init=log_rt_prior,
+            xs=noise,
+        )
+
+        # Expanding according to the number of days
+        if duration is None:
+            duration = self.n_obs
+        elif duration > self.n_obs:
+            raise ValueError(
+                f"Duration should be less than or equal to {self.n_obs}."
+            )
+
+        return RtWeeklyDiffSample(
+            rt=jnp.repeat(jnp.exp(log_rt.flatten()), 7)[
+                self.weekday_data_starts : (
+                    self.weekday_data_starts + duration
+                )
+            ],
+        )
--- a/model/src/pyrenew/process/__init__.py
+++ b/model/src/pyrenew/process/__init__.py
@@ -5,6 +5,7 @@
 from pyrenew.process.ar import ARProcess
 from pyrenew.process.firstdifferencear import FirstDifferenceARProcess
 from pyrenew.process.rtrandomwalk import RtRandomWalkProcess
+from pyrenew.process.rtweeklydiff import RtWeeklyDiff
 from pyrenew.process.simplerandomwalk import SimpleRandomWalkProcess
 
 __all__ = [
@@ -12,4 +13,5 @@ __all__ = [
     "FirstDifferenceARProcess",
     "RtRandomWalkProcess",
     "SimpleRandomWalkProcess",
+    "RtWeeklyDiff",
 ]
--- a/model/src/test/test_rtweeklydiff.py
+++ b/model/src/test/test_rtweeklydiff.py
@@ -1,0 +1,93 @@
+"""
+Test the rtweeklydiff module
+"""
+
+import jax.numpy as jnp
+import numpy as np
+import numpyro as npro
+from numpy.testing import assert_array_equal
+from pyrenew.deterministic import DeterministicVariable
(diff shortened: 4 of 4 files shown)


dylanhmorris (commented):
I would strongly suggest reusing `FirstDifferenceAR` in the implementation here rather than re-implementing the stochastic process.

model/src/pyrenew/process/rtweeklydiff.py
```diff
@@ -0,0 +1,222 @@
...
+            dist.Normal(0, s_r),
+            sample_shape=(self.n_weeks,),
+        )
+
+        # Running the process
+        _, log_rt = lax.scan(
+            f=self.autoreg_process,
+            init=jnp.hstack([log_rt_prior, b]),
+            xs=noise,
+        )
```
gvegayon, line 206: I'm guessing `FirstDifferenceAR` would replace this bit? But I'm unsure how it would go. LMK if you have any time to discuss, @dylanhmorris.

model/src/pyrenew/process/rtweeklydiff.py
gvegayon, line 1: `rtperiodicdiff.py` by @damonbayer

model/src/pyrenew/process/rtweeklydiff.py
dylanhmorris: Discussed sync, but you'd use it for lines 194-206

model/src/pyrenew/process/rtweeklydiff.py
gvegayon: Implemented in c301316

model/src/pyrenew/process/rtperiodicdiff.py
```diff
@@ -203,11 +203,12 @@ def sample(
         )
 
         # Running the process
-        _, log_rt = lax.scan(
-            f=self.autoreg_process,
-            init=jnp.hstack([log_rt_prior, b]),
-            xs=noise,
-        )
+        ar_diff = FirstDifferenceARProcess(autoreg=b, noise_sd=noise)
```
dylanhmorris, line 206: `noise_sd` should be `s_r`, not `noise`. `noise no longer needs to be sampled separately (it'll be sampled under the hood in `arr_diff`)

gvegayon: Addressed in fe82dd1

model/src/pyrenew/process/rtweeklydiff.py
gvegayon: Added in 0bf05cf (and fixed later on)

AFg6K7h4fhy2 (dismissed):
.

model/src/pyrenew/latent/infections.py
AFg6K7h4fhy2, line 1: Removing "the" makes more sense here.

model/src/pyrenew/latent/infections.py
```diff
@@ -14,7 +14,7 @@
 
 class InfectionsSample(NamedTuple):
     """
-    A container for holding the output from the InfectionsSample.
+    A container for holding the output from the `latent.Infections.sample()`.
```
AFg6K7h4fhy2, line 17: Removing "the" makes more sense here.

model/src/pyrenew/process/firstdifferencear.py
```diff
@@ -67,7 +67,9 @@ def sample(
         tuple
         """
         rocs, *_ = self.rate_of_change_proc.sample(
```
damonbayer, line 69:
```suggestion
        rates_of_change, *_ = self.rate_of_change_proc.sample(
```

model/src/pyrenew/process/firstdifferencear.py
```diff
@@ -67,7 +67,9 @@ def sample(
...
-            duration, inits=init_rate_of_change, name=name + "_rate_of_change"
+            duration=duration,
+            inits=jnp.atleast_1d(init_rate_of_change),
+            name=name + "_rate_of_change",
         )
         return (init_val + jnp.cumsum(rocs.flatten()),)
```
damonbayer, line 74:
```suggestion
        return (init_val + jnp.cumsum(rates_of_change.flatten()),)
```

model/src/pyrenew/process/rtperiodicdiff.py
```diff
@@ -0,0 +1,270 @@
...
+        sigma_r: RandomVariable,
+        site_name: str = "rt_periodic_diff",
+    ) -> None:
+        """
+        Default constructor for RtPeriodicDiffProcess class.
+
+        Parameters
+        ----------
+        n_obs : int
+            Number of observations.
```
damonbayer, line 61: observations of what?

model/src/pyrenew/process/rtperiodicdiff.py
```diff
@@ -0,0 +1,270 @@
...
+        data_starts : int
+            Relative point at which data starts, must be between 0 and
+            period_size - 1.
+        log_rt_prior : RandomVariable
+            Log Rt prior for the first two observations.
+        autoreg : RandomVariable
+            Autoregressive parameter.
+        sigma_r : RandomVariable
```
damonbayer, line 69: suggest changing to something more immediately understandable `rt_periodic_diff_sd` or similar

model/src/pyrenew/process/rtperiodicdiff.py
```diff
@@ -0,0 +1,270 @@
...
+        """
+
+        # Nweeks should be a positive integer
+        assert n_obs > 0, f"n_obs should be a positive integer. It is {n_obs}."
+
+        # Data starts should be a positive integer
+        assert 0 <= data_starts, (
+            "data_starts should be a positive integer."
+            + f"It is {data_starts}."
+        )
```
damonbayer, line 135: Check if `n_obs` is actually an int. Check if `n_obs <= period_size - 1`

model/src/pyrenew/process/rtperiodicdiff.py
```diff
@@ -0,0 +1,270 @@
...
+            init_val=log_rt_prior[1],
+            init_rate_of_change=log_rt_prior[1] - log_rt_prior[0],
+        )[0]
+
+        # Expanding according to the number of days
+        if duration is None:
+            duration = self.n_obs
+        elif duration > self.n_obs:
+            raise ValueError(
+                f"Duration should be less than or equal to {self.n_obs}."
```
damonbayer, line 209:
```suggestion
                f"Duration should be less than or equal to n_obs ({self.n_obs})."
```

model/src/pyrenew/process/rtperiodicdiff.py
gvegayon, line 61:
```suggestion
            Size of the returned sample.
```

gvegayon: Solved in e0c64c0

model/src/pyrenew/process/rtperiodicdiff.py
gvegayon: Addressed in e0c64c0

model/src/pyrenew/process/rtperiodicdiff.py
gvegayon: Addressed in e0c64c0

model/src/pyrenew/process/rtperiodicdiff.py
gvegayon: `n_obs` can be smaller than the period. I just added a test, and it works OK. I also added a test on `validate` that checks for integers.

ref: aab6f6e

model/src/pyrenew/process/rtperiodicdiff.py
dylanhmorris: Suggest naming the argument `n_timepoints` for clarity

model/src/pyrenew/latent/hospitaladmissions.py
```diff
@@ -14,7 +14,7 @@
 
 class HospAdmissionsSample(NamedTuple):
     """
-    A container for holding the output from HospAdmissionsSample.sample.
+    A container for holding the output from `lantent.HospAdmissions.sample()`.
```
AFg6K7h4fhy2, line 17: "latent" is spelled incorrectly.

"from [the] ..." works better.

model/src/pyrenew/process/rtperiodicdiff.py
```diff
@@ -207,10 +209,11 @@ def sample(
...
         # Expanding according to the number of days
         if duration is None:
-            duration = self.n_obs
-        elif duration > self.n_obs:
+            duration = self.n_timepoints
+        elif duration > self.n_timepoints:
             raise ValueError(
-                f"Duration should be less than or equal to n_obs ({self.n_obs})."
+                "Duration should be less than or equal "
+                f"to n_timepoints ({self.n_timepoints })."
```
dylanhmorris, line 216: It appears that there's both a `self.n_timepoints` class attributes and also a `duration` passed as a mandatory argument to `self.sample()`. Should that definitely be the case?

model/src/pyrenew/process/rtperiodicdiff.py
damonbayer: Sorry, I meant to check these for `data_starts`:

Check if `data_starts` is actually an int. Check if `data_starts <= period_size - 1`

model/src/pyrenew/process/rtperiodicdiff.py
```diff
@@ -0,0 +1,277 @@
...
+            Standard deviation of the noise.
+        site_name : str, optional
+            Name of the site. Defaults to "rt_weekly_diff".
+
+        Returns
+        -------
+        None
+        """
+
+        assert 0 <= data_starts <= 6, (
```
damonbayer, line 263: This could be a generic check for `data_starts <= period_size - 1`, which could be handled in `RtPeriodicDiffProcess`.

model/src/pyrenew/latent/hospitaladmissions.py
dylanhmorris: I'd omit the "the" and keep the function call notation. Perhaps:

```suggestion
    A container to hold the output of `latent.HospAdmissions.sample()`.
```

model/src/pyrenew/process/rtperiodicdiff.py
```diff
@@ -0,0 +1,277 @@
...
+        self.periodic_diff_sd = periodic_diff_sd
+        self.site_name = site_name
+
+        return None
+
+    @staticmethod
+    def validate(
+        n_timepoints: int,
+        data_starts: int,
+        prior: any,
```
AFg6K7h4fhy2, line 102: Minor note. On line 91, we have "log_rt_prior", whereas validate takes in `prior`. This makes sense because the prior doesn't necessarily have to be a `log_rt_prior`, it just has to pass what's in `validate`. However, the docstring on line 117 supports the interpretation of the prior as `log_rt_prior`, and the other parameters ingested by `validate` are unchanged in name, yet could be (e.g., `n_timepoints` to `ts`). Having consistency by having `validate` take in `log_rt_prior` (what is expected to be one) in place of `prior` might be good.

model/src/pyrenew/process/rtperiodicdiff.py
damonbayer: Not sure I understand your question. Perhaps we can address during coworking time today.

model/src/pyrenew/process/rtperiodicdiff.py
gvegayon: When I implemented this, I had in mind an error I saw about `jit` and objects sizes. When sizing/filtering as a function of a sampled quantity, jax.jit was not happy. This is not the case, so I removed `n_timepoints` in favor of `duration` in 4e4a12e.

model/src/pyrenew/process/rtperiodicdiff.py
gvegayon: Good point, addressed in 4e4a12e.

model/src/pyrenew/process/rtperiodicdiff.py
AFg6K7h4fhy2: To address this question in a more concrete way, Do we want `assert` statements and other non-test checks as soon as the variables are _seen_ or only right before they are used.

model/src/pyrenew/process/rtperiodicdiff.py
```diff
@@ -0,0 +1,266 @@
...
+            duration=n_periods,
+            init_val=log_rt_prior[1],
+            init_rate_of_change=log_rt_prior[1] - log_rt_prior[0],
+        )[0]
+
+        return RtPeriodicDiffProcessProcessSample(
+            rt=jnp.repeat(jnp.exp(log_rt.flatten()), self.period_size)[
+                self.data_starts : (self.data_starts + duration)
+            ],
+        )
```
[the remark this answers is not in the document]
gvegayon: A possibility is directly indexing the result from `pyrenew.models.RtInfectionsRenewalModel.sample()`, e.g.,:

```python
Rt, *_ = self.sample_rt(...)[:n_timepoints]
```

See [here](https://github.com/CDCgov/multisignal-epi-inference/blob/77dc81960a7540e2ad3df5cd889cf1ada37cb81a/model/src/pyrenew/model/rtinfectionsrenewalmodel.py#L276-L279). The problem with this approach, we would be sampling RVs that are discarded afterwards, so not sure what would be the behavior of the MCMC algorithm here.

model/src/pyrenew/process/rtperiodicdiff.py
```diff
@@ -0,0 +1,277 @@
...
+        """
+
+        # Nweeks should be a positive integer
+        assert isinstance(
+            n_timepoints, int
+        ), f"n_timepoints  should be an integer. It is {type(n_timepoints )}."
+
+        assert (
+            n_timepoints > 0
+        ), f"n_timepoints  should be a positive integer. It is {n_timepoints }."
```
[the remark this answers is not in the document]
gvegayon: In theory, there is no ceiling, so I don't see why we would add something like that.

model/src/pyrenew/process/rtperiodicdiff.py
gvegayon: Good point

model/src/pyrenew/process/rtperiodicdiff.py
gvegayon: Addressed in 4e4a12e and 963baed.

model/src/pyrenew/process/rtperiodicdiff.py
gvegayon: Addressed in 439367c

model/src/pyrenew/latent/hospitaladmissions.py
gvegayon: Addressed in 439367c

model/src/pyrenew/process/rtperiodicdiff.py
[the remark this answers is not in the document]
dylanhmorris: A number of array creation functions (deterministic or stochastic) exist in the project. I think we should standardize on any timeseries/sequence `sample(duration=n)` or `compute(duration=n)` call:

- returning an array (which I'll call $y(t)$ ) of exactly length `n`. 
- If $x \ge 1$ initial values are provided, including them as the first $x \le$ `n` entries in $y(t)$

So I agree that if `pyrenew.process.ARProcess.sample(duration=n)` currently returns an array of length `n+1` that includes the inits, that should be changed so it returns an array of length `n` (but still includes the inits)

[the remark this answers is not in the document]
gvegayon: To make sure all is cross-referenced, PR #144 to this branch fixes the issue.

gvegayon:
Working on the conflicts

gvegayon:
Ready for review, @dylanhmorris and @damonbayer .

damonbayer (commented):
Looks good, but I had one question about `duration`. Why did we opt for this instead of `n_timepoints`? They appear to represent the same concept, but maybe I am missing something.

gvegayon:
> Looks good, but I had one question about `duration`. Why did we opt for this instead of `n_timepoints`? They appear to represent the same concept, but maybe I am missing something.

I am a bit fuzzy about this. The `duration` keyword is used for processes only, while `n_timepoints` for models. I don't have strong feelings about it, but `duration` seemed like a natural name for processes, so I didn't want to change that. But the number of observations in a model didn't sound like `duration`, so I opted for using `n_timepoints`. Perhaps this could be homogenized to `length` or something? I suggest continuing this discussion in an issue.

dylanhmorris:
Happy to standardize on `n_timepoints`. Probably a separate PR.

damonbayer (approved):

dylanhmorris (approved):

PR #138, opened 2024-05-28
New metaclass.DistributionalRV

gvegayon:
This PR does the following:

- Creates the class `pyrenew.metaclass.DistributionalRV(RandomVariable)`.
- Refactors `latent.I0(RandomVariable)` to `latent.I0(DistributionalRV)`, saving lines of code.
- Refactors `latent.InfectHospRate(RandomVariable)` to `latent.InfectHospRate(DistributionalRV)`, saving lines of code.
- Updates tests and tutorials accordignly.


model/docs/getting-started.qmd
```diff
@@ -98,6 +98,13 @@ latent_infections = Infections()
 observation_process = PoissonObservation()
 ```
 
+Notice that `Infections0` is an instance of the `DistributionalRV(RandomVariable)` class; therefore, the following could also be used:
```
damonbayer, line 101: I would prefer to remove the `latent.I0` and `latent.InfectHospRate` classes altogether and have this be the only way to do things.

gvegayon: Sure. Confirm this, @dylanhmorris?

dylanhmorris: Happy for it to be removed for now. A couple thoughts:

1. This comes back to the "Wrapper classes for readability y/n?" question we were discussing earlier. In the case of simple scalar parameters I'm not sure the wrapper class is needed.
2. In the future, one might imagine wanting infrastructure for dealing, e.g., with static versus time-varying IHRs. In that scenario, an/some `InfectHospRate` class(es) might be the appropriate solution. But as those would _not_ all be simple wrappers of `DistributionalRV`, again I think removing the class for now makes sense.

gvegayon: 1. Yes, in this case, it is beyond readability. All objects passed to `Model` need a `sample` method. `numpyro.distributions.Distribution` does not. So we have the wrapper class that allows you to do so.
2. Absolutely, in the future we could have more complex methods, that's what I had in mind when I first created these classes, but we can revisit when the time comes.

model/src/pyrenew/metaclass.py
```diff
@@ -129,6 +131,99 @@ def validate(**kwargs) -> None:
...
+                f"numpyro.distributions.Distribution, got {dist}"
+            )
+
+        return None
+
+    def sample(
+        self,
+        obs: ArrayLike | None = None,
+        **kwargs,
+    ) -> tuple:
```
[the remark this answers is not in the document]
gvegayon: No, just missed it :P

model/docs/getting-started.qmd
dylanhmorris: Let's make this a separate issue so we can land this PR.

gvegayon: Addressed by c95d031 and 7ef0e55.

dylanhmorris (approved):
LGTM

PR #140, opened 2024-05-29
Refactor transform module to wrap `numpyro.distributions.transforms`

gvegayon:
- [x] Removes the module, tests, and documentation section on the sphinx website.
- [x] Updates the functions using the module and replaces them with `numpyro.distributions.transforms`.
- [x] Updates calls using `transform.inverse()` to `transform.inv()`.
- [x] Adds a new class to mimic `pyrenew.transform.LogTransform` using numpyro's `ExpTransform` as baseline instance.


damonbayer:
1. I think any custom transforms we want to keep should still be in their own module.
2. Can we not just define `LogTransform` as `ExpTransform().inv`? Perhaps that doesn't get us all the functionality of a true numpyro transform (maybe it does), but it at least gives us the transform and its inv, which is all we need.

```python
import numpyro.distributions.transforms as nT
import pyrenew.transform as pT
import jax.numpy as jnp
from numpy.testing import assert_array_almost_equal

test_array = jnp.array([0.1, 0.2, 0.4])

LogTransform = nT.ExpTransform().inv

LogTransform(test_array)
LogTransform.inv(test_array)
```

gvegayon:
> 1. I think any custom transforms we want to keep should still be in their own module.
> 2. Can we not just define `LogTransform` as `ExpTransform().inv`? Perhaps that doesn't get us all the functionality of a true numpyro transform (maybe it does), but it at least gives us the transform and its inv, which is all we need.
> 
> ```python
> import numpyro.distributions.transforms as nT
> import pyrenew.transform as pT
> import jax.numpy as jnp
> from numpy.testing import assert_array_almost_equal
> 
> test_array = jnp.array([0.1, 0.2, 0.4])
> 
> LogTransform = nT.ExpTransform().inv
> 
> LogTransform(test_array)
> LogTransform.inv(test_array)
> ```

I updated the code to use `ExpTransform().inv`. About the other transforms, maybe the scaledlogit is the one to keep? But I have my doubts as to where to put it, as it would be a module with a single function.

```python
#140  ScaledLogitTransform
assert_array_almost_equal(
    pT.ScaledLogitTransform(4)(test_array),
    nT.ComposeTransform([nT.AffineTransform(0,1/4), nT.SigmoidTransform().inv])(test_array)
)

assert_array_almost_equal(
    pT.ScaledLogitTransform(4).inverse(test_array),
    nT.ComposeTransform([nT.AffineTransform(0,1/4), nT.SigmoidTransform().inv]).inv(test_array)
)
```

dylanhmorris:
> I updated the code to use ExpTransform().inv. About the other transforms, maybe the scaledlogit is the one to keep? But I have my doubts as to where to put it, as it would be a module with a single function.

I don't think we want to make the user remember which transforms are in `numpyro.distributions.transforms` and which are in `pyrenew.transform`. Counterproposal: expose transforms from `numpyro.distributions.transforms` to the user via `pyrenew.transform`:

That is: have a `pyrenew/transform` module that's a folder, with an `__init__.py` that reads something like:

```python
from pyrenew.transfom.scaledlogittransform import ScaledLogitTransform
from numpyro.distributions.transforms import ExpTransform

__all__ = [
    "ScaledLogitTransform",
    "ExpTransform"
]
```

Then the user can do

```python
from pyrenew.transform import ExpTransform
```

Are there reasons not to do this? Documentation? politeness/attribution?

damonbayer:
@dylanhmorris I would want to ensure that all transforms from numpyro are re-exported in our module. Ideally, without having to list them all out.

dylanhmorris:
This is a case in which an `import *` might be allowed. Something like:

`pyrenew/transform/__init__.py`:
```python
from numpyro.distributions.transforms import * 
from numpyro.distributions.transforms import __all__ as numpyro_public_transforms

from pyrenew.transform.scaledlogittransform import ScaledLogitTransform

__all__ = [
    "ScaledLogitTransform"
] + numpyro_public_transforms
```

AFg6K7h4fhy2 (commented):
I was initially uncertain that `nt._InverseTransform()` was correctly specified but it seems to be the case. It's worth noting that each transform has an inverse property: 

```python
@property
def inv(self):
    inv = None
    if self._inv is not None:
        inv = self._inv()
    if inv is None:
        inv = _InverseTransform(self)
        self._inv = weakref.ref(inv)
    return inv
```

For others to reference, [here](https://num.pyro.ai/en/stable/_modules/numpyro/distributions/transforms.html) is the page on Numpyro transforms, including (note, other transforms, e.g. `_InverseTransform` are defined in the above link):

```python
__all__ = [
    "biject_to",
    "AbsTransform",
    "AffineTransform",
    "CholeskyTransform",
    "ComposeTransform",
    "CorrCholeskyTransform",
    "CorrMatrixCholeskyTransform",
    "ExpTransform",
    "IdentityTransform",
    "L1BallTransform",
    "LowerCholeskyTransform",
    "ScaledUnitLowerCholeskyTransform",
    "LowerCholeskyAffine",
    "PermuteTransform",
    "PowerTransform",
    "RealFastFourierTransform",
    "ReshapeTransform",
    "SigmoidTransform",
    "SimplexToOrderedTransform",
    "SoftplusTransform",
    "SoftplusLowerCholeskyTransform",
    "StickBreakingTransform",
    "Transform",
    "UnpackTransform",
    "ZeroSumTransform",
]
```

model/src/pyrenew/process/rtrandomwalk.py
```diff
@@ -3,7 +3,7 @@
 
 import numpyro as npro
 import numpyro.distributions as dist
-from numpyro.distributions import transforms as nt
+import pyrenew.transformation as nt
```
dylanhmorris, line 6: Should it be just `t` or `pt` now that it's our own module?

model/src/pyrenew/transformation/__init__.py
```diff
@@ -0,0 +1,12 @@
+"""
+This module exposes numpyro's transformations module to the user,
+and adds customed transformations to it.
```
dylanhmorris, line 3:
```suggestion
and defines and adds additional custom transformations
```

model/src/pyrenew/transformation/builtin.py
```diff
@@ -0,0 +1,31 @@
+"""
+Built-in transformations using the numpyro library.
```
dylanhmorris, line 2:
```suggestion
Built-in pyrenew transformations created using `numpyro.distributions.transforms`.
```

model/src/pyrenew/process/rtrandomwalk.py
```diff
@@ -27,7 +27,7 @@ def __init__(
         Rt0_dist: dist.Distribution = dist.TruncatedNormal(
             loc=1.2, scale=0.2, low=0
         ),
-        Rt_transform: AbstractTransform = LogTransform(),
+        Rt_transform: nt.Transform | None = None,
         Rt_rw_dist: dist.Distribution = dist.Normal(0, 0.025),
```
dylanhmorris, line 31: Not for this PR, but we should get rid of the defaults here.

model/src/pyrenew/regression.py
```diff
@@ -10,8 +10,8 @@
 
 import numpyro
 import numpyro.distributions as dist
+import pyrenew.transformation as nt
```
dylanhmorris, line 13: As above, should it be just `t` or `pt` now that it's our own module?

dylanhmorris (commented):
A few minor style/documentation edits

model/src/pyrenew/regression.py
gvegayon: Fixed here ef26272.

model/src/pyrenew/process/rtrandomwalk.py
gvegayon: Fixed here ef26272.

model/src/pyrenew/process/rtrandomwalk.py
gvegayon: Created one #145

dylanhmorris (approved):
LGTM

PR #143, opened 2024-05-30
Process returns tuple with array equal to duration (patch to PR 123)

### LINKED ISSUE #142: Standardize output length for functions that yield sequences/timeseries
> A number of array creation functions (deterministic or stochastic) exist in the project. I think we should standardize on any timeseries/sequence `sample(duration=n)` or `compute(duration=n)` call:
>
>- returning an array (which I'll call $y(t)$ ) of exactly length `n`. 
>- If $x \ge 1$ initial values are provided, including them as the first $x \le$ `n` entries in $y(t)$

_Originally posted by @dylanhmorris in https://github.com/CDCgov/multisignal-epi-inference/pull/131#discussion_r1621392498_

gvegayon:
This addresses and should close #142. The PR does the following:

- Corrects all classes in `pyrenew.process` to return tuples with arrays of shape `(duration,)`.
- Makes `n_timepoints` optional for models under `pyrenew.model`. Only required when observed data is not passed, the length of the time series is inferred from the observed data.
- Consequently, updates tests and tutorials, only using the `n_timepoints` argument when simulating data.
- It also adds additional `assert`ions to the process testing to make sure the length of the output matches `duration`.

**Notice this PR will be merged in to #123**.

cc @damonbayer and @dylanhmorris


gvegayon:
Hold my horses, I realized this should be merged into #131 instead! Thank goodness we have `git cherry-pick`. I'm closing this.

Issue #127, opened 2024-05-20
All `RandomVariable`s should be recorded (even if they are deterministic)


damonbayer:
## Goal

All `RandomVariable`s should be recorded, even if they are deterministic. Currently, when you use create a `DeterministicVariable`, that variable is not "recorded" (included in the MCMC output).

## Context
You could have one `HospitalAdmissionsModel` where the IHR is a non-deterministic `RandomVariable` and one where the IHR is deterministic. If the deterministic `RandomVariable` is not recorded, then the two models will have differently-shaped outputs and post processing (e.g., plotting standard figures) will be complicated.

dylanhmorris:
My inclination here would be for `DeterministicVariable.sample()` to call `numpyro.deterministic()` under the hood, at least by default. Could potentially have a `track=False` or `record=False` setting to change that default behavior. 

https://github.com/CDCgov/multisignal-epi-inference/blob/d8727f1956363fd8759e0be6ff6aa55e639ceb38/model/src/pyrenew/deterministic/deterministic.py#L65-L84

PR #148, opened 2024-06-03
Record deterministic `RandomVariables` by default

damonbayer:
Closes https://github.com/CDCgov/multisignal-epi-inference/issues/127


dylanhmorris:
If `record=True` is going to be the default (and I think it should be, though I'm persuadable), then I think `label` should not have a default argument, as this will make it easy for the user to end up with variable name collisions, as occurs in the test suite here:

```
E       AssertionError: all sites must have unique names but got `a_random_variable` duplicated
```

Also, since the `label` attribute will be passed to `numpyro.deterministic()` as the `name` argument, I think it should be renamed `name` for consistency.

https://github.com/CDCgov/multisignal-epi-inference/blob/83dd9ddf11df1b0e5a5059f34fb7e1a84e466b9e/model/src/pyrenew/deterministic/deterministic.py#L17-L29

damonbayer:
@dylanhmorris @gvegayon Ready for review, but expecting many merge conflicts, depending on when other PR's land.

gvegayon (approved):
Looks good! One minor comment. Should we be more explicit about key word arguments? I feel like we should try to enforce (at least internally and during tutorials) using keywords instead of passing arguments as is. We could deal with this in another issue later.

cc @dylanhmorris

dylanhmorris (approved):
LGTM!

Issue #150, opened 2024-06-04
Read columns with correct data types in datasets module


damonbayer:
Date columns are currently being read as strings.

PR #153, opened 2024-06-04
Read columns with correct data types in datasets module

damonbayer:
Closes https://github.com/CDCgov/multisignal-epi-inference/issues/150


gvegayon:
I can confirm that the tutorial using the dataset reads the columns as dates.

gvegayon (approved):

PR #154, opened 2024-06-04
Renaming datautils to arrayutils

gvegayon:


dylanhmorris:
Image test fails @gvegayon. Otherwise looks good.

gvegayon:
> Image test fails @gvegayon. Otherwise looks good.

Yeah, locally I had to generate the new image, but there's no reason why this should be the case as I am not modifying any critical part of the code.

dylanhmorris (approved):
LGTM

PR #158, opened 2024-06-05
Adding the periodic and ~~weekly~~ day of the week effects and refactoring

gvegayon:
- Creates `arrayutils.PeriodicBroadcaster`.
- Refactors `process.RtPeriodicDiffProcessSample` to leverage the `PeriodicBroadcaster` class.
- Adds two new processes `PeriodicEffect` and `WeeklyEffect`, the former using `PeriodicBroadcaster`.
- Adds the corresponding tests for the new classes.
- Adds a tutorial showing how the `PeriodicBroadcaster` works + a case sampling from a transform Dirichlet distribution (for the Day of the week effect).


gvegayon:
This is now ready for review, @damonbayer.

docs/source/tutorials/example-with-datasets.rst
damonbayer, line 1: I don't think this should be deleted.

model/docs/example_with_datasets.qmd
```diff
@@ -23,7 +23,7 @@ H(t) & = \omega(t) p_\mathrm{hosp}(t) \sum_{\tau = 0}^{T_d} d(\tau) I(t-\tau)
 \end{align*}
 $$
 
-Were $d(\tau)$ is the infection to hospitalization interval, $I(t)$ is the number of latent infections at time $t$, $p_\mathrm{hosp}(t)$ is the infection to hospitalization rate, and $\omega(t)$ is the weekday effect at time $t$; the last section provides an example building a weekly effect `RandomVariable`.
+Were $d(\tau)$ is the infection to hospitalization interval, $I(t)$ is the number of latent infections at time $t$, $p_\mathrm{hosp}(t)$ is the infection to hospitalization rate, and $\omega(t)$ is the day-of-the-week effect at time $t$; the last section provides an example building such `RandomVariable`.
```
damonbayer, line 26:
```suggestion
Were $d(\tau)$ is the infection to hospitalization interval, $I(t)$ is the number of latent infections at time $t$, $p_\mathrm{hosp}(t)$ is the infection to hospitalization rate, and $\omega(t)$ is the day-of-the-week effect at time $t$; the last section provides an example building such a `RandomVariable`.
```

docs/source/tutorials/example-with-datasets.rst
gvegayon: It should, the .rst files under `docs/source/tutorials` are only placeholders reflecting the contents of `model/docs/`. The `example-with-datasets.rst` should have been removed when the tutorial was renamed `example_with_datasets.qmd`.

damonbayer: should we also remove [getting-started.rst](https://github.com/CDCgov/multisignal-epi-inference/blob/main/docs/source/tutorials/getting-started.rst)?

gvegayon:
At some point, files under `docs/source/tutorials` were removed unintentionally. 1a75196 fixes that. Also, a figure for testing was added outside of `model/`. 9b2d667 fixes the latter.

docs/source/tutorials/example-with-datasets.rst
gvegayon: Removed now

docs/source/tutorials/.gitignore
damonbayer, line 1: Was this removed intentionally? If so, why?

damonbayer: I thought it might be a consequence of a tested, but not committed version of `.pre-commit-rst-placeholder.sh` where you delete everything in the tutorials dir

gvegayon: Intentially. At some point, we only wanted to keep track of `index.rst`, but when new tutorials are added, `git status` was not reporting new `rst` files generated by `.pre-commit-rst-placeholder.sh`.

Now that I think of it, perhaps we could modify the script to also add newly generated `rst` files via `git add`?

model/docs/example_with_datasets.qmd
```diff
@@ -297,21 +297,20 @@ We can also take a look at the latent infections:
 out2 = hosp_model.plot_posterior(var="latent_infections", ylab="Latent Infections")
 ```
 
-## Round 2: Incorporating weekday effects
+## Round 2: Incorporating day-of-the-week effects
```
damonbayer, line 300: Can we use the day of week effect provided by process, rather than writing our own here?

model/equations.md
damonbayer, line 1: I never knew this file existed.

model/src/pyrenew/arrayutils.py
```diff
@@ -92,3 +94,158 @@ def pad_x_to_match_y(
...
+        data_starts: int,
+        period_size: int,
+        broadcast_type: str,
+    ) -> None:
+        """
+        Default constructor for PeriodicBroadcaster class.
+
+        Parameters
+        ----------
+        data_starts : int
```
damonbayer, line 132: suggest a more generic name, like `offset`

model/src/pyrenew/arrayutils.py
```diff
@@ -92,3 +94,158 @@ def pad_x_to_match_y(
...
+
+        assert data_starts <= period_size - 1, (
+            "data_starts should be less than or equal to period_size - 1."
+            f"It is {data_starts}. It should be less than or equal "
+            f"to {period_size - 1}."
+        )
+
+        # Broadcast type should be either "repeat" or "tile"
+        assert broadcast_type in ["repeat", "tile"], (
+            "broadcast_type should be either 'repeat' or 'tile'."
```
damonbayer, line 209:
```suggestion
            "broadcast_type should be either 'repeat' or 'tile'. "
```

spacing so it looks right when printed on a single line

model/src/pyrenew/arrayutils.py
```diff
@@ -92,3 +94,158 @@ def pad_x_to_match_y(
...
+        -------
+        ArrayLike
+            Broadcasted array.
+        """
+
+        if self.broadcast_type == "repeat":
+            return jnp.repeat(data, self.period_size)[
+                self.data_starts : (self.data_starts + n_timepoints)
+            ]
+        else:
```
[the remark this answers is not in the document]
damonbayer: Or use a dictionary.

model/src/pyrenew/latent/hospitaladmissions.py
```diff
@@ -49,7 +49,7 @@ class HospitalAdmissions(RandomVariable):
     by the probability of being hospitalized :math:`p_\mathrm{hosp}(t)`.
 
     To account for day-of-week effects in hospital reporting, we use an
-    estimated *weekday effect* :math:`\omega(t)`. If :math:`t` and :math:`t'`
+    estimated *day of the weekday effect* :math:`\omega(t)`. If :math:`t` and :math:`t'`
```
damonbayer, line 52:
```suggestion
    estimated *day of the week effect* :math:`\omega(t)`. If :math:`t` and :math:`t'`
```

model/src/pyrenew/latent/hospitaladmissions.py
```diff
@@ -66,7 +66,7 @@ def __init__(
         infection_to_admission_interval_rv: RandomVariable,
         infect_hosp_rate_rv: RandomVariable,
         observed_hosp_admissions_varname: str = "observed_hosp_admissions",
-        weekday_effect_rv: RandomVariable | None = None,
+        dayofweek_effect_rv: RandomVariable | None = None,
```
damonbayer, line 69:
```suggestion
        day_of_week_effect_rv: RandomVariable | None = None,
```

change throughout

model/src/pyrenew/process/periodiceffect.py
```diff
@@ -0,0 +1,161 @@
...
+        Default constructor for PeriodicEffect class.
+
+        Parameters
+        ----------
+        data_starts : int
+            Relative point at which data starts, must be between 0 and
+            period_size - 1.
+        period_size : int
+            Size of the period.
+        prior : RandomVariable
```
damonbayer, line 48: suggest something more descriptive like `broadcast_values`

model/src/pyrenew/process/periodiceffect.py
```diff
@@ -0,0 +1,161 @@
...
+        The sampled value.
+    """
+
+    value: jnp.ndarray
+
+    def __repr__(self):
+        return f"PeriodicEffectSample(value={self.value})"
+
+
+class PeriodicEffect(RandomVariable):
```
damonbayer, line 27: I think we can get rid of the `PeriodicEffect` or we should create an analogous class for `broadcast_type="repeat"`.

docs/source/tutorials/.gitignore
damonbayer: I wonder if that will create an endless pre-commit loop.

damonbayer (changes requested):
Mostly suggesting variable name changes. Can we also get codecov for this patch to 100%?

docs/source/tutorials/.gitignore
gvegayon: Only if the file changes. `git add` won't add anything if no change is detected (even if the file is new, I think).

model/docs/example_with_datasets.qmd
gvegayon: We could, but I also like having more examples extending the module. Perhaps, we could simplify this by avoiding the unnecessary callbacks (as we discussed today?)

model/src/pyrenew/arrayutils.py
gvegayon: Done in 85b9b00

model/src/pyrenew/latent/hospitaladmissions.py
gvegayon: Addressed in ba5137e

model/src/pyrenew/process/periodiceffect.py
gvegayon: Fixed in 1d1ab6b. I'm using `quantity_to_broadcast`, which is easy to replace, if you want to.

model/docs/periodic_effects.qmd
```diff
@@ -0,0 +1,102 @@
...
+```
+
+```{python}
+with npro.handlers.seed(rng_seed=20):
+    sim_data = rt_proc.sample(duration=30)
+
+# Plotting the Rt values
+import matplotlib.pyplot as plt
+
+plt.plot(sim_data.rt)
```
damonbayer, line 48:
```suggestion
plt.step(np.arange(len(sim_data.rt)), sim_data.rt, where="post")
```

model/docs/periodic_effects.qmd
```diff
@@ -0,0 +1,102 @@
...
+Like before, we can use the `sample` method to generate samples from the day of week effect:
+
+```{python}
+with npro.handlers.seed(rng_seed=20):
+    sim_data = dayofweek.sample(duration=30)
+
+# Plotting the effect values
+import matplotlib.pyplot as plt
+
+plt.plot(sim_data.value)
```
damonbayer, line 91:
```suggestion
plt.step(np.arange(len(sim_data.value)), sim_data.value, where="post")
```

damonbayer:
It's a bit odd that all of the hospitalization code lives in the `latent` sub-package, because it handles observations as well. Also, the weekday effect is definitely more related to observations. Also, I'm not sure what the process sub-package is supposed to be for. Not sure what to make of these observations. Perhaps related to https://github.com/CDCgov/multisignal-epi-inference/issues/163.

model/src/pyrenew/latent/hospitaladmissions.py
```diff
@@ -49,7 +49,7 @@ class HospitalAdmissions(RandomVariable):
     by the probability of being hospitalized :math:`p_\mathrm{hosp}(t)`.
 
     To account for day-of-week effects in hospital reporting, we use an
-    estimated *weekday effect* :math:`\omega(t)`. If :math:`t` and :math:`t'`
+    estimated *day of the week effect* :math:`\omega(t)`. If :math:`t` and :math:`t'`
```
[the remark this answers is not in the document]
gvegayon: Note we do have a reference on the Wastewater model in the same docs [here](https://github.com/CDCgov/multisignal-epi-inference/blob/1d1ab6b4a9facaae251faea0b8f3135fa09ca394/model/src/pyrenew/latent/hospitaladmissions.py#L43). The broadcasting stuff is related to EpiAware, but like the nylon (urban legend)... Still, happy to add a link so it gains more visibility.

gvegayon:
> It's a bit odd that all of the hospitalization code lives in the `latent` sub-package, because it handles observations as well. Also, the weekday effect is definitely more related to observations. Also, I'm not sure what the `process` sub-package is supposed to be for. Seems like it has some generic math and some things specific to Rt. Not sure what to make of these observations. Perhaps related to #163.

That is too broad of a comment for this PR :). Perhaps submit a few tickets? A couple of answers:

1. The latent module can handle observations as it is a possibility (optional feat). I don't see any harm in keeping that unless you feel strongly about it.
2. The way it is built, the day-of-week effect is for a latent variable; how would you observe that?
3. The process module is for random processes, which have the distinction of having a duration. Perhaps @dylanhmorris can weight in?

damonbayer:
1 and 3 are helpful comments. Thanks.

I don't quite follow 2, but I think I may be misled by the variable names in this PR
https://github.com/CDCgov/multisignal-epi-inference/blob/a6cfc9568b038f7ef6f1c97ca715aa469f085f2a/model/src/pyrenew/latent/hospitaladmissions.py#L192-L195


The `observed_hosp_admissions` actually get passed later to an observation process as a mean parameter, right?

gvegayon:
> 1 and 3 are helpful comments. Thanks.
> 
> I don't quite follow 2, but I think I may be misled by the variable names in this PR
> 
> https://github.com/CDCgov/multisignal-epi-inference/blob/a6cfc9568b038f7ef6f1c97ca715aa469f085f2a/model/src/pyrenew/latent/hospitaladmissions.py#L192-L195
> 
> The `observed_hosp_admissions` actually get passed later to an observation process as a mean parameter, right?

100% agree the names are misleading. I've updated them across the functions, tests, and tutorial. It should be ready for a final review now.

model/src/pyrenew/latent/hospitaladmissions.py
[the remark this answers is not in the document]
seabbs: Is there a link out to EpiAware? I think doing it by function makes sense but I don't feel that strongly.

docs/source/tutorials/.gitignore
damonbayer: Made a new issue https://github.com/CDCgov/multisignal-epi-inference/issues/194

model/docs/example_with_datasets.qmd
damonbayer: Resolving this by linking to the built in class above, but keeping the code here.

damonbayer (approved):

PR #161, opened 2024-06-05
Refactor convolve scanner factory functions

### LINKED ISSUE #147: Add transforms to `new_convolve_scanner`, make identity transforms default for `new_double_scanner`
This will:

1. Align the APIs
2. ~~Make it possible to implement logistic susceptibility transforms using `new_convolve_scanner`~~ Actually the specified change won't suffice for this b/c the logistic susceptibility adjustment needs cumulative incidence, so it will need a different (heavier) scanner.

## Out of scope

- an n-fold convolve scanner for arbitrary `n` is worth exploring but out of scope for now.
- a scanner that keeps track of cumulative incidence

dylanhmorris:
Ready for review. 

## This PR:
- Renames `new_double_scanner` to `new_double_convolve_scanner` for consistency.
- Closes #147 by adding transform functionality to `new_convolve_scanner`
- Improves documentation for both scanner factories.
- Adds a unit test for the convolve scanner functions (which previously did not have one)
- This does _not_ have default transforms as on reflection factories this low-level probably should not have defaults at all.

## Out of scope
- More tests needed but I think that is a separate issue/PR.
- Also separate: a factory for convolve scanners that carry along either the full history (more general / less efficient) or cumulative incidence (less general / more efficient). That is needed to implement susceptibility adjustments that depend on keeping track of cumulative incidence (e.g. the `logistic_susceptibility_adjustment`)
- #177


model/src/pyrenew/convolve.py
```diff
@@ -14,33 +14,70 @@
...
 from typing import Callable
-
+from pyrenew.transformation import IdentityTransform
 import jax.numpy as jnp
 from jax.typing import ArrayLike
 
 
-def new_convolve_scanner(discrete_dist_flipped: ArrayLike) -> Callable:
+def new_convolve_scanner(discrete_dist_flipped: ArrayLike,
+                         transform: Callable = None) -> Callable:
```
damonbayer, line 23: Why default to None instead of identity?

dylanhmorris: The [mutable default argument](https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments) gotcha. In many use cases, it isn't a problem (and very likely isn't here), but I have just gotten in the habit of always using this pattern when a default value is of a mutable type.

```python
if arg is None:
   arg = mutable_default()
```

model/src/pyrenew/convolve.py
```diff
@@ -17,62 +17,110 @@
 
 import jax.numpy as jnp
 from jax.typing import ArrayLike
+from pyrenew.transformation import IdentityTransform
```
gvegayon, line 20: Question: isn't this patter better?

```suggestion
from pyrenew.transformation.builtin import IdentityTransform
```

gvegayon: No, that's not better, because the transform module is imported via `__init__.py`. That said, I'm still a bit puzzled. Perhaps this is better:

```suggestion
from numpyro.distributions.transforms import IdentityTransform
```

model/src/pyrenew/convolve.py
```diff
@@ -17,62 +17,110 @@
...
+
+    Where `math`:\vec{d} is a vector of length `math`:n,
+    `math`:m(t) is a scalar for each value of time `math`:t,
+    and `math`:f is a scalar-valued function.
+
+    Given `math`:\vec{d}, and optionally `math`:f,
+    this factory function returns a new function that
+    peforms one step of this process while scanning along
+    an array of  multipliers (i.e. an array
+    giving the values of `math`:m(t)) using :py:func:jax.lax.scan.
```
gvegayon, line 71: This section is not rendering properly on the website:

image

model/src/pyrenew/convolve.py
```diff
@@ -17,62 +17,110 @@
...
     Callable
-        A scanner function that applies two sequential convolutions and transformations.
-        It takes a history subset and a tuple of multipliers, computes the transformations
-        and dot products, and returns the updated history subset and a tuple of new values.
+        A scanner function that applies two sets of
+        convolution, multiply, and transform operations
+        in sequence to construct a new array by scanning
+        along a pair of input arrays that are equal in
+        length to each other.
     """
```
gvegayon, line 121: A little bit of detail via `Notes` could be nice giving a math description as you are doing in `new_convolve_scanner()`.

model/src/pyrenew/latent/infection_functions.py
```diff
@@ -6,7 +6,8 @@
 import jax
 import jax.numpy as jnp
 from jax.typing import ArrayLike
-from pyrenew.convolve import new_convolve_scanner, new_double_scanner
+from pyrenew.convolve import new_convolve_scanner, new_double_convolve_scanner
+from pyrenew.transformation import ExpTransform, IdentityTransform
```
gvegayon, line 10: Yes, now that I think of it, perhaps importing directly from numpyro may be better.

```suggestion
from numpyro.distributions.transforms import ExpTransform, IdentityTransform
```

model/src/pyrenew/latent/infection_functions.py
```diff
@@ -169,12 +170,12 @@ def compute_infections_from_rt_with_feedback(
...
     et cetera.
     """
-    feedback_scanner = new_double_scanner(
+    feedback_scanner = new_double_convolve_scanner(
         dists=(
             reversed_infection_feedback_pmf,
             reversed_generation_interval_pmf,
         ),
-        transforms=(jnp.exp, lambda x: x),
+        transforms=(ExpTransform(), IdentityTransform()),
```
gvegayon, line 178: Nice!

gvegayon (commented):
Looks good! Just a couple of comments.

model/src/pyrenew/convolve.py
damonbayer: I think it is fine as is.

model/src/pyrenew/convolve.py
damonbayer: addressed in https://github.com/CDCgov/multisignal-epi-inference/pull/161/commits/b3e8d98f127323de17b3dfef089429f814dfb54f

damonbayer (commented):
Should we check the size/shape/type of inputs?

model/src/pyrenew/convolve.py
dylanhmorris: I agree, I think if we're going to wrap `numpyro.distributions.transforms` we can and should internally refer to it through our wrapper.

model/src/pyrenew/convolve.py
dylanhmorris: Thanks, apologies for the bad sphinx

dylanhmorris:
> Should we check the size/shape/type of inputs?

@damonbayer are you talking about checking the inputs to the factories or about having the functions they produce/return perform input checks?

dylanhmorris:
Things that occur to me:
1. Check that the double scanner tuples are of exactly length two. As written, it will error if they are less.
2. Check that the `dists` are `ArrayLike` and the transforms are callable.
3. Check that the pair of `dists` are of the same length (in the future, we might want to have the double scanner just use history subsets of the length of the longer array, and automagically sub-subset when taking the dot product with the shorter array, but that feels a bit implicit.

Stricter stuff that I'm more reluctant to implement:
1. Could try to check that the transforms are shape-preserving for `jax` arrays, but that's a bit tricky.
2. Could force the arrays that go into the dot products to be flat.

damonbayer:
@dylanhmorris Sounds good to me.

model/src/pyrenew/convolve.py
dylanhmorris: Done in 8c4a098

model/src/pyrenew/convolve.py
```diff
@@ -19,68 +21,146 @@
...
         history_subset: ArrayLike,
         multipliers: tuple[float, float],
     ) -> tuple[ArrayLike, tuple[float, float]]:  # numpydoc ignore=GL08
         m1, m2 = multipliers
-        m_net1 = t1(m1 * jnp.dot(d1, history_subset))
-        new_val = t2(m2 * m_net1 * jnp.dot(d2, history_subset))
+        m_net1 = t1(m1 * jnp.dot(arr1, history_subset))
+        new_val = t2(m2 * m_net1 * jnp.dot(arr2, history_subset))
         latest = jnp.hstack([history_subset[1:], new_val])
         return latest, (new_val, m_net1)
```
damonbayer, line 165: @gvegayon Do you understand why `m_net1` is returned?

dylanhmorris: 1. It's the result of the first dot/multiply/transform operation, so it's useful for tracking and debugging in many cases.

2.In the specific case of a renewal with infection feedback, `m_net1` is the adjusted (with feedback) $\mathcal{R}(t)$ value. 

Should be better documented.

damonbayer: How does that work with what `jax.lax.scan` expects?

> `f` returns a pair where the first element represents a new value for the loop carry and the second represents a slice of the output.

dylanhmorris: This bit from the same docs is the key:

> Unlike that Python version, both `xs` and `ys` may be arbitrary pytree values, and so multiple arrays can be scanned over at once and produce multiple output arrays. `None` is actually a special case of this, as it represents an empty pytree.

In this case, we end up producing a tuple of arrays, the first an array of `new_val` values, the second an array of `m_net1` values.

Issue #155, opened 2024-06-05
Standardize between `snake_case` and `kebab-case` in names of documentation files


dylanhmorris:
Right now we have a mix.

```
model/docs/extending_pyrenew.qmd
model/docs/getting-started.qmd
model/docs/pyrenew_demo.qmd
```

I favor `snake_case` since we're using that for python files.

Issue #171, opened 2024-06-10
Modularize model/makefile


damonbayer:
## Goal

The makefile should be a bit more flexible and allow us to convert specific files or all the files.

dylanhmorris:
Closed by #172

PR #172, opened 2024-06-10
Standardize documentation file names and modularize docs in makefile

damonbayer:
Closes https://github.com/CDCgov/multisignal-epi-inference/issues/155 and https://github.com/CDCgov/multisignal-epi-inference/issues/171.

This PR renames the documentation files to `snake_case`, and lightly refactors `model/makefile` to be a little more convenient. It also includes a fix so that the `.py` files do not include the raw YAML header, which prevented the python code from running as is.

You can now

- `make docs/x.md` to convert `x.qmd` to `x.md`.
- `make docs/x.ipynb` to convert `x.qmd` to `x.ipynb`.
- `make docs/x.py` to convert `x.py` to `x.py` (with conversion to `x.ipynb` as an intermediate step).

You can also
- `make docs_md` to convert all `qmd`s to `md`s
- `make docs_ipynb` to convert all `qmd`s to `ipynbs`s
- `make docs_py` to convert all `qmd`s to `py`s (with conversion to `ipynb` as an intermediate step).


model/Makefile
```diff
@@ -25,44 +25,50 @@ install:
...
 
-docs/getting-started.md: docs/getting-started.qmd
-	poetry run quarto render docs/getting-started.qmd
+# Derive the corresponding Python files
+PY_FILES := $(NOTEBOOK_FILES:.ipynb=.py)
 
-docs/example-with-datasets.md: docs/example-with-datasets.qmd
-	poetry run quarto render docs/example-with-datasets.qmd
+# Target to build all markdown files
+docs: $(MD_FILES)
```
dylanhmorris, line 54: My only question is whether to include `$(NOTEBOOK_FILES)` and `$(PY_FILES)` in the phony `docs` target (versus and/or in addition to creating their own phony targets)

dylanhmorris (approved):
Thanks for the refactor. One question but it doesn't have to block merge.

model/Makefile
damonbayer: As in `make docs` would create the `md`, `ipynb`, and `py` files? And we would have a separate make `docs/md` to just create the `md`s?

dylanhmorris: right. Also might prefer `docs_md`/`docs_py`/`docs_ipynb` to `docs/md` etc for the names of the phony targets to avoid the appearance of subdirectories.

model/Makefile
```diff
@@ -14,6 +13,7 @@ help:
 	@echo "  docs           : Render the documentation"
 	@echo "  clean          : Clean the generated files"
 	@echo "  docs/notebooks : Convert the Quarto files to Jupyter notebooks"
+	@echo "  docs/py        : Convert the Jupyter notebooks to Python scripts"
```
dylanhmorris, line 16:
```suggestion
	@echo "  docs_clean     : Remove the generated documentation files"
	@echo "  docs_md        : Convert the Quarto files to rendered Markdown files"
	@echo "  docs_ipynb     : Convert the Quarto files to Jupyter notebooks"
	@echo "  docs_py        : Convert the Jupyter notebooks to Python scripts"
``````

dylanhmorris (commented):
Need to update the help message.

dylanhmorris (approved):
LGTM

Issue #156, opened 2024-06-05
Standardize language: `duration` vs `n_timepoints`


damonbayer:
Happy to standardize on `n_timepoints`. Probably a separate PR.

_Originally posted by @dylanhmorris in https://github.com/CDCgov/multisignal-epi-inference/issues/131#issuecomment-2150492937_

PR #179, opened 2024-06-12
Renaming variables throughout for clarity

damonbayer:
Closes https://github.com/CDCgov/multisignal-epi-inference/issues/156.

No changes to any of the model workings. Just renaming variables to make things easier to follow.

For example:

- objects named `x_dist` that are actually random variables are renamed `x_rv`
- object with vague names like `latent` are renamed to be more specific like `latent_infections` or `latent_hosp_admissions`
- `predicted` as a descriptor is replaced by `observed` or `sampled` where appropriate
- acronyms like `ihr` are replaced by more descriptive terms like `infection_hosp_rate`
- `i0` is consistently formatted as `I0`

### DIFF
--- a/model/pyproject.toml
+++ b/model/pyproject.toml
@@ -32,6 +32,23 @@ pytest-cov = "^5.0.0"
 pytest-mpl = "^0.17.0"
 numpydoc = "^1.7.0"
 
+[tool.numpydoc_validation]
+checks = [
+    "GL03",
+    "GL08",
+    "SS01",
+    "PR03",
+    "PR04",
+    "PR07",
+    "RT01"
+]
+ignore = [
+    "ES01",
+    "SA01",
+    "EX01",
+    "SS06",
+    "RT05"
+]
 
 [build-system]
 requires = ["poetry-core"]
--- a/model/pyproject.toml
+++ b/model/pyproject.toml
@@ -49,6 +49,11 @@ ignore = [
     "SS06",
     "RT05"
 ]
+exclude = [  # don't report on objects that match any of these regex
+    '\.undocumented_method$',
+    '\.__repr__$',
+    '\.__call__$'
+]
 
 [build-system]
 requires = ["poetry-core"]
--- a/model/src/pyrenew/deterministic/nullrv.py
+++ b/model/src/pyrenew/deterministic/nullrv.py
@@ -129,7 +129,7 @@ class NullObservation(NullVariable):
 
     def sample(
         self,
-        predicted: ArrayLike,
+        mu: ArrayLike,
         obs: ArrayLike | None = None,
         name: str | None = None,
         **kwargs,
@@ -139,8 +139,8 @@ class NullObservation(NullVariable):
 
         Parameters
         ----------
-        predicted : ArrayLike
-            Rate parameter of the Poisson distribution.
+        mu : ArrayLike
+            Rate parameter of the Poisson distribution. #TODO
         obs : ArrayLike, optional
             Observed data. Defaults to None.
         name : str, optional
--- a/model/src/pyrenew/deterministic/nullrv.py
+++ b/model/src/pyrenew/deterministic/nullrv.py
@@ -140,7 +140,7 @@ class NullObservation(NullVariable):
         Parameters
         ----------
         mu : ArrayLike
-            Rate parameter of the Poisson distribution. #TODO
+            Unused parameter, represents mean of non-null distributions
         obs : ArrayLike, optional
             Observed data. Defaults to None.
         name : str, optional
--- a/model/src/pyrenew/latent/hospitaladmissions.py
+++ b/model/src/pyrenew/latent/hospitaladmissions.py
@@ -20,15 +20,15 @@ class HospAdmissionsSample(NamedTuple):
     ----------
     infection_hosp_rate : float, optional
         The infection-to-hospitalization rate. Defaults to None.
-    predicted : ArrayLike or None
-        The predicted number of hospital admissions. Defaults to None.
+    observed_hosp_admissions : ArrayLike or None
+        The observed number of hospital admissions. Defaults to None.
     """
 
     infection_hosp_rate: float | None = None
-    predicted: ArrayLike | None = None
+    observed_hosp_admissions: ArrayLike | None = None
 
     def __repr__(self):
-        return f"HospAdmissionsSample(infection_hosp_rate={self.IRH}, predicted={self.predicted})"
+        return f"HospAdmissionsSample(infection_hosp_rate={self.infection_hosp_rate}, observed_hosp_admissions={self.observed_hosp_admissions})"
 
 
 class HospitalAdmissions(RandomVariable):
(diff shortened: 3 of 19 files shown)


model/src/pyrenew/model/admissionsmodel.py
```diff
@@ -245,46 +258,57 @@ def sample(
...
-                sampled = None
-            else:
-                sampled = au.pad_x_to_match_y(
-                    sampled_obs, latent, jnp.nan, pad_direction="start"
+                # this is to accommodate the current version of test_model_hosp_no_obs_model. Not sure if we want this behavior
+                sampled_observed_hosp_admissions = au.pad_x_to_match_y(
+                    sampled_observed_hosp_admissions,
+                    latent_hosp_admissions,
+                    jnp.nan,
+                    pad_direction="start",
```
gvegayon, line 305: So, if `sampled_observed_hosp_admissions` is `None`, this works OK by just generating a `nan` array?

gvegayon (dismissed):
Minor comment

model/pyproject.toml
AFg6K7h4fhy2, line 1: As noted, I will investigate this further. Should this investigation prove inconclusive, my belief at present is that a duplicated `numpydoc_validation` across `pyproject.toml` files is not a critical or even moderately pressing issue, i.e. you can merge this PR with this `pyproject.toml` as it stands and be fine.

model/src/pyrenew/deterministic/nullrv.py
```diff
@@ -139,8 +139,8 @@ def sample(
 
         Parameters
         ----------
-        predicted : ArrayLike
-            Rate parameter of the Poisson distribution.
+        mu : ArrayLike
+            Rate parameter of the Poisson distribution. #TODO
```
AFg6K7h4fhy2, line 143: I enjoy, in personal projects, using `TODO`s in docstrings. Dylan has strong opinions against the use of TODOs in most places save for GitHub issues. I agree with this, though do not feel as strongly as Dylan does. The `TODO` here and the norm of having them is in the MSR code is your call as project lead, however it might be useful for everyone to exchange several sentences on the value of including `TODO`s in code.

model/src/pyrenew/latent/hospitaladmissions.py
```diff
@@ -170,32 +173,37 @@ def sample(
...
+            * self.weekday_effect_rv.sample(**kwargs)[0]
         )
 
         # Applying probability of hospitalization effect
-        predicted_admissions = (
-            predicted_admissions
-            * self.hosp_report_prob_dist.sample(**kwargs)[0]
+        observed_hosp_admissions = (
+            observed_hosp_admissions
+            * self.hosp_report_prob_rv.sample(**kwargs)[0]
```
AFg6K7h4fhy2, line 200: Low confidence in this remark, but my intuition is that one of the two `observed_hosp_admissions` can be changed for clarity. Perhaps the `observed_hosp_admissions` in `()` can be changed to `hosp_admissions` &rarr;

```python
observed_hosp_admissions = (
    hosp_admissions
    * self.hosp_report_prob_rv.sample(**kwargs)[0]
```

model/src/pyrenew/latent/hospitaladmissions.py
```diff
@@ -63,73 +63,76 @@ class HospitalAdmissions(RandomVariable):
...
-        infection_to_admission_interval: RandomVariable,
-        infect_hosp_rate_dist: RandomVariable,
-        admissions_predicted_varname: str = "predicted_admissions",
-        weekday_effect_dist: Optional[RandomVariable] = None,
-        hosp_report_prob_dist: Optional[RandomVariable] = None,
+        infection_to_admission_interval_rv: RandomVariable,
+        infect_hosp_rate_rv: RandomVariable,
+        observed_hosp_admissions_varname: str = "observed_hosp_admissions",
+        weekday_effect_rv: Optional[RandomVariable] = None,
+        hosp_report_prob_rv: Optional[RandomVariable] = None,
```
AFg6K7h4fhy2, line 70: Though they seem minor, I really like these changes.

model/src/pyrenew/latent/hospitaladmissions.py
damonbayer: yes, I think one of these should be `latent_hosp_admissions`

model/src/pyrenew/deterministic/nullrv.py
damonbayer: Forgot about this. Thanks for catching.

model/src/test/test_latent_admissions.py
```diff
@@ -60,16 +60,16 @@ def test_admissions_sample():
     )
 
     hosp1 = HospitalAdmissions(
-        infection_to_admission_interval=inf_hosp,
-        infect_hosp_rate_dist=DistributionalRV(
+        infection_to_admission_interval_rv=inf_hosp,
+        infect_hosp_rate_rv=DistributionalRV(
```
AFg6K7h4fhy2, line 64: One could argue for `distrv` here. Depends on what you value. 

Pros:

* More information to the reader

Cons:

* Longer than `rv`
* Now need to do this for every non-just-`rv`
* Using `rv` for all types of `rv`, regarding of `rv`-type, is parsimonious

model/src/pyrenew/model/rtinfectionsrenewalmodel.py
```diff
@@ -299,48 +299,64 @@ def sample(
         # Sampling from Rt (possibly with a given Rt, depending on
         # the Rt_process (RandomVariable) object.)
         Rt, *_ = self.sample_rt(
-            duration=n_timepoints,
+            n_timepoints=n_timepoints,
```
AFg6K7h4fhy2, line 302: This naming (i.e. `n_timepoints`) might change with or guide further discussion regarding the representation of time in this codebase. It is difficult for me to see what is better than `n_timepoints` (maybe `n_timeunits`?) but I thought this was worth flagging.

AFg6K7h4fhy2 (commented):
Looking good for the most part (see comments for flags and or suggestions).

model/src/pyrenew/model/admissionsmodel.py
damonbayer: All tests pass with this code removed, so I will remove it.

model/src/pyrenew/latent/hospitaladmissions.py
AFg6K7h4fhy2, line 70: This is a minor nuisance grandfathered in from before we decided to cease using the semi-outdated `from typing import Optional` and instead use `|`. 

```suggestion
        weekday_effect_rv: RandomVariable | None = None,
        hosp_report_prob_rv: RandomVariable | None = None,
```

AFg6K7h4fhy2 (commented):
Another small remark regarding outdated use of `typing`.

model/src/pyrenew/model/admissionsmodel.py
```diff
@@ -22,22 +22,34 @@ class HospModelSample(NamedTuple):
...
+    sampled_observed_hosp_admissions: ArrayLike | None = None
 
     def __repr__(self):
-        return f"HospModelSample(Rt={self.Rt}, latent_infections={self.latent_infections}, IHR={self.IHR}, latent_admissions={self.latent_admissions}, sampled_admissions={self.sampled_admissions})"
+        """String representation of the HospModelSample
+
+        Returns
+        -------
+        str
+        """
```
AFg6K7h4fhy2, line 45: Pretty sure we can ignore `__repr__` docstrings. 

```suggestion
```

Screenshot 2024-06-12 at 11 53 18

model/src/test/test_latent_admissions.py
damonbayer: It could be a deterministic random variable, so I think the current name is good.

model/src/pyrenew/model/admissionsmodel.py
damonbayer: that's already in the repo, no? It still complains...

AFg6K7h4fhy2: When I run this on my end without the `__repr__` docstring I get no error. The only issue I get is with `ruff` finding that `from typing import Optional` is not used.

AFg6K7h4fhy2 (approved):
Everything looks good. See comment history.

PR #184, opened 2024-06-12
Removing defaults from RtRandomWalkProcess

### LINKED ISSUE #145: Get rid default transformation and default rw distribution in `RtRandomWalkProcess`
Sometimes defaults are good... sometimes:

> Not for this PR, but we should get rid of the defaults here.

_Originally posted by @dylanhmorris in https://github.com/CDCgov/multisignal-epi-inference/pull/140#discussion_r1622736071_

gvegayon:
What the title says: closes #145.

### DIFF
--- a/model/src/pyrenew/process/rtrandomwalk.py
+++ b/model/src/pyrenew/process/rtrandomwalk.py
@@ -24,34 +24,26 @@ class RtRandomWalkProcess(RandomVariable):
 
     def __init__(
         self,
-        Rt0_dist: dist.Distribution = dist.TruncatedNormal(
-            loc=1.2, scale=0.2, low=0
-        ),
-        Rt_transform: t.Transform | None = None,
-        Rt_rw_dist: dist.Distribution = dist.Normal(0, 0.025),
+        Rt0_dist: dist.Distribution,
+        Rt_transform: t.Transform,
+        Rt_rw_dist: dist.Distribution,
     ) -> None:
         """
         Default constructor
 
         Parameters
         ----------
-        Rt0_dist : dist.Distribution, optional
-            Initial distribution of Rt, defaults to
-            dist.TruncatedNormal( loc=1.2, scale=0.2, low=0 )
-
-        Rt_transform : numpyro.distributions.transformers.Transform, optional
-            Transformation applied to the sampled Rt0, defaults
-            to ExpTransform().inv.
-        Rt_rw_dist : dist.Distribution, optional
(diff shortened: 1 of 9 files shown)


model/src/pyrenew/process/rtrandomwalk.py
```diff
@@ -24,35 +24,27 @@ class RtRandomWalkProcess(RandomVariable):
...
     def __init__(
         self,
-        Rt0_dist: dist.Distribution = dist.TruncatedNormal(
-            loc=1.2, scale=0.2, low=0
-        ),
-        Rt_transform: t.Transform | None = None,
-        Rt_rw_dist: dist.Distribution = dist.Normal(0, 0.025),
+        Rt0_dist: dist.Distribution,
+        Rt_transform: t.Transform,
+        Rt_rw_dist: dist.Distribution,
```
damonbayer, line 29:
```suggestion
        Rt0_dist: dist.Distribution,
        Rt_rw_dist: dist.Distribution,
        Rt_transform: t.Transform | None,
```
reorder because `Rt0_dist` and `Rt_rw_dist` are required and `Rt_transform` is optional.
If `Rt_transform is None` make it the `IdentityTransform`

damonbayer (changes requested):
Looks good, just want to re-order function arguments and use IdentityTransform when no transform is supplied.

model/src/pyrenew/process/rtrandomwalk.py
gvegayon: Addressed in ce3ef74.

model/src/pyrenew/process/rtrandomwalk.py
```diff
@@ -24,34 +24,29 @@ class RtRandomWalkProcess(RandomVariable):
 
     def __init__(
         self,
-        Rt0_dist: dist.Distribution = dist.TruncatedNormal(
-            loc=1.2, scale=0.2, low=0
-        ),
+        Rt0_dist: dist.Distribution,
+        Rt_rw_dist: dist.Distribution,
```
damonbayer, line 28: I think these should both be random variables.

damonbayer: I'm open to this being a separate issue.

damonbayer (approved):

model/src/pyrenew/process/rtrandomwalk.py
gvegayon: 100% Agree on this one

Issue #167, opened 2024-06-07
Fix offset in Hospital Admissions Model Figures


damonbayer:
The posterior samples and observations are not offset correctly.
![image](https://github.com/CDCgov/multisignal-epi-inference/assets/8095080/aad253f5-c500-4fcf-ae06-e329261118b5)

Issue #185, opened 2024-06-12
Time handling (related to data class)


gvegayon:
As models become more complex, we need a way to keep track of time units. Whatever solution we end up implementing, it should address the following issues (from internal notes):

- [ ] Convert model time to human time, Ex: `[0, 1, 2] -> [2024-01-01, 2024-01-02, 2024-01-03]` 
- [ ] Convert a weekly array to daily, Ex: `[1,2,3] - > [1,1,1,1,1,1,1,2,2,2,2,2,2,2,3,3,3,3,3,3,3]` 
- [ ] Convert a daily array starting on Tuesday to a daily array starting on Sunday. Ex: `[1, 2, 3] -> [NA, NA, 1, 2, 3]` 
- [ ] Aggregate 7 days of incident admissions to a weekly total. Ex: `[1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14] -> [28, 77]` 

This is closely related to #40.

gvegayon:
Some thoughts about this issue:

1. Part of the solution could be including a `clock()` member for data arrays. In principle, it could be a soft wrapper of `jax.numpy.array` that contains information about (a) the relative time unit, and (b) the time unit of the first element of the array. co-pilot hints we can achieve this using delegation; here is a code snippet illustrating this:

    ```python
    import jax.numpy as jnp
    
    class PyRenewArray:
        def __init__(self, data, time_unit=None, time_reference=None):
            self.data = jnp.array(data)
            self.time_unit = time_unit
            self.time_reference = time_reference
    
        def __getattr__(self, attr):
            # delegate attribute access to the underlying jax.numpy array
            return getattr(self.data, attr)
    
    x = PyRenewArray([1, 2, 3])
    print(x.mean())  # calls jnp.mean(x.data)
    ```

2. Either `PyRenewArray` has a method, or there's a function that allows *syncing* arrays; for instance, in the case of `latent.Infections()`, the sampling function could take all arguments (`Rt`, `I0`, `gen_time`) and sync them before doing anything:

    ```py
    def sample(Rt, I0, gen_time):
        
        # Make sure all are in the same time unit
        sync_arrays(Rt, I0, gen_time, unit = "days")
    ```
    
    `syn_arrays()` would get the `time_unit` and `time_reference` from the `PyRenewArray` attributes to:
    
    - Padded left to match the first array starting point.
    - Broadcast/interpolated/aggregated to daily units.
[...]

PR #186, opened 2024-06-13
Fix offset in Hospital Admissions Model Figures and clean up code

damonbayer:
Closes https://github.com/CDCgov/multisignal-epi-inference/issues/167, but in an imperfect way. We need a better system for handling offsets/padding throughout the project. This is an okay fix for now to get the tutorial looking nice.

A more robust implementation should be addressed when working on https://github.com/CDCgov/multisignal-epi-inference/issues/185.


AFg6K7h4fhy2 (approved):
Looks good. For others checking this PR out, please reference the website and the changes in the plotted figures.

PR #187, opened 2024-06-13
Remove unneeded abstractions in model code

damonbayer:
I don't see any reason to have functions in the class that just call one line functions. We might as well call them directly. We actually do this already for some parts of the model, so this is also a more consistent approach.

My overall ambition is to make models as lightweight as possible and encourage users to make models with minimal overhead. Possibly, going as far as to not provide any prebuilt models in the package itself (but provide ample examples in documentation).


gvegayon:
I like getting lighter code. I would just wait for @dylanhmorris' take on this before merging.

model/src/pyrenew/model/rtinfectionsrenewalmodel.py
AFg6K7h4fhy2, line 1: Some thoughts on this, but before so, here is my original comment, which Damon gave a thumbs up to:

> So is the idea w/ the abstraction reduction PR that since each object I0, Rt_process_rv, etc... will need to be sampled, it makes more sense to cover each of them within sample than for each of them to have their own sample() method? 

In the case that each sampling is done individually, i.e. you have

* `sample_rt()` 
* `sample_IO()`
* `sample_infections_latent()`
* `sample_infection_obs_process()`

as they are, then someone might think and ask, "Oh, each has their own sample function. Hmm, are these sampling functions meaningfully different beyond name?". 

To this person, Trevor (and possibly Damon) might answer, "No. The only case perhaps that can be made here concerns `sample_infection_obs_process()`, which rightly doesn't simply ingest `**kwargs`. 

Differentiating `sample()` calls based on what is being sampled makes explicit how `sample()` is being used; however, if your users already have an understanding of what will be sampled, which we of course have developed in working on this, then the abundance of different sampling functions might seem needlessly abstracted. The new user and newish infectious disease modeller might have less mature a sense of this, and for them it maybe makes more sense to delimit `sample()` calls based on sampled object. 

Alternatively, if documentation can successfully funnel new users to the location of the samplings, i.e. the new `def sample()`, if this PR is merged, then the aforementioned sampling-explicit approach seems just as useful but also more verbose, in which case this PR seems beneficial. 

Two cruxes in my mind for this move is how much people using MSR (1) understand what we are sampling using Numpyro and (2) understand where to find our `sample()` calls. 

Additionally, we should think of situations where one might want to sample just `Rt` or just `IO` and not have to go through with also sampling `latent_infections`, which the new approach does not permit but the previous approach permits. Although, we could make clear that this design decision is a deliberate restriction on how our tools are used.

damonbayer:
Noting that I updated the original description for anyone who had previously read it.

dylanhmorris (approved):
Once this is merged I think the `_rv` suffix for those class attribute `RandomVariables` can be removed, but that's a separate PR.

damonbayer:
> Once this is merged I think the `_rv` suffix for those class attribute `RandomVariables` can be removed, but that's a separate PR.

Happy to discuss further. I know people are generally opposed to variable names that state the variable type, but I think it is an easy remedy to avoid having code like `latent_hosp_admissions = latent_hosp_admissions.sample`.

damonbayer:
I have mangled the git history here somehow. Retrying on https://github.com/CDCgov/multisignal-epi-inference/pull/207.

PR #192, opened 2024-06-14
Pseudo-Random Number Generation For None Input For JAX PRNGKey In Metaclass 

AFg6K7h4fhy2:
In `metaclass.py`, the MCMC run require a pseudo-random key (`jax.random.PRNGKey()`). Without this PR, if the user does not specify their own PRNGKey, MSR instantiates its own PRNGKey using a magic number (54). With this pull request, a random integer between 0 and 100000 is generated and this integer is then used for the PRNGKey when, as a function argument, the PRNGKey is None.


AFg6K7h4fhy2:
In `test_hospitalizations.py` (and other test files) would people like me to also change the following behavior (example) or it's fine?

```python
model1.run(
        num_warmup=500,
        num_samples=500,
        rng_key=jax.random.PRNGKey(272),
        observed_hosp_admissions=model1_samp.sampled_observed_hosp_admissions,
    )
```

AFg6K7h4fhy2:
I grandfathered in the type `jax.random.PRNGKey` in 

```python
def run(
        self,
        num_warmup,
        num_samples,
        rng_key: jax.random.PRNGKey | None = None,
        nuts_args: dict = None,
        mcmc_args: dict = None,
        **kwargs,
    ) -> None:
```

However, the tests fail here, because of `jax.random.PRNGKey`. Inspection suggests that `ArrayLike` from `from jax.typing import ArrayLike` is the correct type to be using: 

```
>>> import jax.random as jr
>>> rand_int_key = jr.PRNGKey(0)
>>> rand_int = jr.randint(
...                 rand_int_key, shape=(), minval=0, maxval=100000
...             )
>>> rng_key = jr.PRNGKey(rand_int)
>>> type(rng_key)
<class 'jaxlib.xla_extension.ArrayImpl'>
```

model/src/pyrenew/metaclass.py
```diff
@@ -339,6 +340,12 @@ def run(
                 nuts_args=nuts_args,
                 mcmc_args=mcmc_args,
             )
+        if rng_key is None:
+            rand_int_key = jr.PRNGKey(0)
+            rand_int = jr.randint(
```
damonbayer, line 345: Doesn't this guarantee that any time `rng_key` is `None`, we will generate the same `rand_in` and therefore get the same samples from the model? Is that what is desired?

AFg6K7h4fhy2:
Some links:

* [JAX random module](https://jax.readthedocs.io/en/latest/jax.random.html)
* [JAX key](https://jax.readthedocs.io/en/latest/_autosummary/jax.random.key.html#jax.random.key)
* [JAX PRNGKey](https://jax.readthedocs.io/en/latest/_autosummary/jax.random.PRNGKey.html#jax.random.PRNGKey)
* [JEP 9263: Typed keys & pluggable RNGs](https://jax.readthedocs.io/en/latest/jep/9263-typed-keys.html#jep-9263) (same link as Damon provided)
* [Type Annotation Roadmap for JAX](https://jax.readthedocs.io/en/latest/jep/12049-type-annotations.html)


For the following, `ArrayLike` still works, but (see last link) `KeyArray` might be implemented as a type soon.

```
>>> import jax.random as jr
>>> import numpy as np
>>> rand_int = np.random.randint(0, 100000)
>>> rng_key = jr.key(rand_int)
>>> type(rng_key)
<class 'jax._src.prng.PRNGKeyArray'>
```

AFg6K7h4fhy2:
NOTE: At some point, having `rng_key: ArrayLike | None = None` _might_ be problematic in the sense that, while `jax.random.key()` and `jax.random.PRNGKey()` both are `ArrayLike` (for the time being, a `KeyArray` type in `jax.typing` might be made soon), the `ArrayLike` has to be from `jax.random.key()` or `jax.random.PRNGKey()` not from a typical JAX array, e.g. `jax.numpy.array([2,3,1])`.

damonbayer:
Can we add a test that checks that two models with no rng key do not produce the same results?

model/src/pyrenew/metaclass.py
```diff
@@ -339,6 +341,9 @@ def run(
                 nuts_args=nuts_args,
                 mcmc_args=mcmc_args,
             )
+        if rng_key is None:
+            rand_int = np.random.randint(0, 100000)
```
damonbayer, line 345:
```suggestion
            rand_int = np.random.randint(
                np.iinfo(np.int64).min, np.iinfo(np.int64).max
            )
```

For extra randomness

damonbayer (changes requested):
To avoid so much repeated code, I suggest writing some helper functions.

One to create the model, so you can have 
```python
model_1 = create_model()
model_2 = create_model()
```

and one to sample from the model so you can do

```python
sample_model(model_1, rng_key = jr.key(54))
sample_model(model_2, rng_key = jr.key(54))
```

damonbayer (approved):

PR #191, opened 2024-06-13
Pyrenew tutorial review

sbidari:
Some ideas for improving the documentation

1. Introduction to renewal model framework and some of the model components implemented in the package. The model equations/information from [here](https://github.com/CDCgov/multisignal-epi-inference/blob/main/model/README.md) and [here](https://github.com/CDCgov/multisignal-epi-inference/blob/main/model/equations.md) can perhaps be moved to the documentation
2. Include installation instructions in the beginning of the documentation either in a separate section or within the getting started section.
3. I am not entirely sure what is the difference between the `HospitalAdmissionsModel` in Pyrenew demo and the one in Fitting a Hospital Admissions-only model. Perhaps clarify what the two tutorials are trying to convey. 
4. If I understand correctly there are two different model implementation - Hospital Admissions and Reproduction Number Renewal Infections. But it isn't clear from the tutorial, suggesting reorganizing the tutorials to make these two model instances examples clear.
5. Change the order of tutorial and reference in the documentation

I can make most of these changes (might need some help on 1) if we think these would be useful.

Issue #7, opened 2024-03-08
Outline Needs for Developer Documentation


ghost:
The idea is to add on the README.md file some resources for new member teams who are not familiar with the renewal models. Details TBD.

Ref: https://github.com/cdcent/cfa-multisignal-renewal/issues/33
author: @gvegayoncdc

gvegayon:
@AFg6K7h4fhy2 @damonbayer @dylanhmorris @cshelley @SamuelBrand1, we started an issue about this a while ago. We can put this on the J sprint, but ideally, do that once we get #40 done.

damonbayer:
I propose this to be implemented as a new section of the site (like "References" or "Tutorials") called "Developer Documentation"

It should cover things like

- Resources for renewal models
- Links to relevant JAX / NumPyro documentation
- Installation for developers with `poetry install --with dev`
- Explanation of helpful utilities in makefiles that do things like convert quarto documents to python for easier debugging.
- Contributor guidelines (related to https://github.com/CDCgov/multisignal-epi-inference/issues/113)

gvegayon:
Another thing to add ([ref](https://github.com/CDCgov/multisignal-epi-inference/pull/123/#issuecomment-2127695207)): Provide a description and step-by-step of the modules testing. Put special attention to the matplotlib component of pytest.

damonbayer:
Also add instructions for generating docs locally with note of manually modified files

AFg6K7h4fhy2:
Example problem:

```
typos....................................................................Failed
- hook id: typos
- exit code: 2

error: `arange` should be `arrange`
  --> scratch/InfectionSeedingProcess_demo.py:18:15
   |
18 | I0_long = jnp.arange(10)
   |               ^^^^^^
   |
error: `arange` should be `arrange`
  --> model/src/pyrenew/latent/infection_seeding_method.py:185:25
    |
185 |             rate * (jnp.arange(self.n_timepoints) - self.t_pre_seed)
    |                         ^^^^^^
    |

```

Example solution: 

Go to `_typos.toml`. Add `arange = "arange"` under `[default.extend-words]`. 

In the case that words are truly misspelled, then please make changes to these words before pushing your commits.

damonbayer:
Also add guidance for variable naming conventions

damonbayer:
From https://github.com/CDCgov/multisignal-epi-inference/pull/197/files, we could add 
[...]

PR #197, opened 2024-06-17
Adding Repository Structure And Resources Sections To The README

AFg6K7h4fhy2:
This is a small pull request that adds two simple items to the README file: the structure of the MSR repository via `tree` and a Resources section. 

A Repository Structure section enables users to get an impression of the MSR repository very quickly. While MSR's structure will change, requiring the Repository Structure section of the README to be manually revised, revisions need not be done frequently, since the design of MSR currently is such that further updates are unlikely to deviate significantly enough for the existing but outdated `tree` output to cease to be useful. The only request I have here for reviewers is to provide some thoughts for what to ignore when running `tree`.

Resources are helpful for newcomers and for those who've forgotten where to find particular items relevant to the project. I have added four resources. More resources can be added by others, including the reviewers. 

As for the placement of the sections in the README, I thought that after the General Disclaimer worked well, since Public Domain Standard Notice becomes less MSR and or CDC specific.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,111 +22,16 @@ This repository was created for use by CDC programs to collaborate on public hea
 
 The structure of the MSR, ignoring the `docs` folder, `__init__.py`s, and image files, can be found via: `tree -I "docs|__init__.py|*.png|*.csv"`. Note, this structure will quite certainly change over 2024 and beyond, yet having this snapshot should still be useful for establishing an impression of the codebase.
 
-```
-.
-├── README.md
-├── _typos.toml
-├── model
-│   ├── Dockerfile
-│   ├── LICENSE
-│   ├── Makefile
-│   ├── README.md
-│   ├── equations.md
-│   ├── pyproject.toml
-│   └── src
-│       ├── pyrenew
-│       │   ├── arrayutils.py
-│       │   ├── convolve.py
-│       │   ├── datasets
-│       │   │   ├── data-raw
-│       │   │   │   ├── README.md
-│       │   │   │   ├── generation_interval.py
-│       │   │   │   ├── infection_admission_interval.py
-│       │   │   │   └── wastewater.R
-│       │   │   ├── generation_interval.py
-│       │   │   ├── generation_interval.tsv
-│       │   │   ├── infection_admission_interval.py
-│       │   │   ├── infection_admission_interval.tsv
-│       │   │   ├── wastewater.py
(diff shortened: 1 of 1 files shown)


damonbayer:
Thanks, Trevor. I agree adding more content to help people get oriented would be helpful.

Re: the MSR Relevant Resources, one sentence or fragment about each resource would be helpful (e.g. For an intro to `x`, we recommend `y`.

Re: the tree structure, I think anything related to the internal structure of the repo would be better suited for [our developer documentation](https://github.com/CDCgov/multisignal-epi-inference/issues/7). Maybe we can remove it from this PR and add a note to that issue, if you agree. In that case, I might suggest we only give the command to print the `tree` and omit the output of the command.

damonbayer (changes requested):
Thanks, Trevor! I agree that adding more content to help people get oriented would be helpful.

Re: the MSR Relevant Resources, one sentence or fragment about each resource would be helpful (e.g. For an intro to `x`, we recommend `y`.

Re: the tree structure, I think anything related to the internal structure of the repo would be better suited for [our developer documentation](https://github.com/CDCgov/multisignal-epi-inference/issues/7). Maybe we can remove it from this PR and add a note to that issue, if you agree. In that case, I might suggest we only give the command to print the `tree` and omit the output of the command.

AFg6K7h4fhy2:
@damonbayer I agree with you that a one sentence description can help. I still believe that having the repository structure displayed in the README would be a net good (perhaps if a Last Updated: YYYY-MM-DD) if provided _somewhere_. I am fine with this _somewhere_ being the developer documentation while leaving a note on how to display the structure for oneself, as you suggested.

sbidari (commented):
Thanks Trevor! These resources would be helpful to newcomers. 
Agree with @damonbayer Re including a sentence or more about the resources. 

I think the model equation sheet could be moved to the documentation along with a brief introduction to the renewal model.

AFg6K7h4fhy2:
If this 

> I think the model equation sheet could be moved to the documentation along with a brief introduction to the renewal model.

expected to be covered in #191 ? 

@sbidari

sbidari:
> Is this
> 
> > I think the model equation sheet could be moved to the documentation along with a brief introduction to the renewal model.
> 
> expected to be covered in #191 ?
> 
> @sbidari

Yes, just requested review for some proposed changes in documentation.

damonbayer (approved):

PR #199, opened 2024-06-18
Sync package versions and update numpyro

damonbayer:
We had different requirements across different sub-folders. Some of these are technically unchanged, but now the same version requirement is written everywhere.


dylanhmorris (approved):
LGTM, thanks!

PR #203, opened 2024-06-18
Fix dangling sentence in demo doc

brandomr:
This is just a simple fix to the demo document, as it had a sentence that cut short prematurely.


damonbayer (approved):
Thanks, @brandomr!

Issue #202, opened 2024-06-18
Create methods for simulating from prior and predictive distributions


damonbayer:
Pre-requisite to https://github.com/CDCgov/multisignal-epi-inference/issues/169.

Model class should have `model.sample_prior()` and `model.predict(samples)` methods.

AFg6K7h4fhy2:
This was mentioned in Teams.

> Did you have an (eventual) solution in mind beyond kwargs? I feel it works well for what we have now but might change in the future if there is something better.

Response:

> It would be nice if you didn’t have to pass any arguments at all. I would want to somehow differentiate between a model and a model with data.

PR #204, opened 2024-06-18
Clean up Rt infections renewal model

damonbayer:
Cleaning up this model in preparation for https://github.com/CDCgov/multisignal-epi-inference/issues/202.


model/Makefile
gvegayon, line 1: Why are you removing this?

model/src/pyrenew/latent/infections.py
```diff
@@ -18,11 +17,11 @@ class InfectionsSample(NamedTuple):
 
     Attributes
     ----------
-    infections : ArrayLike | None, optional
+    post_seed_infections : ArrayLike | None, optional
         The estimated latent infections. Defaults to None.
     """
 
-    infections: ArrayLike | None = None
+    post_seed_infections: ArrayLike | None = None
```
gvegayon, line 24: Is it always the case that it will be *post* seed?

gvegayon (commented):
Main comment: Why is the test image being removed?

damonbayer:
> Main comment: Why is the test image being removed?

The test was failing and I don't really understand the point of the test. The figure is based on underlying data. Why not just test that the underlying data is what it is supposed to be?

model/Makefile
AFg6K7h4fhy2: I have the same question with caveat that I would appreciate some background on which exact images are being tested and why they need to be tested.

AFg6K7h4fhy2:
> > Main comment: Why is the test image being removed?
> 
>     1. The test was failing and can't get the image I produce locally to match the one produced in CI.
> 
>     2. I don't really understand the point of the test.
> 
>     3. If we want to keep a version of the test, we should just evaluate the underlying data the figure is based on, not the actual image file.

Agree on point 3 unless there is something I am missing @gvegayon .

gvegayon:
> > > Main comment: Why is the test image being removed?
> > 
> > 
> > ```
> > 1. The test was failing and can't get the image I produce locally to match the one produced in CI.
> > 
> > 2. I don't really understand the point of the test.
> > 
> > 3. If we want to keep a version of the test, we should just evaluate the underlying data the figure is based on, not the actual image file.
> > ```
> 
> Agree on point 3 unless there is something I am missing @gvegayon .

I 100% Agree with both, so then, instead of removing the test completely, I would drop in a replacement test that checks the data under the figure.

damonbayer:
I'm just not yet sure what a good test would be. We don't test anything related to the posterior anywhere else, so I would be more inclined to push "develop tests related to posterior inference" to a separate issue.

gvegayon:
How about saving the data that it generates now? Just like the image approach, but with model outcomes

damonbayer:
Willing to try it, but I worry that we could get test failures based on very minor package updates. I'm also not sure if the rng would lead to the same output on all platforms.

model/src/pyrenew/latent/infections.py
damonbayer: Yes, the `sample` function requires the seed infections as an argument, so this is refactored to only return the post-seed infections.

model/src/pyrenew/model/rtinfectionsrenewalmodel.py
```diff
@@ -311,57 +313,41 @@ def sample(
 
         # Sampling initial infections
         I0, *_ = self.sample_I0(**kwargs)
-        I0_size = I0.size
         # Sampling from the latent process
-        latent_infections, *_ = self.sample_infections_latent(
+        post_seed_latent_infections, *_ = self.sample_infections_latent(
```
AFg6K7h4fhy2, line 317: Minor: Do you risk is low that someone misconstrues `seed` here for reproducibility / something involved in `numpyro.sample()` vs. interpreting `post_seed_latent_infections` as total infectiousness after seeded infections $I_{v:0}$ have been added?

I think the risk is low, given that those who might be interested in using this enough to peruse the source code would understand this.

model/src/pyrenew/model/rtinfectionsrenewalmodel.py
```diff
@@ -311,57 +313,41 @@ def sample(
...
+            **kwargs,
+        )
+
+        all_latent_infections = jnp.hstack([I0, post_seed_latent_infections])
+        npro.deterministic("latent_infections", all_latent_infections)
 
         observed_infections = au.pad_x_to_match_y(
             observed_infections,
-            latent_infections,
+            all_latent_infections,
```
AFg6K7h4fhy2, line 338: Alternate option: `all_latent_infections` &rarr; `total_latent_infections` 

Thoughts?

dylanhmorris: I prefer `all` as it doesn't imply summation which `total` might.

model/src/pyrenew/model/rtinfectionsrenewalmodel.py
dylanhmorris: I think we should discuss this in a separate issue; have created one: #206

AFg6K7h4fhy2: > I think we should discuss this in a separate issue; have created one: #206

Thank you Dylan. I missed this.

model/src/pyrenew/model/rtinfectionsrenewalmodel.py
AFg6K7h4fhy2: Okay, seems good to me.

AFg6K7h4fhy2 (commented):
Finished reviewing the remaining files. Other than what is mentioned in #206 and maybe just a reminder for us to address the plot testing or to make a final decision on testing the data directly instead of the plotted data, this PR looks useful and appropriate.

damonbayer:
Based on the [Teams discussion](https://teams.microsoft.com/l/message/19:064fb29dee654148b647dc8b9bab0782@thread.tacv2/1717450231741?tenantId=9ce70869-60db-44fd-abe8-d2767077fc8f&groupId=4f34eba7-8fcb-4a26-9b34-a434ea777f0c&parentMessageId=1717450231741&teamName=CFA-Predict&channelName=Help&createdTime=1717450231741) I referenced, I believe this could be merged (but I would need an approving PR), and we can defer a wider discussion of testing posterior inference to some other time and place.

dylanhmorris:
I agree on both counts.

dylanhmorris (approved):

dylanhmorris:
> Perhaps it would be a good topic for an STF team meeting? @dylanhmorris

Can you open an issue for this in the team materials repo @damonbayer?

Issue #215, opened 2024-06-26
Incorporating attributes `timeseries_start` and `timeseries_unit` in all instances of `RandomVariable`, adding the corresponding values of `*_start` and `*_unit` when needed (e.g., the weekly effects should have `timeseries_unit = 7`).


gvegayon:
- `RandomVariable` instances should have:

  - `timeseries_start : int = None`.
  - `timeseries_unit: int = 1`, instantiated during construction.

- There should be a new tutorial/extended documentation discussing time handling in `pyrenew`.

- Random variables for which `timeseries_unit` is relevant should be explicitly passed as an argument during construction; or initialized during construction if the unit is clear, e.g., weekly effects should have `timeseries_unit = 7`.

PR #217, opened 2024-06-26
Adding `t_start` and `t_unit`

gvegayon:
This PR does what #215 lists. Important to note is that the new tutorial sets some design considerations that will be useful for implementing the methods needed to deal with time in the model.


model/docs/time.qmd
gvegayon, line 1: Question: What do you think about the points listed here? One thing we did not discuss was that `timeseries_start` can be negative (e.g., seeding); what are your thoughts?

model/src/pyrenew/metaclass.py
```diff
@@ -95,14 +95,80 @@ def _assert_sample_and_rtype(
...
+    timeseries_unit : int
+        The unit of the time series, e.g., 1: Daily, 7: Weekly.
     """
 
+    timeseries_start: int = None
+    timeseries_unit: int = None
+
+    def __init_subclass__(cls) -> None:
+        """Initializes the subclass with default values."""
+        cls.timeseries_unit = 1
```
gvegayon, line 122: I am setting a default value instantiated during construction for `timeseries_unit`. I am not doing the same for `timeseries_start` because if we, for instance, set it to be zero, we cannot check whether that was set on purpose, or the user forgot to explicitly set the start value, what are your thoughts? After writing this, I am inclining to say no defaults, set all to `None` and then we assume that the `timeseries_unit` equals 1 by default across the package.

model/src/pyrenew/process/periodiceffect.py
```diff
@@ -61,6 +61,11 @@ def __init__(
             broadcast_type="tile",
         )
 
+        self.set_timeseries(
+            timeseries_start=0,
+            timeseries_unit=1,
+        )
```
gvegayon, line 67: Although the broadcaster class in the library talks about period length/unit, I opted to say that samples using `PeriodicBroadcaster` return data at the daily level starting observation time 0. What do you think about this?

model/docs/time.qmd
dylanhmorris: I think we should permit negative `timeseries_start` values for now. The one argument I could see in the other direction in that `timeseries_start` should always in theory be able to be the index of some fundamental zero-indexed time vector. But the whole point of having `timeseries_start` is a construct is so that vector index does not necessarily have to correspond to model timepoint. So I think allowing negative model timepoints is good generality / flexibility.

model/src/pyrenew/metaclass.py
```diff
@@ -95,14 +95,80 @@ def _assert_sample_and_rtype(
...
+
+        Returns
+        -------
+        None
+        """
+
+        if timeseries_unit is None:
+            timeseries_unit = 1
+        else:
+            # Period size should be a positive integer
```
dylanhmorris, line 153:
```suggestion
            # Timeseries unit should be a positive integer
```

model/src/pyrenew/metaclass.py
dylanhmorris: 1. Why do this via an `__init_subclass__` method rather than via the default constructor?
2. Even at initialization, isn't it more futureproof to use the setting method (with validation) that you define below?

model/src/pyrenew/process/periodiceffect.py
dylanhmorris: Any reason not to make this configurable? Or are you just thinking about defaults? My instinct would be to make it mandatory to specify at minimum the unit, given that it's a broadcasting operation.

dylanhmorris (commented):
Responses to your questions @gvegayon

damonbayer:
Largely agreeing with @dylanhmorris's comments. Also wanted to say that I would be open to shorter names, since these will be referenced so often. Some ideas: `time_start/unit`, `ts_start/unit`, or `t_start/unit`.

damonbayer (commented):
Largely agreeing with @dylanhmorris's comments. Also wanted to say that I would be open to shorter names, since these will be referenced so often. Some ideas: `time_start/unit`, `ts_start/unit`, or `t_start/unit`.

model/src/pyrenew/metaclass.py
gvegayon: 1. B/c the `__init__` method may be instance specific, the `__init_subclass__` gives the flexibility of always doing something during instantiation, regardless of the `__init__` method of the instance.
2. Could be! I was wondering that. Furthermore, I wonder if we want `timeseries_unit` to be 1 by default. Maybe, we could just make it None. What do you think?

model/src/pyrenew/process/periodiceffect.py
gvegayon: I like that, probably make both arguments mandatory then: `timeseries_start` and `timeseries_unit`.

gvegayon:
> Largely agreeing with @dylanhmorris's comments. Also wanted to say that I would be open to shorter names, since these will be referenced so often. Some ideas: `time_start/unit`, `ts_start/unit`, or `t_start/unit`.

Agree, at some point I was using `ts_start/ts_unit`. Will return there!

dylanhmorris:
I think the `s` is superfluous; I vote for `t_start/unit`

model/src/pyrenew/latent/infection_seeding_process.py
```diff
@@ -35,6 +42,11 @@ def __init__(
         self.infection_seed_method = infection_seed_method
         self.name = name
 
+        self.set_timeseries(
+            t_start=-self.infection_seed_method.n_timepoints,
```
dylanhmorris, line 46: I don't think this should be hard-coded, though it is certainly a sensible default.

model/src/pyrenew/metaclass.py
dylanhmorris: I prefer default `None` for both attributes at the metaclass level, which also obviates using `__initsubclass__`

dylanhmorris: Default `None` for both attributes
```suggestion
```

dylanhmorris: Suggestion below.

gvegayon: Solved

model/docs/time.qmd
```diff
@@ -0,0 +1,38 @@
+---
+title: Time handling in pyrenew
+format: gfm
+engine: jupyter
+---
+
+The fundamental time unit in `pyrenew` is "days." `pyrenew` deals with time having `RandomVariable`s carrying information about (i) the time unit in terms of days `t_unit` and (ii) the starting time, `t_start`, measured in days since the first observation. Time series data is anchored to the first observation in the model, which is considered day 0.
```
dylanhmorris, line 7: I don't think we should specify that fundamental unit = days. I think we should specify that any given model has a fundamental unit, which could represent any fixed duration duration (e.g. a day).

> Time series data is anchored to the first observation in the model, which is considered day 0.

I don't think we should enforce this.

>  the starting time, `t_start`, measured in days

Is it perhaps clearer to measure `t_start` in the `RandomVariable`'s units, not in fundamental units?

```suggestion
Every `pyrenew` model has a _fundamental time unit_ in which all durations are measured. This should be a unit of approximately fixed length (e.g. "months" would not make for a good fundamental time unit). For many models of interest, this will be days, and we will proceed with this tutorial treating days as our fundamental unit.

 `pyrenew` deals with time having `RandomVariable`s carry information about (i) their own time unit expressed relative to the fundamental unit (`t_unit`) and (ii) the starting time, `t_start`, measured relative to `t = 0` in model time in units of the `RandomVariable`'s time unit. 
```

dylanhmorris (commented):
Key question @gvegayon. Would also appreciate thoughts from @damonbayer

model/docs/time.qmd
damonbayer: Can we add somehting about what the fundamental time unit is used for?

damonbayer: > I don't think we should specify that fundamental unit = days. I think we should specify that any given model has a fundamental unit, which could represent any fixed duration duration (e.g. a day).

Can we add something about what the fundamental time unit is used for?

> > Time series data is anchored to the first observation in the model, which is considered day 0.
> 
> I don't think we should enforce this.
> 
> > the starting time, `t_start`, measured in days
> 
> Is it perhaps clearer to measure `t_start` in the `RandomVariable`'s units, not in fundamental units?

I don't think so. Consider when the fundamental unit is "days." It seems easier to think of/specify "a weekly process starting on day 10" than "a weekly process starting on week 1.42."

dylanhmorris: > I don't think so. Consider when the fundamental unit is "days." It seems easier to think of/specify "a weekly process starting on day 10" than "a weekly process starting on week 1.42."

Good point. And when it _is_ a round number of non-fundamental units the conversion is also easy.

dylanhmorris: > Can we add something about what the fundamental time unit is used for?

Revised suggestion 

```suggestion
Every `pyrenew` model has a _fundamental discrete time unit_. All other time-aware arrays should be expressible as integer fractions or multiples of this fundamental unit.

The fundamental time unit should represent a period of fixed (or approximately fixed) duration. That is, "days" could make for a good fundamental time unit; "months" would not, since different months represent different absolute lengths of time. 

For many infectious disease renewal models of interest, the fundamental time unit will be days, and we will proceed with this tutorial treating days as our fundamental unit.

 `pyrenew` deals with time having `RandomVariable`s carry information about (i) their own time unit expressed relative to the fundamental unit (`t_unit`) and (ii) the starting time, `t_start`, measured relative to `t = 0` in model time in units of the `RandomVariable`'s time unit. 
```

dylanhmorris: My main remaining question is whether we should permit (or anticipate permitting) time units that are integer fractions (as well as integer multiples) of the fundamental time unit. I lean no, as I think supporting this safely might be too much headache for too little gain.

damonbayer: Agreed. I think "1 fundamental time unit" should be the smallest time delta allowed in the model.

gvegayon: I'll add something about delta time units as part of the design principles to be considered during implementation.

gvegayon: Now that I read it, it sounds good as we are not putting any constraints on what the fundamental time unit should be.

model/docs/periodic_effects.qmd
```diff
@@ -74,8 +74,9 @@ mysimplex = dist.TransformedDistribution(
 
 # Constructing the day of week effect
 dayofweek = process.DayOfWeekEffect(
-    offset=0,
+    offset = 0,
```
dylanhmorris, line 77:
```suggestion
    offset=0,
```

model/docs/periodic_effects.qmd
```diff
@@ -74,8 +74,9 @@ mysimplex = dist.TransformedDistribution(
...
     quantity_to_broadcast=metaclass.DistributionalRV(mysimplex, "simp"),
+    t_start= 0,
```
dylanhmorris, line 79:
```suggestion
    t_start=0,
```

model/docs/time.qmd
```diff
@@ -0,0 +1,44 @@
+---
+title: Time handling in pyrenew
+format: gfm
+engine: jupyter
+---
+
+Every `pyrenew` model has a _fundamental discrete time unit_. All other time-aware arrays should be expressible as integer multiples of this fundamental unit.
```
dylanhmorris, line 7:
```suggestion
Every `pyrenew` model has a _fundamental discrete time unit_. All time-aware arrays used in the model should have units expressible as integer multiples of this fundamental unit.
```

model/src/pyrenew/latent/infection_seeding_process.py
```diff
@@ -34,6 +44,13 @@ def __init__(
         self.I_pre_seed_rv = I_pre_seed_rv
         self.infection_seed_method = infection_seed_method
         self.name = name
+        if t_start is None:
+            t_start = -infection_seed_method.n_timepoints
```
dylanhmorris, line 48: Are we sure we want this default? CC @damonbayer

dylanhmorris (commented):
Looks very good. One key question.

model/src/pyrenew/latent/infection_seeding_process.py
damonbayer: Leaning toward yes. I think this is in line with my previous comment about this being an unenforced convention.

dylanhmorris: Agree that if "seeding" is going to default to any time period, the period ending at $t=0$ is the right one.

Related question: we agreed to measure `t_start` in fundamental units, not array units. Are `InfectionSeedingProcess`es and/or `InfectionSeedMethod`s going to be allowed to have non-fundamental units? If so, we potentially need a unit conversion operation here. If not, we need to enforce `t_unit = 1` for both.

My take: 
- it is worth allowing them to have non-fundamental units
- It is possibly worth enforcing that an `InfectionSeedingProcess` has _the same_ units as its `InfectionSeedMethod`

gvegayon: Good point. Created an issue to discuss this.

gvegayon:
Ready for last pass, @damonbayer @dylanhmorris

model/docs/time.qmd
```diff
@@ -0,0 +1,44 @@
...
+engine: jupyter
+---
+
+Every `pyrenew` model has a _fundamental discrete time unit_. All time-aware arrays used in the model should have units expressible as integer multiples of this fundamental unit.
+
+The fundamental time unit should represent a period of fixed (or approximately fixed) duration. That is, "days" could make for a good fundamental time unit; "months" would not, since different months represent different absolute lengths of time.
+
+For many infectious disease renewal models of interest, the fundamental time unit will be days, and we will proceed with this tutorial treating days as our fundamental unit.
+
+ `pyrenew` deals with time having `RandomVariable`s carry information about (i) their own time unit expressed relative to the fundamental unit (`t_unit`) and (ii) the starting time, `t_start`, measured relative to `t = 0` in model time in units of the `RandomVariable`'s time unit.
```
dylanhmorris, line 13: Typo 

```suggestion
 `pyrenew` deals with time having `RandomVariable`s carry information about (i) their own time unit expressed relative to the fundamental unit (`t_unit`) and (ii) the starting time, `t_start`, measured relative to `t = 0` in model time in fundamental time units.
```

model/src/pyrenew/metaclass.py
```diff
@@ -95,14 +95,73 @@ def _assert_sample_and_rtype(
...
+
+    Both attributes give information about the output of the `sample()` method,
+    in other words, the relative time units of the returning value.
+
+    Attributes
+    ----------
+    t_start : int
+        The start of the time series.
+    t_unit : int
+        The unit of the time series, e.g., 1: Daily, 7: Weekly.
```
dylanhmorris, line 114:
```suggestion
        The unit of the time series relative to the model's fundamental 
         (smallest) time unit. e.g. if the fundamental unit is days, 
         then 1 corresponds to units of days and 7 to units of weeks.
```

model/src/pyrenew/metaclass.py
```diff
@@ -95,14 +95,73 @@ def _assert_sample_and_rtype(
...
+    ) -> None:
+        """
+        Set the time series start and unit
+
+        Parameters
+        ----------
+        t_start : int
+            The start of the time series relative to the model time. It could be negative, indicating that the sample returns observations prior to the model time.
+        t_unit : int
+            The unit of the time series, e.g., 1: Daily, 7: Weekly.
```
dylanhmorris, line 139:
```suggestion
        The unit of the time series relative to the model's fundamental 
         (smallest) time unit. e.g. if the fundamental unit is days, 
         then 1 corresponds to units of days and 7 to units of weeks.
```

model/src/pyrenew/process/periodiceffect.py
```diff
@@ -47,6 +49,10 @@ def __init__(
             Size of the period.
         quantity_to_broadcast : RandomVariable
             Values to be broadcasted (repeated or tiled).
+        t_start : int
+            Start time of the process.
+        t_unit : int
+            Unit of time.
```
dylanhmorris, line 55:
```suggestion
            Unit of time relative to the model's fundamental (smallest) time unit. 
```

dylanhmorris (changes requested):
Need to add the time tutorial to the tutorial table of contents. It currently gets rendered and included in the website (can be found via search), but it's not listed under tutorials.

model/src/pyrenew/metaclass.py
```diff
@@ -95,14 +95,83 @@ def _assert_sample_and_rtype(
 class RandomVariable(metaclass=ABCMeta):
     """
     Abstract base class for latent and observed random variables.
+
+    Notes
+    -----
+    `RandomVariable`s in `pyrenew` can be time-aware, meaning that they can
```
dylanhmorris, line 101: this line does not render properly 

Screenshot 2024-07-03 at 4 29 52 PM

dylanhmorris (approved):
One non-critical issue with how the rendered docs look. @gvegayon, your call whether to fix here or merge and make a separate issue to fix.

gvegayon:
> One non-critical issue with how the rendered docs look. @gvegayon, your call whether to fix here or merge and make a separate issue to fix.

Fixed the code in the PR and added an issue referencing this discussion. I think it would be a good thing to review the website and check how the docstrings are rendered.

dylanhmorris (approved):
LGTM. Thanks for creating the issue.

model/src/pyrenew/metaclass.py
dylanhmorris: Renders properly as of 4bfa988

PR #218, opened 2024-06-26
Create methods for simulating from prior and computing predictive distributions

damonbayer:
Adds `posterior_predictive()` and `prior_predictive()` methods to model class. See model/docs/example_with_datasets.qmd for a demonstration of the `posterior_predictive` function.
I don't love the interface, but I think it's the best we can do, based on how numpyro works.
I'm also not sure the best way to handle the documentation, since they are just thin wrappers around numpyro functions.

~~The visualizations leave a lot to be desired. I'm hoping @sbidari could maybe work on them as our resident ArviZ expert. In particular, I think the use of Highest posterior density intervals make the results look worse than they are. As far as I can tell, there is no way to make a plot with equal tail intervals in ArviZ, which would be surprising, if true.~~

~~I kind of think this PR could just be creating the methods and a separate PR could be dedicated to visualizations.~~

Visualization to be addressed in https://github.com/CDCgov/multisignal-epi-inference/issues/221.

Closes https://github.com/CDCgov/multisignal-epi-inference/issues/202.


model/src/pyrenew/metaclass.py
```diff
@@ -412,3 +412,76 @@ def plot_posterior(
...
+            *args,
+            guide=guide,
+            params=params,
+            num_samples=num_samples,
+            return_sites=return_sites,
+            infer_discrete=infer_discrete,
+            parallel=parallel,
+            batch_ndims=batch_ndims,
+            exclude_deterministic=exclude_deterministic,
+        )(rng_key, **kwargs)
```
AFg6K7h4fhy2, line 453: This was mentioned in Teams. 

> Did you have an (eventual) solution in mind beyond kwargs? I feel it works well for what we have now but might change in the future if there is something better. 

Response:

> It would be nice if you didn’t have to pass any arguments at all. I would want to somehow differentiate between a model and a model with data.

model/src/pyrenew/metaclass.py
AFg6K7h4fhy2, line 1: This looks good. Flagging for others that params in `posterior_predictive` and `prior_predictive` can be referenced in <https://num.pyro.ai/en/stable/utilities.html#module-numpyro.infer.util>. See the issue linked to this PR for a comment on the use of `**kwargs`.

damonbayer: Any thoughts on what kind of documentation we need for these functions?

AFg6K7h4fhy2: I think if we had left all the standard `Predictive` params to `*args` then it would make sense to merely document `rng_key` and then make a `Notes` section pointing others to NumPyro. Given the current situation, where all the `Predictive` params are hinted (which I think is good), we can perhaps have much shorter descriptions for each of them, but still have with a Notes section and the link to the NumPyro page I linked to.

AFg6K7h4fhy2: For the sake of this PR, I am fine with retaining `numpydoc ignore=GL08` here for the time being.

model/src/pyrenew/metaclass.py
```diff
@@ -412,3 +412,76 @@ def plot_posterior(
             draws_col=draws_col,
             obs_col=obs_col,
         )
+
+    def posterior_predictive(
+        self,
+        rng_key: ArrayLike | None = None,
+        *args,
```
dylanhmorris, line 419: Slight preference for `*args` before any keyword args

https://docs.python.org/3/glossary.html#term-parameter
```suggestion
        *args,
        rng_key: ArrayLike | None = None,
```

model/src/pyrenew/metaclass.py
```diff
@@ -412,3 +412,76 @@ def plot_posterior(
...
+        guide: Callable | None = None,
+        params: dict | None = None,
+        num_samples: int | None = None,
+        return_sites: list[str] | None = None,
+        infer_discrete: bool = False,
+        parallel: bool = False,
+        batch_ndims: int | None = None,
+        exclude_deterministic: bool = True,
+        **kwargs,
+    ) -> dict:
```
dylanhmorris, line 429: To check my understanding: these keyword arguments need to be enumerated explicitly rather than lumped under `**kwargs`, even though we just use the `Predictive` constructor defaults because `Predictive`'s `__init__()` method does not support arbitrary `**kwargs` but its `__call__()` method __does__, and we want to enable the user to specify __both__ keyword arguments for the `Predictive` constructor and keyword arguments for the `__call__()`.

model/src/pyrenew/metaclass.py
```diff
@@ -412,3 +412,76 @@ def plot_posterior(
...
+            infer_discrete=infer_discrete,
+            parallel=parallel,
+            batch_ndims=batch_ndims,
+            exclude_deterministic=exclude_deterministic,
+        )(rng_key, **kwargs)
+
+    def prior_predictive(
+        self,
+        rng_key: ArrayLike | None = None,
+        *args,
```
dylanhmorris, line 458: As above 
```suggestion
        *args,
        rng_key: ArrayLike | None = None,
```

model/src/pyrenew/metaclass.py
dylanhmorris, line 429: Docstring?

model/src/pyrenew/metaclass.py
```diff
@@ -412,3 +412,76 @@ def plot_posterior(
...
+        guide: Callable | None = None,
+        params: dict | None = None,
+        num_samples: int = None,
+        return_sites: list[str] | None = None,
+        infer_discrete: bool = False,
+        parallel: bool = False,
+        batch_ndims: int | None = None,
+        exclude_deterministic: bool = True,
+        **kwargs,
+    ) -> dict:
```
dylanhmorris, line 468: Docstring?

model/src/pyrenew/metaclass.py
dylanhmorris: Searched around for a canonical pattern for this problem and didn't find one. Some links of interest:

- https://stackoverflow.com/a/309000
- https://github.com/AntoineD/docstring-inheritance

sbidari:
> The visualizations leave a lot to be desired. I'm hoping @sbidari could maybe work on them as our resident ArviZ expert. In particular, I think the use of Highest posterior density intervals make the results look worse than they are. As far as I can tell, there is no way to make a plot with equal tail intervals in ArviZ, which would be surprising, if true.

I don't think there is a built in `eti_plot` feature in ArviZ. We can obtain equal tail intervals using `stat_focus="median"` (see [here](https://python.arviz.org/en/latest/api/generated/arviz.summary.html#arviz.summary)). Passing these eti intervals in place of hdi intervals in `plot_hdi` should work although it's a bit hacky solution

model/src/pyrenew/metaclass.py
damonbayer: Yes. One downside is that the function signature for the Predictive constructor could change underneath us.

seabbs:
> Attention: Patch coverage is 28.57143% with 10 lines in your changes missing coverage. Please review.

Being annoying but unless a new issue is made this should probably be addressed here

damonbayer:
@seabbs Sorry, this PR is a bit of a mess. Converting to draft to indicate that is not ready for a full review. Just wanted to get some early feedback.

AFg6K7h4fhy2 (approved):
Thank you for making more efficient some of the `test_random_key.py` code. Also, I believe the changes you've made in `metaclass.py` to parameter order and how `*args` and `**kwargs` used satisfy Dylan's suggested actions. For a short time I believed there might be something more that could be done in `test_predictive.py` but what you have seems fine.

model/src/pyrenew/metaclass.py
dylanhmorris: Got it; thanks. We can think about alternative approaches in the future, but I think this is fine for now.

model/src/test/test_random_key.py
```diff
@@ -70,44 +90,91 @@ def test_rng_keys_produce_correct_samples():
...
+    ]
     # using same rng_key should get same run samples
     assert np.array_equal(
-        model_01.mcmc.get_samples()["Rt"][0],
-        model_02.mcmc.get_samples()["Rt"][0],
+        models[0].mcmc.get_samples()["Rt"][0],
+        models[1].mcmc.get_samples()["Rt"][0],
+    )
+
+    assert np.array_equal(
```
dylanhmorris, line 124: Suggest [`numpy.testing.assert_array_equal`](https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_array_equal.html) throughout

dylanhmorris: Eventually we might move to [`chex.assert_trees_all_equal`](https://chex.readthedocs.io/en/latest/api.html#chex.assert_trees_all_equal) if we decide to depend on `chex` for testing (see #223)

dylanhmorris (changes requested):
Minor suggestion on assertions, otherwise looks good @damonbayer

gvegayon (commented):
I think it looks good, @damonbayer. The only thing, I would make the thin wrappers ever thinner. Added a comment pointing to the `Model.run` method which is a thin wrapper of another numpyro method ([link](https://github.com/CDCgov/multisignal-epi-inference/blob/33411037820d254cc1733a678bce6706c43ac0cd/model/src/pyrenew/metaclass.py#L313-L352)).

model/src/pyrenew/metaclass.py
damonbayer: I like it.

model/src/pyrenew/metaclass.py
```diff
@@ -412,3 +412,95 @@ def plot_posterior(
...
+        if self.mcmc is None:
+            raise ValueError(
+                "No posterior samples available. Run model with model.run()."
+            )
+
+        if rng_key is None:
+            rand_int = np.random.randint(
+                np.iinfo(np.int64).min, np.iinfo(np.int64).max
+            )
+            rng_key = jr.key(rand_int)
```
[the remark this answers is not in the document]
dylanhmorris: yes, `np.random.randint` output should be reproducible given a current setting for `np.random.seed`

```python
import numpy as np
for k in range(5):
    np.random.seed(52)
    print(np.random.randint(np.iinfo(np.int64).min, np.iinfo(np.int64).max))

# 5960333743209621148
# 5960333743209621148
# 5960333743209621148
# 5960333743209621148
# 5960333743209621148
```

model/src/test/test_random_key.py
damonbayer: addressed in https://github.com/CDCgov/multisignal-epi-inference/pull/218/commits/8e154ce3270afbea7f1d177483b1cd78ab110510

model/src/test/test_predictive.py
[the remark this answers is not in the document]
damonbayer: addressed in https://github.com/CDCgov/multisignal-epi-inference/pull/218/commits/a0702c7b9c292d13630a75c0e7ee6dfa44b74f12

model/src/pyrenew/metaclass.py
dylanhmorris: I like that too. `**kwargs` to the `__call__()` but a dict of `predictive_args` if you want to use non-default values in the `Predictive` constructor

damonbayer: Addressed in https://github.com/CDCgov/multisignal-epi-inference/pull/218/commits/a7f5557bacd4cea30ebf8b41da7a3e6d47d028bc.

model/src/pyrenew/metaclass.py
```diff
@@ -448,36 +447,27 @@ def posterior_predictive(
...
     ) -> dict:
         """
         A wrapper for numpyro.infer.Predictive to generate prior predictive samples.
 
+        Parameters
+        ----------
+        rng_key : ArrayLike, optional
+            Random key for the Predictive function call. Defaults to None.
+        numpyro_predictive_args : dict, optional
+            Dictionary of arguments to be passed to numpyro.inference.Predictive.
```
dylanhmorris, line 469:
```suggestion
            Dictionary of arguments to be passed to the numpyro.inference.Predictive constructor.
```

model/src/pyrenew/metaclass.py
```diff
@@ -415,21 +415,20 @@ def plot_posterior(
...
     ) -> dict:
         """
         A wrapper for numpyro.infer.Predictive to generate posterior predictive samples.
 
+        Parameters
+        ----------
+        rng_key : ArrayLike, optional
+            Random key for the Predictive function call. Defaults to None.
+        numpyro_predictive_args : dict, optional
+            Dictionary of arguments to be passed to numpyro.inference.Predictive.
```
dylanhmorris, line 430:
```suggestion
            Dictionary of arguments to be passed to the numpyro.inference.Predictive constructor.
```

model/src/pyrenew/metaclass.py
```diff
@@ -448,36 +447,27 @@ def posterior_predictive(
...
         """
         A wrapper for numpyro.infer.Predictive to generate prior predictive samples.
 
+        Parameters
+        ----------
+        rng_key : ArrayLike, optional
+            Random key for the Predictive function call. Defaults to None.
+        numpyro_predictive_args : dict, optional
+            Dictionary of arguments to be passed to numpyro.inference.Predictive.
+
```
dylanhmorris, line 470:
```suggestion
**kwargs 
    Additional named arguments passed to the `__call__()` method of numpyro.inference.Predictive

```

model/src/pyrenew/metaclass.py
```diff
@@ -415,21 +415,20 @@ def plot_posterior(
...
         """
         A wrapper for numpyro.infer.Predictive to generate posterior predictive samples.
 
+        Parameters
+        ----------
+        rng_key : ArrayLike, optional
+            Random key for the Predictive function call. Defaults to None.
+        numpyro_predictive_args : dict, optional
+            Dictionary of arguments to be passed to numpyro.inference.Predictive.
+
```
dylanhmorris, line 431:
```suggestion
**kwargs 
    Additional named arguments passed to the `__call__()` method of numpyro.inference.Predictive

```

dylanhmorris (approved):
Minor docstring suggestions, not critical. Otherwise looks good.

dylanhmorris (approved):
LGTM!

Issue #159, opened 2024-06-05
Format Quarto docs in pre-commit


damonbayer:
There is support for formatting quarto docs from within VS Code with the `Format Cell` command, which operates on a single cell, not the whole document. I would like to be able to perform this formatting for all cells in our `qmd` files as part of our `pre-commit` checks but I'm not sure if this is currently easy to do.

Some relevant links:
- A issue in blacken-docs, which is closed without any changes https://github.com/adamchainz/blacken-docs/issues/222
- A comment from a Posit developer stating that they are interested in doing this in the future: https://github.com/quarto-dev/quarto/issues/111#issuecomment-1828819410

PR #220, opened 2024-06-27
Update tutorials formatting

damonbayer:
See https://github.com/CDCgov/multisignal-epi-inference/issues/159.

### DIFF
--- a/model/docs/example_with_datasets.qmd
+++ b/model/docs/example_with_datasets.qmd
@@ -306,6 +306,7 @@ We can use [ArviZ](https://www.arviz.org/) to visualize the results. Let's start
 ```{python}
 # | label: convert-inferenceData
 import arviz as az
+
 idata = az.from_numpyro(hosp_model.mcmc)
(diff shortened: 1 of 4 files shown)


dylanhmorris:
Confirming this does _not_ close #159, correct?

damonbayer:
@dylanhmorris That's correct. I only linked to https://github.com/CDCgov/multisignal-epi-inference/issues/159 to explain why it was not done automatically.

sbidari:
I am confused. Is this formatting done automatically, or no? @damonbayer @dylanhmorris

damonbayer:
It can be done semi-automatically in VS Code. However, it only works on a cell-by-cell basis, so you have to go through each cell and issue a formatting command (or re-save the document if you have format on save enabled). As far as I know, there is no way to format an entire `qmd` at once, but that would be really nice to have in pre-commit.

dylanhmorris (approved):
LGTM

Issue #114, opened 2024-05-09
Remove warning messages from tutorials


gvegayon:
## Goal

Deal with warnings and extraneous messages in the tutorials.

## Context

https://github.com/CDCgov/multisignal-epi-inference/issues/91#issuecomment-2103247777

## Required features

Review the rendered tutorials on the website and remove (when possible) avoidable warning messages like:

```bash
/home/runner/.cache/pypoetry/virtualenvs/pyrenew-ay_vsbmF-py3.12/lib/python3.12/site-packages/tqdm/auto.py:21: TqdmWarning: IProgress not found. Please update jupyter and ipywidgets. See https://ipywidgets.readthedocs.io/en/stable/user_install.html
  from .autonotebook import tqdm as notebook_tqdm
```

## Specifications

- [ ] pyrenew_demo.qmd
- [ ] example-with-dataset.qmd
- [ ] getting-started.qmd

PR #234, opened 2024-07-01
add arviz plot for posterior predictive

sbidari:
Add 'hdi_type' plot with equal tailed credible intervals to demonstrate posterior predictive distribution using arviz

(second try after closing previous one due to mangled commit history)


damonbayer (approved):
This looks nice and I think could be merged. Did you try anything with [plot_lm](https://python.arviz.org/en/stable/api/generated/arviz.plot_lm.html), [plot_ppc](https://python.arviz.org/en/stable/api/generated/arviz.plot_ppc.html), or [plot_ts](https://python.arviz.org/en/stable/api/generated/arviz.plot_ts.html)?

sbidari:
>Did you try anything with [plot_lm](https://python.arviz.org/en/stable/api/generated/arviz.plot_lm.html), [plot_ppc](https://python.arviz.org/en/stable/api/generated/arviz.plot_ppc.html), or [plot_ts](https://python.arviz.org/en/stable/api/generated/arviz.plot_ts.html)?

I couldn't get the 'timeseries' type plots (plot_lm and plot_ts) to work for the hospital admissions model. I suspect it has something to do with the padding as those plots work for the basic model implementation. The output from [plot_lm](https://python.arviz.org/en/stable/api/generated/arviz.plot_lm.html) would be similar to figure generated here.  

I was not sure of we wanted multiple posterior predictive plots. I can include figures using [plot_ppc](https://python.arviz.org/en/stable/api/generated/arviz.plot_ppc.html) if you'd like.

damonbayer:
> I couldn't get the 'timeseries' type plots (plot_lm and plot_ts) to work for the hospital admissions model. I suspect it has something to do with the padding as those plots work for the basic model implementation.

Can you demonstrate those plots working somewhere? In this tutorial or a different one.
> I was not sure of we wanted multiple posterior predictive plots. I can include figures using [plot_ppc](https://python.arviz.org/en/stable/api/generated/arviz.plot_ppc.html) if you'd like. @damonbayer

Yes, please.

sbidari:
> > I couldn't get the 'timeseries' type plots (plot_lm and plot_ts) to work for the hospital admissions model. I suspect it has something to do with the padding as those plots work for the basic model implementation.
> 
> Can you demonstrate those plots working somewhere? In this tutorial or a different one.

Ok! I will add demonstration of these to [getting started](https://cdcgov.github.io/multisignal-epi-inference/tutorials/getting_started.html) tutorial.

damonbayer:
> Am I missing something here? @damonbayer

It is because of the padding and/or seeding. I would try padding in the data in `InferenceData` with missing values for now. If that doesn't work, we can change the model a bit or manually modify the samples.

All of this will be handled better once we have time incorporated throughout the project.

sbidari:
That's what I thought too but this is for the model fit without padding shown below:
```{code}
hosp_model.run(
    num_samples=1000,
    num_warmup=1000,
    data_observed_hosp_admissions=dat["daily_hosp_admits"].to_numpy(),
    rng_key=jax.random.PRNGKey(54),
    mcmc_args=dict(progress_bar=False,num_chains=1),
)
```
Since we are not padding the observation data here, padding shouldn't come into play, right?

damonbayer:
> Since we are not padding the observation data here, padding shouldn't come into play, right?

Just the seeding, then.

damonbayer:
> This solves the dimension issue but now the posterior predictive samples are just observed data.

Oops. We will need to change the model code.

sbidari:
I noticed for both `prior_predictive` and `posterior_predictive` if called using `data_observed_hosp_admissions=dat["daily_hosp_admits"].to_numpy().astype(float)` argument then the predictive distributions returned is simply the observed data.

model/src/pyrenew/model/admissionsmodel.py
```diff
@@ -208,7 +208,7 @@ def sample(
                     observed_hosp_admissions,
                     *_,
                 ) = self.hosp_admission_obs_process_rv.sample(
-                    mu=latent_hosp_admissions,
+                    mu=latent_hosp_admissions[i0_size + padding :],
```
damonbayer, line 211: Mostly a note to myself: I actually think we can remove the entire if-else portion of the model.

damonbayer (changes requested):
@sbidari Can you fix this? The image and the link to `plot_ppc` are both broken.
CleanShot 2024-07-08 at 09 56 36@2x

model/docs/example_with_datasets.qmd
```diff
@@ -260,6 +259,42 @@ out = hosp_model.plot_posterior(
...
+    ),
+)
+
+az.plot_lm(
+    "negbinom_rv",
+    idata=idata,
+);
+```
+
+We can also use the [`plot_ppc`](https://python.arviz.org/en/latest/api/generated/arviz.plot_ppc.html) to compare the posterior predictive samples with the observed data.
```
damonbayer, line 285:
```suggestion
We can also use the `plot_ppc` method to compare the posterior predictive samples with the observed data.
```

model/docs/example_with_datasets.qmd
```diff
@@ -279,7 +279,7 @@ idata = az.from_numpyro(
 az.plot_lm(
     "negbinom_rv",
     idata=idata,
-)
+);
```
damonbayer, line 282: I think the issue might be with capturing the output? In other plots we always do `out = plot`

sbidari: Is the issue with the website rendering? It appears good in my local machine. But I can't remember how to download the tar file for the website

damonbayer: > Is the issue with the website rendering? It appears good in my local machine. But I can't remember how to download the tar file for the website

I think render to `md` works, but there is a problem with the `rst`. This is why I just opened https://github.com/CDCgov/multisignal-epi-inference/pull/250.

model/docs/example_with_datasets.qmd
```diff
@@ -260,6 +259,43 @@ out = hosp_model.plot_posterior(
...
+
+az.plot_lm(
+    "negbinom_rv",
+    idata=idata,
+);
+```
+
+We can also use the `plot_ppc` method to compare the posterior predictive samples with the observed data.
+
+```{python}
```
damonbayer, line 288: Can you explain what this plot is showing?

model/docs/example_with_datasets.qmd
```diff
@@ -260,6 +259,43 @@ out = hosp_model.plot_posterior(
...
+    posterior_predictive=hosp_model.posterior_predictive(
+        n_timepoints_to_simulate=len(dat["daily_hosp_admits"])
+    ),
+    prior=hosp_model.prior_predictive(
+        n_timepoints_to_simulate=len(dat["daily_hosp_admits"]),
+        numpyro_predictive_args={"num_samples": 1000},
+    ),
+)
+
+az.plot_lm(
```
damonbayer, line 280: Really glad this is working. Can we change it to use `kind_pp="hdi"`?

damonbayer (changes requested):
This is looking really good. Left a few more comments. I think that this could also use a final pass to explain or comment on the figures and clean up their axis labels / titles.

model/docs/example_with_datasets.qmd
```diff
@@ -277,23 +277,36 @@ idata = az.from_numpyro(
...
+)
+
+ax.set_title("Posterior Predictive Plot")
+ax.set_ylabel("Hospital Admissions")
+ax.set_xlabel("Days")
+plt.show()
 ```
 
-We can also use the `plot_ppc` method to compare the posterior predictive samples with the observed data.
+We can assess model fit by comparing the model prediction with the observed data. Below we use the [`plot_ppc`](https://python.arviz.org/en/latest/api/generated/arviz.plot_ppc.html) to compare the posterior predictive samples with the observed data for the first two time period. This allows us to check how well the model predictions approximate observed data.
```
damonbayer, line 296:
```suggestion
We can assess model fit by comparing the model prediction with the observed data. Below we use the [`plot_ppc`](https://python.arviz.org/en/latest/api/generated/arviz.plot_ppc.html) to compare the posterior predictive samples with the observed data for the first two time steps. This allows us to check how well the model predictions approximate observed data.
```

model/docs/example_with_datasets.qmd
```diff
@@ -260,6 +259,59 @@ out = hosp_model.plot_posterior(
...
+    axes=ax,
+)
+
+ax.set_title("Posterior Predictive Plot")
+ax.set_ylabel("Hospital Admissions")
+ax.set_xlabel("Days")
+plt.show()
+```
+
+We can assess model fit by comparing the model prediction with the observed data. Below we use the [`plot_ppc`](https://python.arviz.org/en/latest/api/generated/arviz.plot_ppc.html) to compare the posterior predictive samples with the observed data for the first two time steps. This allows us to check how well the model predictions approximate observed data.
```
damonbayer, line 299: Sphinx doesn't like putting links on `code` text for some reason.
CleanShot 2024-07-08 at 13 43 09@2x

sbidari: Oops Sorry! I think I undid the change to this from my merge

model/docs/example_with_datasets.qmd
```diff
@@ -260,6 +259,63 @@ out = hosp_model.plot_posterior(
 )
 ```
 
+Test `posterior_predictive` and `prior_predictive` methods to generate posterior and prior predictive samples.
```
damonbayer, line 262: We should also plot something from the prior predictive as an example.

model/docs/example_with_datasets.qmd
damonbayer: I still don't understand this plot. Why is are the posterior predictive dots so different from the posterior predictive mean?

model/docs/example_with_datasets.qmd
```diff
@@ -370,6 +426,7 @@ out2 = hosp_model.plot_posterior(
     var="all_latent_infections", ylab="Latent Infections"
 )
 ```
+
 and the distribution of latent infections
```
damonbayer, line 430: Edit this text and the text before the previous figure to differentiate them something like "We can look at individual draws from the posterior distribution of latent infections" and then "We can also look at credible intervals for the posterior distribution of latent infections"

damonbayer (changes requested):
A few small requests.

Also, can we get rid of this error? Either by silencing it or sampling in parallel?

```
/home/runner/work/multisignal-epi-inference/multisignal-epi-inference/model/src/pyrenew/metaclass.py:389: UserWarning: There are not enough devices to run parallel chains: expected 2 but got 1. Chains will be drawn sequentially. If you are running MCMC in CPU, consider using `numpyro.set_host_device_count(2)` at the beginning of your program. You can double-check how many devices are available in your system using `jax.local_device_count()`.
  self.mcmc = MCMC(
```

sbidari:
> I still don't understand this plot. Why is are the posterior predictive dots so different from the posterior predictive mean?

I think because the model fit is not great for the first few time units as also evidenced in the plot of posterior samples and posterior predictive samples above this plot. 
Capture

You can see in this figure the first two observation are outside the hdi interval. I can add text below the figure describing why this plot doesn't look good. This would add to the texts already present that mentions that the model fit is not good for the first half and points to next part where it is fixed.

damonbayer:
@sbidari Sorry if I'm being dense, I don't understand what the plot is showing. I think your comment is about the black dot, but I am trying to understand the relationship between the blue dots and the orange line.

image


The figure in the [arviz documentation](https://python.arviz.org/en/stable/api/generated/arviz.plot_ppc.html) looks quite different. I haven't figure out why.

image

damonbayer:
@sbidari thanks for the explanation. I don't really find any of the `plot_ppc` figures very compelling. Perhaps the best option is just to remove it.

model/docs/example_with_datasets.qmd
sbidari: So, I plotted credible intervals for prior predictive distributions ujsing `plot_hdi`, and this is what it looks like:
Capture
Not a very compelling figure. For the purpose of demonstration, I am not aware of any arviz method to plot prior predictive distribution other than `plot_ppc`

dylanhmorris: Log scale y will help here.

damonbayer: Please also overlay the real data. Maybe it is already there but too tiny to see.

sbidari: I changed y scale and added observed data. I also removed the mean curve since the mean was outside hdi for some regions

damonbayer (changes requested):
This is very close. Please change the ordering so that we show the prior predictive before we show the posterior predictive.

damonbayer (approved):
Looks great, @sbidari! There are some warnings in the tutorial, but that doesn't need to prevent merging. There is a longstanding issue for one warning: https://github.com/CDCgov/multisignal-epi-inference/issues/114

I have just made an issue for the other warning: https://github.com/CDCgov/multisignal-epi-inference/issues/255

model/docs/hospital_admissions_model.qmd
```diff
@@ -497,8 +535,112 @@ And plotting the results:
...
+We will use ArviZ to visualize the posterior/prior predictive distributions.
+
+```{python}
+# | label: posterior-predictive-distribution
+idata_weekday = az.from_numpyro(
+    hosp_model_weekday.mcmc,
+    posterior_predictive=hosp_model_weekday.posterior_predictive(
+        n_timepoints_to_simulate=len(dat_w_padding)
+    ),
+    prior=hosp_model_weekday.prior_predictive(
```
damonbayer, line 551: Just noticed that we are using `prior` for what should be `prior_predictive`. However, it doesn't look like `arviz.from_numpyro` has a `prior_predictive` argument. I have opened an issue https://github.com/arviz-devs/arviz/issues/2358

sbidari: From [this](https://python.arviz.org/en/stable/getting_started/CreatingInferenceData.html#from-numpyro) example, the `prior` argument in `arviz.from_numpyro` takes `prior_predictive` as an input.

damonbayer: Ah, I see that supplying `prior` automatically creates the `prior_predictive` component of the `InferenceData` object.

PR #248, opened 2024-07-05
Changing `Model.sample` To `Model.model`

AFg6K7h4fhy2:
This PR exists to change the following, within `metaclass.py`

```python
self.kernel = NUTS(
    model=self.sample,
    **nuts_args,
)
```

to the following:

```python
self.kernel = NUTS(
    model=self.model,
    **nuts_args,
)
```

This change is due given consensus that `model` is clearer than `sample` in the context above.

### DIFF
--- a/model/src/pyrenew/metaclass.py
+++ b/model/src/pyrenew/metaclass.py
@@ -365,6 +365,7 @@ class Model(metaclass=ABCMeta):
         if mcmc_args is None:
             mcmc_args = dict()
 
+        # placeholder for changing this
         self.kernel = NUTS(
             model=self.sample,
             **nuts_args,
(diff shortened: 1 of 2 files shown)


model/src/pyrenew/metaclass.py
```diff
@@ -365,8 +381,9 @@ def _init_model(
         if mcmc_args is None:
             mcmc_args = dict()
 
+        # placeholder for changing this
```
damonbayer, line 384: Remove this?

damonbayer (approved):

model/src/pyrenew/metaclass.py
AFg6K7h4fhy2: Yes, my bad.

damonbayer (changes requested):
Please also change in the `Predictive` calls for prior and posterior prediction.

damonbayer (approved):
Thanks!

PR #233, opened 2024-07-01
Move tutorials in modeldocs to docssourcetutorials

gvegayon:
Moves tutorials under `./model/docs` to `./docs/source/tutorials/`, closing #232.

PR #250, opened 2024-07-08
Add RST to makefile

damonbayer:
This PR also modifies the use of `sed` to work with BSD `sed` (as found on macOS).

### DIFF
--- a/docs/post_process_rst.py
+++ b/docs/post_process_rst.py
@@ -1,0 +1,23 @@
+import re
+import sys
+
+
+def process_rst_file(rst_file):
+    # Read the content of the rst file
+    with open(rst_file, "r") as file:
+        content = file.read()
+
+    # Replace `.. code:: mermaid` with `.. mermaid::`
+    # .. mermaid::
+    modified_content = re.sub(
+        r"\.\.\ code::\ mermaid", ".. mermaid::", content
+    )
+
+    # Write the modified content back to the file
+    with open(rst_file, "w") as file:
+        file.write(modified_content)
+
+
+if __name__ == "__main__":
+    rst_file = sys.argv[1]
+    process_rst_file(rst_file)
--- a/docs/post_process_rst.py
+++ b/docs/post_process_rst.py
@@ -1,3 +1,4 @@
+# numpydoc ignore=GL08
 import re
 import sys
 
@@ -8,7 +9,6 @@ def process_rst_file(rst_file):
         content = file.read()
 
     # Replace `.. code:: mermaid` with `.. mermaid::`
-    # .. mermaid::
     modified_content = re.sub(
         r"\.\.\ code::\ mermaid", ".. mermaid::", content
     )
--- a/docs/post_process_rst.py
+++ b/docs/post_process_rst.py
@@ -4,6 +4,7 @@ import sys
 
 
 def process_rst_file(rst_file):
+    # numpydoc ignore=GL08
     # Read the content of the rst file
     with open(rst_file, "r") as file:
         content = file.read()
--- a/docs/post_process_rst.py
+++ b/docs/post_process_rst.py
@@ -1,12 +1,17 @@
 # numpydoc ignore=GL08
 import re
 import sys
+from pathlib import Path
 
 
 def process_rst_file(rst_file):
     # numpydoc ignore=GL08
     # Read the content of the rst file
-    with open(rst_file, "r") as file:
+    rst_file_path = Path(rst_file)
+    if rst_file_path.suffix != ".rst":
+        raise ValueError("Invalid file format. Only .rst files are supported.")
+
+    with open(rst_file_path, "r") as file:
         content = file.read()
 
     # Replace `.. code:: mermaid` with `.. mermaid::`
@@ -15,7 +20,7 @@ def process_rst_file(rst_file):
     )
 
     # Write the modified content back to the file
-    with open(rst_file, "w") as file:
+    with open(rst_file_path, "w") as file:
         file.write(modified_content)
 
 
--- a/model/Makefile
+++ b/model/Makefile
@@ -42,7 +42,7 @@ docs/%.ipynb: docs/%.qmd
 
 # Pattern rule to build a Python file from a notebook file
 docs/%.py: docs/%.ipynb
-	sed -i.bak '' 's/"cell_type": "raw"/"cell_type": "markdown"/g' $<
+	sed -i.bak 's/"cell_type": "raw"/"cell_type": "markdown"/g' $<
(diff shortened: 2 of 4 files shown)


model/Makefile
```diff
@@ -30,21 +30,31 @@ test:
 docs/%.md: docs/%.qmd
 	poetry run quarto render $<
 
+# Pattern rule to build a reStructuredText file from a qmd file
+docs/%.rst: docs/%.qmd
+	poetry run quarto render $< --to rst
+	sed -i.bak 's/.. code:: mermaid/.. mermaid::/g' docs/$*.rst
```
gvegayon, line 36: Where else is this done? Are you trying to replace [this](https://github.com/CDCgov/multisignal-epi-inference/blob/7822c37028776a2f609916fd13ca8c7106fb10d0/.github/workflows/website.yaml#L41-L50)?

gvegayon:
I think it looks good. But I have a more general/philosophical-kind of question: Should we use the projects' (note the apostrophe) programming language instead of bash/zsh/powershell?

I would appreciate @jasonasher, @seabbs, and @swo's (among others) take on this. The broad question is, should we aim for a more OS-agnostic approach for writing Makefiles/related scripts by leveraging the projects' programming language?

model/Makefile
damonbayer: Not replace, but replicate.

model/Makefile
```diff
@@ -30,21 +30,31 @@ test:
...
 # Pattern rule to build a notebook file from a qmd file
 docs/%.ipynb: docs/%.qmd
 	poetry run quarto convert $< --output $@
 
 # Pattern rule to build a Python file from a notebook file
 docs/%.py: docs/%.ipynb
-	sed -i '' 's/"cell_type": "raw"/"cell_type": "markdown"/g' $<
+	sed -i.bak 's/"cell_type": "raw"/"cell_type": "markdown"/g' $<
 	poetry run jupyter nbconvert --to python $<
+	rm $<.bak
```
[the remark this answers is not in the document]
damonbayer: We came to a similar conclusion in https://github.com/cdcent/cfa-predict-handbook/pull/208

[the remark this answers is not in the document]
dylanhmorris: If there's an easy and readable solution using only the python standard library + existing documentation dependencies, I favor that.

It's nice for both cross-platform robustness and for readability limit the amount of shell magic or quasi-magic in Makefiles.

dylanhmorris:
> I think it looks good. But I have a more general/philosophical-kind of question: Should we use the projects' (note the apostrophe) programming language instead of bash/zsh/powershell?
> 
> I would appreciate @jasonasher, @seabbs, @SamuelBrand1, and @swo's (among others) take on this. The broad question is, should we aim for a more OS-agnostic approach for writing Makefiles/related scripts by leveraging the projects' programming language?

For things like custom regex operations I lean "yes": https://github.com/CDCgov/multisignal-epi-inference/pull/250#discussion_r1671011563

I'm less sure about basic shell operations. I don't think I'd favor `python create_dir.py my_new_dir` over `mkdir -p my_new_dir`

damonbayer:
> I think it looks good. But I have a more general/philosophical-kind of question: Should we use the projects' (note the apostrophe) programming language instead of bash/zsh/powershell?
> 
> I would appreciate @jasonasher, @seabbs, @SamuelBrand1, and @swo's (among others) take on this. The broad question is, should we aim for a more OS-agnostic approach for writing Makefiles/related scripts by leveraging the projects' programming language?

Doing this would increase cross-platform compatibility without introducing any dependencies. I don't see any downsides, other than a one time cost of converting existing scripts. @swo had a very similar discussion here: https://github.com/cdcent/cfa-predict-handbook/pull/220.

damonbayer:
> I think it looks good. But I have a more general/philosophical-kind of question: Should we use the projects' (note the apostrophe) programming language instead of bash/zsh/powershell?
> 
> I would appreciate @jasonasher, @seabbs, @SamuelBrand1, and @swo's (among others) take on this. The broad question is, should we aim for a more OS-agnostic approach for writing Makefiles/related scripts by leveraging the projects' programming language?

In this case, both writing in python and changing the `sed` command achieve cross-platform compatibility without adding new dependencies, so I don't see a particular reason to use one or the other. The only general proclamation I would make is "Ensure cross-platform compatibility while being mindful of introducing dependencies and complexity."

model/Makefile
damonbayer: I see this has a thumbs up @dylanhmorris and @gvegayon. I would accept an approving review (possibly with a new issue to unify the RST post-processing) or a request for changes.

dylanhmorris: I will review; can you create the issue about unified post-processing?

model/docs/.gitignore
```diff
@@ -8,3 +8,4 @@ _compiled
 
 *.ipynb
 *.py
+*.rst
```
dylanhmorris, line 11: Have we made a decision on the future / necessity of placeholder `.rst` files like [this one](https://github.com/CDCgov/multisignal-epi-inference/blob/main/docs/source/tutorials/basic_renewal_model.rst). @gvegayon can you remind me what the justification for them was?

dylanhmorris (commented):
One question, otherwise looks good.

model/docs/.gitignore
damonbayer: The justification is that "suggest and edit" feature on the docs website points to that file, but the file would not otherwise exist in the repo until it is rendered via the github action.

dylanhmorris: afaict git-tracked `.rst` should not end up in `model/docs` even now (and `model/docs` may no longer exist at all after #233) so this seems fine.

docs/post_process_rst.py
```diff
@@ -0,0 +1,23 @@
+import re
+import sys
+
+
+def process_rst_file(rst_file):
+    # Read the content of the rst file
```
dylanhmorris, line 8: maybe error if extension is not `.rst`?

dylanhmorris (approved):
LGTM, one minor thought/suggestion that doesn't need to block merge.

damonbayer:
Wanted to note that this post-processing is somewhat superfluous. We could just change instances of
```
```mermaid
```
to
```
```{mermaid}
```

Then quarto will render theme as images, which is not quite as nice as the javascript rendering.

dylanhmorris (commented):
LGTM

dylanhmorris:
> Then quarto will render theme as images, which is not quite as nice as the javascript rendering.

I think the js rendering is worth the trouble

PR #253, opened 2024-07-09
Add `__call__` As Alias For `RandomVariable` `sample` Method

AFg6K7h4fhy2:
This PR is aimed at cleaning up code in tutorials, tests, and elsewhere. The `RandomVariable` class in `metaclass.py` has an abstract sample method. Subclasses of `RandomVariable` typically are sampled. By having an alias for sample in `__call__`, MSR users need not continually write `RV.sample()`, as this will be done automatically.


damonbayer (commented):
Thanks, @AFg6K7h4fhy2! There are some more places where explicit `sample()` calls could be replaced with the what you have done in this PR (for example, in the two models). Do you think we should pursue this in this PR or elsewhere?

AFg6K7h4fhy2:
> Thanks, @AFg6K7h4fhy2! There are some more places where explicit `sample()` calls could be replaced with the what you have done in this PR (for example, in the two models). Do you think we should pursue this in this PR or elsewhere?

A silly mistake was made. See DMs. I will cover the rest of these here.

AFg6K7h4fhy2 (commented):
I support your changes @damonbayer

damonbayer (approved):

PR #257, opened 2024-07-10
sample multiple chain sequentially to avoid warning

sbidari:
Added `chain_method="sequential"` to tutorials while sampling multiple chains to avoid warning.

Note, only made changes in basic renewal model in this PR, hospital admissions model was already modified previously.


dylanhmorris (approved):
LGTM

PR #241, opened 2024-07-02
Forecasting Interface

damonbayer:
Closes https://github.com/CDCgov/multisignal-epi-inference/issues/169
Partially addresses https://github.com/CDCgov/multisignal-epi-inference/issues/249.

This PR

1. Upgrades minimum version of`jax` and `numpyro` in light of https://github.com/pyro-ppl/numpyro/issues/1827.
2. Modifies models to have a clearer definition of some time-related concepts.
- `n_timepoints_to_simulate` is always the length of the "data" (whether or not it is observed). I would like to rename it to `n_datapoints` in a separate PR.
- Observations begin after the seeding period and the padding period. Thus the total length of the latent infections is `n_timepoints_to_simulate` + `seeding_period_length` + `padding`
3. Cleans up `hospital_admissions_model.qmd` in light of (2) and to reduce code repetition.
4. Simplifies `admissionsmodel.py` and `rtinfectionsrenewalmodel.py` to avoid unnecessary `if-else` paths.
5. Modifies `simplerandomwalk.py` to use `numpyro.contrib.control_flow.scan` rather than an `expand`ed distribution, which allows for forecasting to "just work."
6. Adds a forecasting example to `hospital_admissions_model.qmd`.

# Original PR text:

Currently working on refactoring the `SimpleRandomWalkProcess` using `scan`, which should make forecasting possible. It works in forward simulation, but is broken in inference.

This works:
```python
from pyrenew.process import SimpleRandomWalkProcess
from numpyro import distributions as dist
import numpyro
srwp = SimpleRandomWalkProcess(dist.Normal(0, 1))
with numpyro.handlers.seed(rng_seed=0):
    x = srwp.sample(n_timepoints=10, init=0.0)
x
```

But `pyrenew_demo.qmd` fails during inference.

It says:
```pycon
TypeError: body_fun output and input must have identical types, got
('ShapedArray(int32[], weak_type=True)', ['ShapedArray(float32[33])', 'DIFFERENT ShapedArray(int32[], weak_type=True) vs. ShapedArray(float0[])', 'ShapedArray(float32[])'], []).
```

PR #261, opened 2024-07-11
Renaming The Infections Seeding Initialization

AFg6K7h4fhy2:
This PR clarifies naming for $I_{v:0}$, i.e. the seeded infections, by framing the variable for infections seeding in terms of "initialization" versus "seeding", which, in most instances, is perceived by users as the value for pseudo-random number generation reproducibility.


damonbayer (changes requested):
Thanks @AFg6K7h4fhy2! I think we should perform more renaming throughout. For instance, `InfectionSeedingProcess` should be `InfectionInitializationProcess`. In general, you should search the whole repo for `seed` and make sure it is only used in the context of RNG seeds. @dylanhmorris do you agree?

AFg6K7h4fhy2:
> Thanks @AFg6K7h4fhy2! I think we should perform more renaming throughout. For instance, `InfectionSeedingProcess` should be `InfectionInitializationProcess`. In general, you should search the whole repo for `seed` and make sure it is only used in the context of RNG seeds. @dylanhmorris do you agree?

I agree. I held because I didn't want to decide unilaterally. Beyond `InfectionSeedingProcess`, I am thinking any other place where we have `Infection` and `Seeding` not in reference to pseudo-random number usage.

dylanhmorris:
> Thanks @AFg6K7h4fhy2! I think we should perform more renaming throughout. For instance, InfectionSeedingProcess should be InfectionInitializationProcess. In general, you should search the whole repo for seed and make sure it is only used in the context of RNG seeds. @dylanhmorris do you agree?

I do. @AFg6K7h4fhy2, I think you can assume `initialization` is the agreed-upon substitute for `seeding` throughout.

AFg6K7h4fhy2:
> I do. @AFg6K7h4fhy2, I think you can assume `initialization` is the agreed-upon substitute for `seeding` throughout.

Sounds good, I will edit this after I get through #241

damonbayer:
Thanks @AFg6K7h4fhy2! I prefer `post_initialization_infections` over `post_initialized_infections`.

AFg6K7h4fhy2:
> Thanks @AFg6K7h4fhy2! I prefer `post_initialization_infections` over `post_initialized_infections`.
> 
> The documentation, for example `docs/source/msei_reference/latent.rst`, also needs to be changed.

Will do on `post_initialized_infections` &rarr; `post_initialization_infections`. Good catch on the `./docs/` changes. I think I was one level off when doing `Command-Shift-F`.

damonbayer:
I think we also want to rename the `InfectionInitializationMethod`s (e.g., `SeedInfectionsExponentialGrowth -> InitializeInfectionsExponentialGrowth`).

AFg6K7h4fhy2:
> I think we also want to rename the `InfectionInitializationMethod`s (e.g., `SeedInfectionsExponentialGrowth -> InitializeInfectionsExponentialGrowth`).

I was thinking that since they were subclasses the connection with seeding infections shouldn't be completely removed, but I see the point. Will change.

damonbayer (approved):
Looks good. Thanks @AFg6K7h4fhy2!

PR #262, opened 2024-07-11
Implements SampledValue

gvegayon:
- Introduces the `SampledValye` class with `value`, `t_start`, and `t_unit` attributes (time-related attributes are optional).
- Updates all `RandomVariable.sample()` calls to return tuples/namedtuples with `SampledValue` instead of `ArrayLike` objects.
- Tweaks one of the tutorials so that the figures show reasonable values.

image

The latter is the new version of the last plot, the previous figure looks like the following:

image

~~- It fixes a hidden bug in the hospital admissions model tutorial where the day-of-the-week effect was not properly used.~~

## For later

Also, in the extending pyrenew tutorial

image
New version of the hospital admissions plot. vs:
image

**This is not implemented as the section on using the day-of-week effect was removed from the tutorial.**


gvegayon:
Just found a bug in the code. Particularly the definition of the day of the week effect in the `hospital_admissions_model.qmd` tutorial. The current implementation does not use all the day-of-the-week effect samples but rather the first observation (Monday/Sunday, depending on where you live). Here is the code:

https://github.com/CDCgov/multisignal-epi-inference/blob/ce055afcaa2998d152ce39b3efd248d98eb3ddef/docs/source/tutorials/hospital_admissions_model.qmd#L502

In other words, instead of returning a tuple/namedtuple, it returns a jax array, which then, when used within `latent.HospitalAdmissions.sample()` fails to use the full vector:

https://github.com/CDCgov/multisignal-epi-inference/blob/ce055afcaa2998d152ce39b3efd248d98eb3ddef/model/src/pyrenew/latent/hospitaladmissions.py#L192-L195

The example should instead return a tuple:

```python
         return (jnp.tile(ans, self.nweeks)[: self.len],)
```

But then, the code fails because the day-of-the-week RV samples a vector of length `dat.shape[0] = 90` and tries multiplying it by a vector of infections of length `dat.shape[0] + padding (21) + seeding (15) = 126`.

For this PR, I'll fix the example by returning a tuple with a `TimeArray` (as all RVs in the PR), and specifying the sample size to be 126. **A better solution will be to refactor `latent.HospitalAdmissions` to use a `TimeArray` instead of a `ArrayLike` for the day-of-the-week effect, passing the `latent_admissions` as a `TimeArray` so it will figure out what is the proper length (and alignment) automatically.**

This could have been identified if we checked that all calls to `RandomVariable.sample()` return a namedtuple or a tuple of `TimeArray`s. I suggest leveraging we now have the `RandomVariable.__call__()` method to introduce a check there and have `RandomVariable.sample()` be renamed to `RandomVariable._sample_()` so it is now considered to be internal.

attn @damonbayer @dylanhmorris

docs/source/tutorials/hospital_admissions_model.qmd
```diff
@@ -499,14 +490,16 @@ class DayOfWeekEffect(metaclass.RandomVariable):
             sample_shape=(7,),
         )
 
-        return jnp.tile(ans, self.nweeks)[: self.len]
+        return (metaclass.TimeArray(
+            jnp.tile(ans, 100)[: n_timepoints]),
```
gvegayon, line 494: I haven't been able to figure out why using `int(jnp.ceil(n_timepoints/7))` instead of hardcoding 100 doesn't work. I understand `jax` usually doesn't like that pattern, but the periodic effect works (see [here](https://github.com/CDCgov/multisignal-epi-inference/blob/ce055afcaa2998d152ce39b3efd248d98eb3ddef/model/src/pyrenew/process/periodiceffect.py#L64-L68)). Pre-specifying the number of weeks this effect needs to be tiled beforehand is not an option as it won't work with the posterior predictive method.

model/src/pyrenew/metaclass.py
```diff
@@ -152,6 +192,15 @@ def set_timeseries(
         -------
         None
         """
+
+        # Either both values are None or both are not None
+        assert (t_unit is not None and t_start is not None) or (
+            t_unit is None and t_start is None
+        ), "Both t_start and t_unit should be None or not None."
```
damonbayer, line 199: I think we could be more helpful here by printing `t_start` and `t_unit` in the error message.

docs/source/tutorials/hospital_admissions_model.qmd
damonbayer: In what way doesn't it work? Do you get an error?

dylanhmorris: I'd suggest staying away from this pattern entirely. I'd have the returned array be of length 7 and then use a mask rather than a tiling to create an array aligned with the actual data. In other words, if I have some vector of dow (here a scaled simplex):
```python
dow_effects = 7 * jnp.array([0.1, 0.2, 0.1, 0.2, 0.1, 0.2, 0.1])
```

And I have a vector of values to which I'd like to associate the d.o.w. effect, which might be more or fewer than 7 entries long:

```python
pre_dow_effect_values_shorter = jnp.array([15, 7, 2, 4, 6])
pre_dow_effect_values_longer = jnp.array([35, 32, 13, 5, 10, 15, 36, 20, 3])
```

I do this with a lookup mask, e.g.:

```python
dow_index_shorter = jnp.array([5, 6, 0, 1, 2])
dow_index_longer = jnp.array([3, 4, 5, 6, 0, 1, 2, 3, 4])
```

I can then do:

```python
dow_effects_shorter = dow_effects[dow_index_shorter]
dow_effects_longer = dow_effects[dow_index_longer]
```

gvegayon: > In what way doesn't it work? Do you get an error?

That's right, @damonbayer. The compiler doesn't like casting floats into ints in some situations. Not always the case, as the periodic effects work just fine. Will work on this next week.

damonbayer: > > In what way doesn't it work? Do you get an error?
> 
> That's right, @damonbayer. The compiler doesn't like casting floats into ints in some situations. Not always the case, as the periodic effects work just fine. Will work on this next week.

Did you try integer division? https://stackoverflow.com/a/17511341

damonbayer:
@gvegayon did you consider implementing this with `__jax_array__`?

gvegayon:
> @gvegayon did you consider implementing this with `__jax_array__`? I think you did, but I couldn’t remember.

I have not. I was checking, and there's not enough documentation to make good use of it. Looking at the source code of `jax`, it seems we would have to implement `__jax_array__` as a function call. I suggest adding this as an issue and moving forward; mostly because it is not properly documented yet.

docs/source/tutorials/hospital_admissions_model.qmd
gvegayon: I'll try that, @damonbayer. Thanks!

gvegayon: Works like a charm, @damonbayer! Thanks.

gvegayon:
The current conflict includes the deleted component on the day-of-the-week effects in https://github.com/CDCgov/multisignal-epi-inference/pull/281. This PR fixes it; should I resolve the conflict by including the previously deleted effect?

model/src/pyrenew/deterministic/deterministic.py
dylanhmorris, line 1: I could be persuaded, but I do not think `DeterministicVariable`s should necessarily be `TimeArray`s.

If we want a deterministic, time-aware RV, wouldn't it be simpler to do it this way?

```python
my_time_aware_value = TimeArray([5, 23, 4], t_start=-4, t_unit=1)
my_time_aware_deterministic = DeterministicVariable(my_time_aware_value)
```

gvegayon:
> > The current conflict includes the deleted component on the day-of-the-week effects in #281. This PR fixes it; should I resolve the conflict by including the previously deleted effect?
> 
> Could we demonstrate it in the "Time handling in pyrenew" tutorial instead?

Surely, but I would like to do that in the next PR then. I think this would be a good opportunity to make use of the `TimeArray` attributes, for which we would need to: (a) encode time in the observed data, which we are not doing right now, and (b) refactor the implementation of the latent hospitalizations module so it would pass that information to the day-of-the-week effect.

model/src/pyrenew/deterministic/deterministic.py
gvegayon: > > The current conflict includes the deleted component on the day-of-the-week effects in #281. This PR fixes it; should I resolve the conflict by including the previously deleted effect?
> 
> Could we demonstrate it in the "Time handling in pyrenew" tutorial instead?

Surely, but I would like to do that in the next PR then. I think this would be a good opportunity to make use of the `TimeArray` attributes, for which we would need to: (a) encode time in the observed data, which we are not doing right now, and (b) refactor the implementation of the latent hospitalizations module so it would pass that information to the day-of-the-week effect.

gvegayon:
> > The current conflict includes the deleted component on the day-of-the-week effects in #281. This PR fixes it; should I resolve the conflict by including the previously deleted effect?
> 
> Could we demonstrate it in the "Time handling in pyrenew" tutorial instead?

Surely, but I would like to do that in the next PR then. I think this would be a good opportunity to make use of the `TimeArray` attributes, for which we would need to: (a) encode time in the observed data, which we are not doing right now, and (b) refactor the implementation of the latent hospitalizations module so it would pass that information to the day-of-the-week effect.

(sorry I added and deleted this comment)

model/src/pyrenew/deterministic/deterministic.py
dylanhmorris: We previously discussed having a `SampledValue` class whose `value` attribute can be an array or something else and whose `t_start` and `t_unit` attributes can be `None`. For example, both of the below approaches represent a time-unaware sampled scalar, but I think the first is clearer about the intent:

```python
var = SampledValue(value=23.523)
var.value
```

versus:
```python
var = TimeArray(array=23.532)
var.array
```

dylanhmorris: One could even image having a time-unaware base `SampledValue` class, and adding the `t_start` and `t_unit` attributes only in some subclass `TimeAwareSampledValue` (versus just having `SampledValue`s have the option of `t_start = None` and `t_unit = None`). But that might add extra hierarchy/complexity with minimal benefit

gvegayon: All now is SampledValue

gvegayon:
Ready for review, @damonbayer and @dylanhmorris

gvegayon:
Fixed the latest conflicts, @damonbayer and @dylanhmorris.

model/src/pyrenew/process/periodiceffect.py
```diff
@@ -110,9 +114,11 @@ def sample(self, duration: int, **kwargs):
         """
 
         return PeriodicEffectSample(
-            value=self.broadcaster(
-                data=self.quantity_to_broadcast.sample(**kwargs)[0],
-                n_timepoints=duration,
+            value=SampledValue(
```
damonbayer, line 117: Should this return a `SampledValue`, not an `PeriodicEffectSample`?

model/src/pyrenew/process/rtperiodicdiff.py
```diff
@@ -188,7 +192,9 @@ def sample(
         )[0]
 
         return RtPeriodicDiffProcessSample(
-            rt=self.broadcaster(jnp.exp(log_rt.flatten()), duration),
+            rt=SampledValue(
```
damonbayer, line 195: Same as before, should we remove the extra layer of abstraction and just return a named tuple with a `SampledValue`?

model/src/pyrenew/process/periodiceffect.py
damonbayer: Actually, I think the plan was to return a regular tuple, and later have a PR to convert everything to named tuples.

model/src/pyrenew/model/admissionsmodel.py
damonbayer, line 1: I think the types in `HospModelSample(NamedTuple)` are now incorrect. We are returning `SampledValue`s instead of `ArrayLike`s.

model/src/pyrenew/model/rtinfectionsrenewalmodel.py
damonbayer, line 1: Update types in `class RtInfectionsRenewalSample(NamedTuple)`

model/src/pyrenew/metaclass.py
```diff
@@ -94,6 +94,46 @@ def _assert_sample_and_rtype(
...
+    ) -> None:
+        """
+        Default constructor for SampledValue
+
+        Parameters
+        ----------
+        value : ArrayLike
+            The sampled value.
+        t_start : int, optional
+            The start time of the data.
```
damonbayer, line 116:
```suggestion
            The start time of the value.
```

model/src/pyrenew/latent/infectionswithfeedback.py
damonbayer, line 1: Update types in `class InfectionsRtFeedbackSample(NamedTuple)`

docs/source/tutorials/time.qmd
```diff
@@ -10,7 +10,7 @@ The fundamental time unit should represent a period of fixed (or approximately f
 
 For many infectious disease renewal models of interest, the fundamental time unit will be days, and we will proceed with this tutorial treating days as our fundamental unit.
 
- `pyrenew` deals with time having `RandomVariable`s carry information about (i) their own time unit expressed relative to the fundamental unit (`t_unit`) and (ii) the starting time, `t_start`, measured relative to `t = 0` in model time in fundamental time units.
+ `pyrenew` deals with time having `RandomVariable`s carry information about (i) their own time unit expressed relative to the fundamental unit (`t_unit`) and (ii) the starting time, `t_start`, measured relative to `t = 0` in model time in fundamental time units. Moreover, return values from `RandomVariable.sample()` are namedtuples with `SampledValue` objects that carry the same information.
```
damonbayer, line 13: This is not currently true. We don't always return NameedTuples

damonbayer (changes requested):
Just a few things to clean up.

docs/source/tutorials/time.qmd
gvegayon: True. Will correct.

model/src/pyrenew/latent/infectionswithfeedback.py
gvegayon: Noted

model/src/pyrenew/deterministic/deterministic.py
```diff
@@ -83,8 +90,14 @@ def sample(
         Returns
         -------
         tuple
-            Containing the stored values during construction.
+            Containing the stored values during construction wrapped in a SampledValue.
```
dylanhmorris, line 93:
```suggestion
            A length-one tuple whose single entry is a :class:`SampledValue` 
            instance with `value=self.vars`, `t_start=self.t_start`, and `t_unit=self.t_unit`.
```

model/src/pyrenew/process/periodiceffect.py
gvegayon: PeriodicEffectSample is (was) a namedtuple before this PR. So there's no need to change here.

model/src/pyrenew/process/rtperiodicdiff.py
gvegayon: Same as my last comment, nothing to do here.

model/src/pyrenew/deterministic/deterministic.py
```diff
@@ -28,13 +30,18 @@ def __init__(
             A tuple with arraylike objects.
```
dylanhmorris, line 30:
```suggestion
            An arraylike object.
```

dylanhmorris: We should discuss elsewhere whether `DeterministicVariable`s should be able to have an arbitrary `pytree` as their `vars` value. I lean yes.

I would also support renaming `vars` to `value`, for consistency with `SampledValue`.

For clarity, both of the above are out of scope for this PR.

model/src/pyrenew/deterministic/deterministic.py
```diff
@@ -83,8 +90,14 @@ def sample(
...
         if record:
             npro.deterministic(self.name, self.vars)
-        return (self.vars,)
+        return (
+            SampledValue(
+                value=self.vars,
+                t_start=self.t_start,
+                t_unit=self.t_unit,
+            ),
+        )
```
dylanhmorris, line 103: Noting here that this will be the gotcha of the new system. If I have 

```python
some_time_aware_sampled_value = stochastic_rv.sample()
```
and I want to pass it somewhere else downstream as a `DeterministicVariable`, I should _not_ do:

```python
my_det_var = DeterministicVariable(vars=some_time_aware_sampled_value,
                                                             name="my_det_var")
```

Instead one should do:

```python
my_det_var = DeterministicVariable(
               vars=some_time_aware_sampled_value.value,
               name="my_det_var", 
               t_start=some_time_aware_sampled_value.t_start,
               t_unit=some_time_aware_sampled_value.t_unit)
```

model/src/pyrenew/model/rtinfectionsrenewalmodel.py
gvegayon: Noted

model/src/pyrenew/process/periodiceffect.py
damonbayer: I would prefer to remove `PeriodicEffectSample` and just return a namedtuple without a special class.

model/src/pyrenew/deterministic/deterministic.py
dylanhmorris: Note that if we do #307 and also take my suggestion above of renaming the `vars` attribute of `DeterministicVariable` to `value`, then we could use unpacking here to turn `SampledValues` into `DeterministicVariables`:

```python
my_det_var = DeterministicVariable(
               name="my_det_var", 
              **some_time_aware_sampled_value)
```
I am on the fence RE: whether this is nicely succinct or too opaque/magic.

dylanhmorris: In any case we should document the gotcha somewhere, perhaps.

model/src/pyrenew/deterministic/deterministicpmf.py
```diff
@@ -18,6 +18,8 @@ def __init__(
         vars: ArrayLike,
         name: str,
         tol: float = 1e-5,
```
dylanhmorris, line 20: Out of scope here but this should presumably have a default set higher up. See also #268

model/src/pyrenew/deterministic/deterministicpmf.py
```diff
@@ -46,7 +53,12 @@ def __init__(
             tol=tol,
         )
 
-        self.basevar = DeterministicVariable(vars, name)
+        self.basevar = DeterministicVariable(
+            vars=vars,
+            name=name,
+            t_start=t_start,
+            t_unit=t_unit,
+        )
```
dylanhmorris, line 61: Beyond the scope of this PR but I think this is most efficiently handled by actual inheritance, e.g. something like:

```python
def __init__(
       self, 
       name: str
       pmf: ArrayLike,
       tol: float = 1e-5,
       **kwargs):
    normed_pmf = validate_discrete_dist_vector(pmf, tol)
    super().__init__(vars=normed_pmf, **kwargs)
```
This avoids having to override the `sample()` method, among other things.

model/src/pyrenew/deterministic/nullrv.py
```diff
@@ -46,10 +47,10 @@ def sample(
...
         tuple
-            Containing None.
+            Containing a SampledValue with None.
         """
 
-        return (None,)
+        return (SampledValue(None),)
 
 
 class NullProcess(NullVariable):
```
dylanhmorris, line 56: Is `NullProcess` still used? It seems to be completely identical to `NullVariable`. Artifact of when there were `RandomProcess`es that were separate/distinct from `RandomVariable`s?

model/src/pyrenew/deterministic/nullrv.py
```diff
@@ -95,10 +96,10 @@ def sample(
...
         tuple
-            Containing None.
+            Containing a SampledValue with None.
         """
 
-        return (None,)
+        return (SampledValue(None),)
 
 
 class NullObservation(NullVariable):
```
dylanhmorris, line 105: Similarly, is this still used?

model/src/pyrenew/deterministic/process.py
```diff
@@ -2,6 +2,7 @@
 
 import jax.numpy as jnp
 from pyrenew.deterministic.deterministic import DeterministicVariable
+from pyrenew.metaclass import SampledValue
 
 
 class DeterministicProcess(DeterministicVariable):
```
dylanhmorris, line 8: I'd also suggest reviewing the implementation here. I think given some other choices we've made we'll perhaps want to modify it. Out of scope for this PR, however.

model/src/pyrenew/deterministic/process.py
```diff
@@ -29,14 +30,26 @@ def sample(
...
+                ),
+            )
+
+        return (
+            SampledValue(
+                value=res.value[:duration],
+                t_start=self.t_start,
+                t_unit=self.t_unit,
+            ),
+        )
```
dylanhmorris, line 55: Double return here. Typo?

model/src/pyrenew/latent/hospitaladmissions.py
```diff
@@ -162,7 +162,7 @@ def sample(
 
         Parameters
         ----------
-        latent : ArrayLike
+        latent : ArrayLike or SampledValue
```
dylanhmorris, line 165:
```suggestion
        latent_infections : ArrayLike
```
As the code written, this cannot be a `SampledValue`, since we do not call `.value` on it. And I think that's at least arguably correct

model/src/pyrenew/latent/hospitaladmissions.py
```diff
@@ -184,25 +184,33 @@ def sample(
...
             self.latent_hospital_admissions_varname, latent_hospital_admissions
         )
 
         return HospitalAdmissionsSample(
-            infection_hosp_rate, latent_hospital_admissions
+            infection_hosp_rate=infection_hosp_rate,
+            latent_hospital_admissions=SampledValue(
+                value=latent_hospital_admissions,
+                t_start=self.infection_to_admission_interval_rv.t_start,
+                t_unit=self.infection_to_admission_interval_rv.t_unit,
```
dylanhmorris, line 214: I don't think these are the relevant t_start/t_unit.

I would have thought they should be the same as the RV itself (or, eventually computed dynamically from the values in a time-aware `latent_infections` that gets passed + attributes of the delay interval, if we eventually go in that direction). For now, suggest:

```suggestion
                t_start=self.t_start,
                t_unit=self.t_unit,
```

model/src/pyrenew/latent/infection_initialization_method.py
```diff
@@ -177,6 +177,7 @@ def initialize_infections(self, I_pre_init: ArrayLike):
                 f"I_pre_init must be an array of size 1. Got size {I_pre_init.size}."
             )
         (rate,) = self.rate()
+        rate = rate.value
```
dylanhmorris, line 180:
```suggestion
        rate = self.rate()[0].value
```

dylanhmorris: Longer term this should be refactored to follow the [variable naming conventions](https://github.com/CDCgov/multisignal-epi-inference/blob/main/docs/source/developer_documentation.rst#pyrenew-principles), which would allow us to do something like:

```python
rate = self.rate_rv()[0].value
```

model/src/pyrenew/latent/infectionswithfeedback.py
```diff
@@ -159,6 +163,7 @@ def sample(
         inf_feedback_strength, *_ = self.infection_feedback_strength(
```
dylanhmorris, line 163: similar comment to sampling `rate` above.

model/src/pyrenew/metaclass.py
```diff
@@ -94,6 +94,46 @@ def _assert_sample_and_rtype(
...
+
+        Returns
+        -------
+        None
+        """
+
+        if value is not None:
+            assert isinstance(
+                value, ArrayLike
+            ), "value should be an array-like object."
```
dylanhmorris, line 128: Are we sure we want this (versus arbitrary `pytrees`?

model/src/pyrenew/metaclass.py
```diff
@@ -94,6 +94,46 @@ def _assert_sample_and_rtype(
     return None
 
 
+class SampledValue:
```
dylanhmorris, line 97: I thought the plan was to have `SampledValue`s be `NamedTuple`s to ensure that they are immutable.

dylanhmorris (changes requested):
Stopping a bit of the way through because I have a fairly major question. Here, `SampledValue` is a regular class. I had thought the plan was to have `SampledValue`s be NamedTuples to ensure that they are immutable.

model/src/pyrenew/metaclass.py
gvegayon: Good question. I would say yes, keep things as ArrayLike objects, but pytrees should be, in principle, OK. Since I'm making the SampledValue a namedtuple, I'm taking this out and we can leave it for a later discussion if we want to enforce a type.

model/src/pyrenew/deterministic/deterministic.py
gvegayon: Issue created

model/src/pyrenew/deterministic/deterministic.py
gvegayon: The current implementation checks that `vars` are indeed `ArrayLike`, which the `SampledValue` should not be, so I don't think this will cause such issues.

model/src/pyrenew/deterministic/deterministicpmf.py
gvegayon: Smells like a wider discussion, if we consider the current preference of using composition instead of inheritance when building models.

model/src/pyrenew/deterministic/nullrv.py
gvegayon: > Is `NullProcess` still used? It seems to be completely identical to `NullVariable`. Artifact of when there were `RandomProcess`es that were separate/distinct from `RandomVariable`s?

Unsure if this is used right now. The main difference is that `process.*` has the `duration` argument. At some point, the idea of making process, latent, observed, etc. different animals came up. This could add to it.

model/src/pyrenew/deterministic/nullrv.py
gvegayon: Same as before.

model/src/pyrenew/deterministic/process.py
gvegayon: Agreed

model/src/pyrenew/deterministic/process.py
gvegayon: I don't think so. Both return are different.

model/src/pyrenew/latent/hospitaladmissions.py
gvegayon: Good catch.

model/src/pyrenew/deterministic/process.py
dylanhmorris: Oh I see one is inside of an `if`. For clarity, I'd suggest this pattern

```python
if <condition>:
      result = <value>
else:
     result = <other value>
return result
```

docs/source/tutorials/time.qmd
gvegayon:
```suggestion
 `pyrenew` deals with time having `RandomVariable`s carry information about (i) their own time unit expressed relative to the fundamental unit (`t_unit`) and (ii) the starting time, `t_start`, measured relative to `t = 0` in model time in fundamental time units. Return values from `RandomVariable.sample()` are `tuples` or `namedtuple`s of `SampledValue` objects. Each such `SampledValue` is optionally time-aware with specifiable `t_start` and `t_unit` attributes.
```

model/src/test/test_infectionsrtfeedback.py
```diff
@@ -157,3 +157,5 @@ def test_infectionsrtfeedback_feedback():
     assert_array_almost_equal(samp1.rt.value, res["rt"])
 
     return None
+
+test_infectionsrtfeedback()
```
damonbayer:
```suggestion

```

gvegayon: Yes, also noted that. Solved now.

model/src/pyrenew/deterministic/deterministic.py
dylanhmorris: I was thinking more about usage guidance than erroring, though the error is nice to have.

model/src/pyrenew/observation/negativebinomial.py
```diff
@@ -93,8 +93,9 @@ def sample(
...
             fn=dist.NegativeBinomial2(
                 mean=mu + self.eps,
-                concentration=concentration,
+                concentration=concentration.value,
             ),
             obs=obs,
         )
-        return (negative_binomial_sample,)
+
+        return (SampledValue(negative_binomial_sample),)
```
dylanhmorris, line 101: I think we should probably pass `self.t_start` and `self.t_unit` here, even if they're typically going to be `None`.

model/src/pyrenew/observation/poisson.py
```diff
@@ -72,4 +72,4 @@ def sample(
             fn=dist.Poisson(rate=mu + self.eps),
             obs=obs,
         )
-        return (poisson_sample,)
+        return (SampledValue(poisson_sample),)
```
dylanhmorris, line 75: As above, I think we should probably pass `self.t_start` and `self.t_unit` here, even if they're typically going to be `None`.

model/src/pyrenew/process/ar.py
```diff
@@ -91,7 +91,7 @@ def _ar_scanner(carry, next):  # numpydoc ignore=GL08
         )
 
         last, ts = lax.scan(_ar_scanner, inits - self.mean, noise)
-        return (jnp.hstack([inits, self.mean + ts.flatten()]),)
+        return (SampledValue(jnp.hstack([inits, self.mean + ts.flatten()])),)
```
dylanhmorris, line 94: As above, I think we should probably pass `self.t_start` and `self.t_unit` here, even if they're typically going to be `None`.

model/src/pyrenew/process/firstdifferencear.py
```diff
@@ -72,7 +72,11 @@ def sample(
             inits=jnp.atleast_1d(init_rate_of_change),
             name=name + "_rate_of_change",
         )
-        return (init_val + jnp.cumsum(rates_of_change.flatten()),)
+        return (
+            SampledValue(
+                init_val + jnp.cumsum(rates_of_change.value.flatten())
+            ),
+        )
```
dylanhmorris, line 79: As above, I think we should probably pass `self.t_start` and `self.t_unit` here, even if they're typically going to be `None`.

model/src/pyrenew/process/periodiceffect.py
```diff
@@ -110,9 +113,11 @@ def sample(self, duration: int, **kwargs):
...
+                self.broadcaster(
+                    data=self.quantity_to_broadcast.sample(**kwargs)[0].value,
+                    n_timepoints=duration,
```
dylanhmorris, line 119: We definitely want these SampledValues be time-aware, I'd think. For now I'd use `self.t_start` and `self.t_unit`, but we may want functionality to optionally compute those at construction from a time-aware `quantity_to_broadcast`

model/src/pyrenew/process/rtperiodicdiff.py
```diff
@@ -188,7 +192,9 @@ def sample(
...
+                self.broadcaster(jnp.exp(log_rt.value.flatten()), duration)
+            ),
```
dylanhmorris, line 197: As above, I think we should probably pass `self.t_start` and `self.t_unit` here.

model/src/pyrenew/process/simplerandomwalk.py
```diff
@@ -77,16 +77,16 @@ def sample(
...
 
         _, x = scan(
             transition,
-            init=init,
+            init=init.value,
             xs=jnp.arange(n_steps - 1),
         )
 
-        return (jnp.hstack([init, x.flatten()]),)
+        return (SampledValue(jnp.hstack([init.value, x.flatten()])),)
```
dylanhmorris, line 89: As above, I think we should probably pass `self.t_start` and `self.t_unit` here, even if they're sometimes going to be `None`.

model/src/test/test_periodiceffect.py
```diff
@@ -44,7 +44,7 @@ def test_periodiceffect() -> None:
     params["offset"] = 5
     pe = PeriodicEffect(**params)
     with npro.handlers.seed(rng_seed=np.random.randint(1, 600)):
-        ans2 = pe(duration=duration).value
+        ans2 = pe(duration=duration).value.value
```
dylanhmorris, line 47: I think this has implications for our naming conventions for the `NamedTuples` returned by `RandomVariable.sample()`. For now I would suggest this:
```suggestion
        ans2 = pe(duration=duration)[0].value
```

model/src/test/test_periodiceffect.py
```diff
@@ -81,9 +81,13 @@ def test_weeklyeffect() -> None:
     pe = PeriodicEffect(**params)
     pe2 = DayOfWeekEffect(**params2)
 
-    ans1 = pe(duration=duration).value
-    ans2 = pe2(duration=duration).value
+    ans1 = pe(duration=duration).value.value
+    ans2 = pe2(duration=duration).value.value
```
dylanhmorris, line 85: As above

model/src/test/test_periodiceffect.py
```diff
@@ -81,9 +81,13 @@ def test_weeklyeffect() -> None:
...
 
     assert_array_equal(ans1, ans2)
 
     return None
+
+
+test_periodiceffect()
+test_weeklyeffect()
```
dylanhmorris, line 93:
```suggestion
```

dylanhmorris (changes requested):
I think we almost always want `RandomVariables` to pass their time information (if any) forward in `SampledValue`s (default unmodified; modified in specific cases where it makes sense)

model/src/pyrenew/observation/negativebinomial.py
gvegayon: Noted

model/src/pyrenew/observation/poisson.py
gvegayon: Noted

model/src/pyrenew/process/periodiceffect.py
gvegayon: Created an issue to list all components that will need to be explicitly timeaware: #320. Added `PeriodicEffect` to the list. 

att @damonbayer

model/src/pyrenew/process/rtperiodicdiff.py
gvegayon: Noted

model/src/pyrenew/process/simplerandomwalk.py
gvegayon: Noted

model/src/test/test_periodiceffect.py
gvegayon: Noted

gvegayon:
It should be good to go now.

model/src/pyrenew/process/ar.py
gvegayon: Noted

dylanhmorris:
Looks like some pre-commit and test fixes still needed @gvegayon

model/src/pyrenew/deterministic/deterministic.py
dylanhmorris: Sounds good. Flag as todo in #304?

model/src/pyrenew/latent/infectionswithfeedback.py
```diff
@@ -195,6 +199,8 @@ def sample(
         numpyro.deterministic("Rt_adjusted", Rt_adj)
 
         return InfectionsRtFeedbackSample(
-            post_initialization_infections=post_initialization_infections,
-            rt=Rt_adj,
+            post_initialization_infections=SampledValue(
+                post_initialization_infections
```
dylanhmorris, line 203:
```suggestion
            post_initialization_infections=SampledValue(
                value=post_initialization_infections,
                t_start=self.t_start,
                t_unit=self.t_unit)
```

dylanhmorris (changes requested):
A few remaining tweaks

dylanhmorris:
This should be ready for review @damonbayer. @AFg6K7h4fhy2 I would like to note that I introduced a typo while resolving merge conflicts and the typos hook caught it, so thanks again for setting that up/troubleshooting it.

model/src/pyrenew/metaclass.py
```diff
@@ -94,6 +94,28 @@ def _assert_sample_and_rtype(
     return None
 
 
+class SampledValue(NamedTuple):
+    """
+    A container for a sampled value from a RandomVariable.
+
+    Attributes
+    ----------
+    value : ArrayLike, optional
```
damonbayer, line 103: When would `value` be optional?

docs/source/tutorials/time.qmd
```diff
@@ -10,9 +10,9 @@ The fundamental time unit should represent a period of fixed (or approximately f
 
 For many infectious disease renewal models of interest, the fundamental time unit will be days, and we will proceed with this tutorial treating days as our fundamental unit.
 
- `pyrenew` deals with time having `RandomVariable`s carry information about (i) their own time unit expressed relative to the fundamental unit (`t_unit`) and (ii) the starting time, `t_start`, measured relative to `t = 0` in model time in fundamental time units.
+`pyrenew` deals with time having `RandomVariable`s carry information about (i) their own time unit expressed relative to the fundamental unit (`t_unit`) and (ii) the starting time, `t_start`, measured relative to `t = 0` in model time in fundamental time units. Return values from `RandomVariable.sample()` are `tuples` or `namedtuple`s of `SampledValue` objects. Each such `SampledValue` is optionally time-aware with specifiable `t_start` and `t_unit` attributes.
```
damonbayer, line 13: `RandomVariable`s having their own `t_start` and `t_unit` seems to contrast with the idea the `RandomVariable.sample()` can return multiple `SampledValue`s with different `t_start` and `t_unit`.

model/src/pyrenew/deterministic/deterministic.py
```diff
@@ -75,16 +81,27 @@ def sample(
         Parameters
         ----------
         record : bool, optional
-            Whether to record the value of the deterministic RandomVariable. Defaults to True.
+            Whether to record the value of the deterministic
```
damonbayer, line 84: Why does this line change? Can we check this type of formatting with precommit?

model/src/pyrenew/deterministic/deterministic.py
```diff
@@ -75,16 +81,27 @@ def sample(
...
             sample calls, should there be any.
 
         Returns
         -------
-        tuple
-            Containing the stored values during construction.
+        tuple[SampledValue]
+            A length-one tuple whose single entry is a
+            :class:`SampledValue`
+            instance with `value=self.vars`,
```
damonbayer, line 95: I think we no longer use `vars`.

model/src/pyrenew/metaclass.py
damonbayer: Answering my own question, in `model/src/pyrenew/deterministic/nullrv.py`.

model/src/pyrenew/deterministic/process.py
```diff
@@ -29,14 +30,28 @@ def sample(
         Returns
         -------
         tuple
-            Containing the stored values during construction.
+            Containing the stored values during construction wrapped in a SampledValue.
```
damonbayer, line 33: This doesn't read well to me. Not sure what it is trying to say.

model/src/pyrenew/deterministic/process.py
```diff
@@ -29,14 +30,28 @@ def sample(
...
+        else:
+            res = (
+                SampledValue(
+                    value=res.value[:duration],
+                    t_start=self.t_start,
+                    t_unit=self.t_unit,
+                ),
+            )
+
+        return res
```
damonbayer, line 57:
```suggestion
            value = jnp.hstack([res.value, jnp.repeat(res.value[-1], dif)])
        else:
            value = res.value[:duration]

        res = SampledValue(
            value,
            t_start=self.t_start,
            t_unit=self.t_unit,
        )

        return (res,)
```

model/src/pyrenew/latent/hospitaladmissions.py
```diff
@@ -20,12 +20,12 @@ class HospitalAdmissionsSample(NamedTuple):
     ----------
     infection_hosp_rate : float, optional
```
damonbayer, line 21: should this be a `SampledValue`?

model/src/pyrenew/process/firstdifferencear.py
```diff
@@ -75,7 +75,13 @@ def sample(
             duration=duration,
             inits=jnp.atleast_1d(init_rate_of_change),
         )
-        return (init_val + jnp.cumsum(rates_of_change.flatten()),)
+        return (
+            SampledValue(
+                init_val + jnp.cumsum(rates_of_change.value.flatten()),
```
damonbayer, line 80: Is the `.flatten` neccessar?

model/src/pyrenew/deterministic/deterministic.py
dylanhmorris: That was me disliking the long line. But yes, we should be able to set a maximum docstring line length etc

model/src/pyrenew/process/firstdifferencear.py
dylanhmorris: Can investigate in #328

model/src/pyrenew/latent/hospitaladmissions.py
dylanhmorris: Yes it should. Fixed in 2177215

model/src/pyrenew/deterministic/process.py
dylanhmorris: Good edit, thanks!

model/src/pyrenew/deterministic/process.py
dylanhmorris: Made an attempt in 9ff9dea but I increasingly think we should get rid of the `DeterministicProcess` class, as I think it should be replaced by other more time-aware ways of handling temporal broadcasting of values. Beyond the scope of this PR, though.

model/src/pyrenew/deterministic/deterministic.py
dylanhmorris: Good catch, thanks. Fixed in b298bc2

damonbayer (approved):

docs/source/tutorials/time.qmd
dylanhmorris: Attempted clarification in 859dbc9: 

https://github.com/CDCgov/multisignal-epi-inference/blob/859dbc96712a887bf2a5daa98156bebbedbc8dab/docs/source/tutorials/time.qmd#L13-L15

dylanhmorris: In the longer run, we should add an example of the second thing to the tutorial, but I think that can wait for the next time-handling PR.

dylanhmorris:
Ready for re-review, @damonbayer

docs/source/tutorials/time.qmd
```diff
@@ -10,7 +10,9 @@ The fundamental time unit should represent a period of fixed (or approximately f
 
 For many infectious disease renewal models of interest, the fundamental time unit will be days, and we will proceed with this tutorial treating days as our fundamental unit.
 
-`pyrenew` deals with time having `RandomVariable`s carry information about (i) their own time unit expressed relative to the fundamental unit (`t_unit`) and (ii) the starting time, `t_start`, measured relative to `t = 0` in model time in fundamental time units. Return values from `RandomVariable.sample()` are `tuples` or `namedtuple`s of `SampledValue` objects. Each such `SampledValue` is optionally time-aware with specifiable `t_start` and `t_unit` attributes.
+`pyrenew` deals with time having `RandomVariable`s carry information about (i) their own time unit expressed relative to the fundamental unit (`t_unit`) and (ii) the starting time, `t_start`, measured relative to `t = 0` in model time in fundamental time units. Return values from `RandomVariable.sample()` are `tuples` or `namedtuple`s of `SampledValue` objects. `SampledValue` objects can have `t_start` and `t_unit` attributes.
```
damonbayer, line 13:
```suggestion
`pyrenew` deals with time by having `RandomVariable`s carry information about

1. their own time unit expressed relative to the fundamental unit (`t_unit`) and
2. the starting time, `t_start`, measured relative to `t = 0` in model time in fundamental time units.

Return values from `RandomVariable.sample()` are `tuples` or `namedtuple`s of `SampledValue` objects. `SampledValue` objects can have `t_start` and `t_unit` attributes.
```

PR #264, opened 2024-07-11
Updating instructions on dev docs and moving Make targets

gvegayon:
- Updates the instructions under `docs/source/developer_documentation.rst` reflecting where the tutorials live now.
- Moves all the `Makefile` targets under `./model/Makefile` to `./docs/source/tutorials/Makefile` (new Makefile).


docs/source/tutorials/Makefile
dylanhmorris, line 1: Should this be its own thing or should it be part of `docs/Makefile`? As written, `docs/Makefile` seems to have commands that do nothing, and I think we should avoid recursive make where possible.

dylanhmorris (commented):
One key question @gvegayon

docs/source/tutorials/Makefile
dylanhmorris: @gvegayon flagging this

gvegayon: The `docs/Makefile` was originally added with the Sphinx website. I tried to add the targets to build the qmd files, but that broke the Sphinx pipeline. The lines adding print info to `make help` are residual from that attempt. I've corrected that. The `docs/source/tutorials/Makefile` is a helper that contains targets to render and convert the qmd files to py and ipynb. Is the same file that used to live under `model/docs/`.

dylanhmorris: In that case I suggest renaming the `.PHONY` targets from `docs_*` to `tutorials_*` for clarity.

dylanhmorris (changes requested):
@gvegayon one suggestion

dylanhmorris:
@gvegayon looks like this needs a pre-commit fix but then should be ready for re-review?

dylanhmorris (approved):

PR #265, opened 2024-07-12
Format Quarto docs in pre-commit

damonbayer:
Closes https://github.com/CDCgov/multisignal-epi-inference/issues/159


quarto_python_formatter.py
```diff
@@ -0,0 +1,71 @@
...
+import tempfile
+from pathlib import Path
+from typing import Match
+
+
+def format_python_code(code: str) -> str:  # numpydoc ignore=RT01
+    """Format Python code using Black."""
+    try:
+        result = subprocess.run(
+            ["black", "-", "-q", "--line-length", "79"],
```
dylanhmorris, line 15: I think it's worth make the formatting call configurable so that things like this can be set in `.pre-commit-config.yaml`

quarto_python_formatter.py
```diff
@@ -0,0 +1,71 @@
...
+    if len(sys.argv) < 2:
+        print(
+            "Usage: python format_qmd_python.py <filename1.qmd> [filename2.qmd ...]"
+        )
+        sys.exit(1)
+
+    for filepath in sys.argv[1:]:
+        path = Path(filepath)
+        if not path.exists():
+            print(f"Error: File {path} does not exist.", file=sys.stderr)
```
dylanhmorris, line 69: Why not actually error here?

dylanhmorris (changes requested):
Looks largely good. A few questions / suggestions

quarto_python_formatter.py
damonbayer: If you passing a bunch of files, but some of them don't exist, it will still process the extant files.

dylanhmorris: Wouldn't you want the pre-commit hook to fail in that case, though?

quarto_python_formatter.py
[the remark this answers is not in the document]
gvegayon: Not sure where, but perhaps somewhere with the `.pre-commit-generate-rst` or whatever that's named. Perhaps there should be like a `utils` or `dev-utils` folder?

quarto_python_formatter.py
damonbayer: addressed in https://github.com/CDCgov/multisignal-epi-inference/pull/265/commits/95efca66f3c3810e13416cc150165f7cf4500b66

quarto_python_formatter.py
damonbayer: addressed in https://github.com/CDCgov/multisignal-epi-inference/pull/265/commits/3c0a47ebab39cbc62df493aa12d06b3b1153d37b

quarto_python_formatter.py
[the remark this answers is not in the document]
damonbayer: addressed in https://github.com/CDCgov/multisignal-epi-inference/pull/265/commits/9b45d6662368d5d718ca5afe3f06fb579e1b39ec

.pre-commit-config.yaml
```diff
@@ -1,56 +1,70 @@
...
-# Secrets
--   repo: https://github.com/Yelp/detect-secrets
+      - id: numpydoc-validation
+  #####
+  # Quarto
+  - repo: local
+    hooks:
+      - id: format-qmd-python
+        name: Format Python in QMD
+        entry: python hook_scripts/quarto_python_formatter.py "-q --line-length 79"
```
dylanhmorris, line 46: This works for me and I think is more readable 
```suggestion
        entry: python hook_scripts/quarto_python_formatter.py 
        args: ["-q --line-length 79"]
```

hook_scripts/quarto_python_formatter.py
```diff
@@ -0,0 +1,79 @@
...
+
+
+if __name__ == "__main__":
+    if len(sys.argv) < 3:
+        print(
+            'Usage: python hook_scripts/quarto_python_formatter.py "BLACK_ARGS" <filename1.qmd> [filename2.qmd ...]'
+        )
+        sys.exit(1)
+
+    black_args = sys.argv[1].split()
```
dylanhmorris, line 70: Long run we may want to move to explicit arg parsing so that the hook's `args` entry can make the expected style of typical pre-commit hook `args:` lists, but that doesn't need to happen now.

dylanhmorris (approved):
One suggestion.

hook_scripts/quarto_python_formatter.py
damonbayer: I could not figure out how to have arbitrarily many `black_args` and `files`.

.pre-commit-config.yaml
damonbayer: I was following advice from pre-commit docs https://pre-commit.com/#arguments-pattern-in-hooks

> When creating local hooks, there's no reason to put command arguments into [args](https://pre-commit.com/#config-args) as there is nothing which can override them -- instead put your arguments directly in the hook [entry](https://pre-commit.com/#hooks-entry).

hook_scripts/quarto_python_formatter.py
dylanhmorris: could handle with `argparser`, but again beyond the scope of this PR.

damonbayer: I tried that for longer than I should have but couldn't get it to work.

PR #266, opened 2024-07-15
Add RV design spec

dylanhmorris:
Draft RV design spec added. Draft PR since I haven't configured it to be part of the official docs, etc, as I want comments first.

### DIFF
--- a/docs/source/random_variable_design.rst
+++ b/docs/source/random_variable_design.rst
@@ -175,7 +175,7 @@ Future possibilities
 ====================
 
 In the future, we might wish to have ``RandomVariable`` constructors
-coerce fixed values to ``DeterminsiticVariables`` via some
+coerce fixed values to ``DeterministicVariables`` via some
 ``ensure_rv()`` function, so that this:
 
 .. code:: python
--- a/docs/source/random_variable_design.rst
+++ b/docs/source/random_variable_design.rst
@@ -1,0 +1,203 @@
+Parameterize each ``RandomVariable`` in terms of other ``RandomVariable``\ s
+============================================================================
+
+If a ``RandomVariable`` ``Var`` needs to know some numerical parameter
+values ``a``, ``b``, etc to perform its ``sample()`` call, those values
+should be represented as constiuent ``RandomVariables`` that are
+attributes of ``Var`` (i.e. ``Var.a_rv``, ``Var.b_rv``) and are are
+``sample()``-d within ``Var.sample()`` (i.e. ``Var.sample()`` gets an
+``a`` value by calling ``self.a_rv.sample()`` and a ``b`` value by
+calling ``self.b_rv.sample()``).
+
+Often, a value ``a`` should to treated as fixed from the point of view
+of ``Var``. Perhaps it is known with near-certainty. Perhaps it is
+unknown, but it is sampled/ inferred elsewhere in a larger model. In
+that case, use the ``DeterministicVariable()`` construct to pass it to
+``Var``, i.e. ``Var.a_rv = DeterministicVariable(fixed_a_value)``. This
+corresponds to the `“maximally random” approach <#maximally-random>`__
+described in detail below.
+
+Introduction to the problem
+---------------------------
+
+Sometimes, a ``RandomVariable`` can only be ``sample()``-d conditional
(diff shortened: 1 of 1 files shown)


docs/source/random_variable_design.rst
```diff
@@ -0,0 +1,203 @@
+Parameterize each ``RandomVariable`` in terms of other ``RandomVariable``\ s
+============================================================================
+
+If a ``RandomVariable`` ``Var`` needs to know some numerical parameter
+values ``a``, ``b``, etc to perform its ``sample()`` call, those values
+should be represented as constiuent ``RandomVariables`` that are
+attributes of ``Var`` (i.e. ``Var.a_rv``, ``Var.b_rv``) and are are
+``sample()``-d within ``Var.sample()`` (i.e. ``Var.sample()`` gets an
+``a`` value by calling ``self.a_rv.sample()`` and a ``b`` value by
+calling ``self.b_rv.sample()``).
```
damonbayer, line 10: I think this might be too strong. Consider or generic observation processes, which have a mean argument. The mean is almost guaranteed to have been sampled by some other, more complex, RandomVariable that should not itself be wrapped by the observation process:

```python
        # Sampling from the latent process
        (
            post_initialization_latent_infections,
            *_,
        ) = self.latent_infections_rv.sample(
            Rt=Rt,
            gen_int=gen_int,
            I0=I0,
            **kwargs,
        )

        observed_infections, *_ = self.infection_obs_process_rv.sample(
            mu=post_initialization_latent_infections[padding:],
            obs=data_observed_infections,
            **kwargs,
        )
```

This document suggests that `post_initialization_latent_infections[padding:]` should be made into a deterministic variable and then passed to the observation process. That seems a bit too much.

dylanhmorris: Concretely, this means allowing random variables to have a mix of parameters defined at construction (which must be `RandomVariable`s, possibly deterministic and parameters passed to the `sample()` call. I would support this, but would specify that such parameters should always be fixed / "observed" / "data" from the RV's point-of-view. This aligns with the `numpyro` pattern of passing `obs` arguments to `numpyro.sample` calls.

gvegayon: I've encountered this too in #262. There's no silver bullet here. What if we need a `TimeArray` to be passed? I think the call to `RV.sample()` should be open to receive whatever (as it does now). Whatever could be a `RandomVariable`, a `scalar, an `array`, a `SampledArray` (aka `TimeArray`).

gvegayon:
I think is looking great, @dylanhmorris. I think this document will be really useful for future users. As with the timearray stuff, I would be OK to merge this as it is: this is our current understanding and agreement of how parametrization should occur, but it can change in the future.

docs/source/random_variable_design.rst
dylanhmorris: We _could_ go stricter though, I do think it's worth thinking about the potential benefits and costs. Will add some thoughts to the document.

AFg6K7h4fhy2: > We could go stricter though, I do think it's worth thinking about the potential benefits and costs. Will add some thoughts to the document.

I agree with Damon & George's earlier sentiments on this but also do think we could go stricter in certain instances, based on what we know about the RVs parameters and potentially also the complexity of the RV / where it lies in the modeling process. Some additional examples might be helpful showcasing the verbosity of the maximally random approach in situations where it feels like "too much"; if these MSR examples end up seeming "not too bad", then I might be more convinced to operate in a maximally random way with MSR.

gvegayon:
Do you need anything from me, @dylanhmorris?

docs/source/random_variable_design.rst
damonbayer: In recent conversations, @dylanhmorris seem to have come to an agreement that RandomVariables should only have a single stochastic element, as a rule of thumb.

damonbayer:
I don't think the existing writeup is very reflective of our current thinking. A revised write up should be based on the recent developments of dynamic and static distributional random variables (https://github.com/CDCgov/multisignal-epi-inference/pull/391)   @dylanhmorris Do you want to continue working in this branch or should we close this PR?

dylanhmorris:
Closing pending further design spec decisions to be made downstream of #407

Issue #178, opened 2024-06-12
Make Concentration in `NegativeBinomialObservation` a RandomVariable


damonbayer:

damonbayer:
In working through this, I am realizing that we may have to forgo the concept of a generic observation process as an argument to a `Model` (currently `NegativeBinomialObservation` and `PoissonObservation`). These were nice because they both took a `mu` and `obs` argument to their `sample` call.

We currently have
```python
NegativeBinomialObservation(concentration=1.0)
```

I initially thought to implement something like 
```python
NegativeBinomialObservation(concentration_rv = some_random_variable)
```

Then `NegativeBinomialObservation.sample` would include
```python
concentration = concentration_rv.sample()
```

But then I realized this would sample a new concentration variable for every observation, which we do not want.

I think the only way around this is to `sample` the `concentration_rv` in the `Model.sample`, which is incompatible with the idea of an interchangeable observation process. So, I have concluded that the generic observation process should be removed. Am I missing anything @gvegayon @dylanhmorris ?

Issue #198, opened 2024-06-17
Remove default name arguments for NumPyro sites from non-model classes


damonbayer:
Some code has default names for NumPyro sites baked in.

Ex: 

https://github.com/CDCgov/multisignal-epi-inference/blob/428bcc4f3ba4840d8209653ecbc285702141c209/model/src/pyrenew/observation/negativebinomial.py#L17-L23

This could lead to problems if there are multiple random variables constructed from the same class. At this low level, we should require users to specify the site names.

PR #267, opened 2024-07-15
Make concentration in `NegativeBinomialObservation` a `RandomVariable`

damonbayer:
Closes https://github.com/CDCgov/multisignal-epi-inference/issues/178.
Partially addresses https://github.com/CDCgov/multisignal-epi-inference/issues/198.

I now also estimate the overdipsersion parameter in `hospital_admissions_model.qmd`.
This makes the inference take quite a while, so maybe we should look at reducing the scope of that tutorial.
It leads to a nice-looking posterior predictive:
fig-output-posterior-predictive-output-1

But funky looking posterior latent infections:
fig-output-infections-distribution-output-1

Not sure if this is just an artifact of the hdi plots or if it indicates something undesirable about the model.

### DIFF
--- a/docs/source/tutorials/basic_renewal_model.qmd
+++ b/docs/source/tutorials/basic_renewal_model.qmd
@@ -132,7 +132,7 @@ rt_proc = RtRandomWalkProcess(
 latent_infections = Infections()
 
 # (5) The observed infections process (with mean at the latent infections)
-observation_process = PoissonObservation()
+observation_process = PoissonObservation("poisson_rv")
 ```
 
 With these five pieces, we can build the basic renewal model as an instance of the  `RtInfectionsRenewalModel` class:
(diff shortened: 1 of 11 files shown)


sbidari:
@damonbayer did you also look at equal tailed intervals for the posterior latent infection?

damonbayer:
> @damonbayer did you also look at equal tailed intervals for the posterior latent infection?

No. I would accept a contribution for that figure, though :)

dylanhmorris:
Caught the problem in #269 from looking at the example figures highlighted here by @damonbayer. Suggest landing that first and merging it in here.

damonbayer:
@dylanhmorris https://github.com/CDCgov/multisignal-epi-inference/pull/269 is merged here

dylanhmorris (approved):
LGTM

sbidari:
The posterior plots in this tutorial doesn't look right. Looks like the posterior samples are all just same array?
Capture1
and the diagnostics are bad
Capture

damonbayer:
There is some variability in latent infections (see my original post), but I agree it is strange.

sbidari:
There is no within chain variability in the posterior samples

dylanhmorris:
The prior on the concentration is not working as written. I tried swapping in a LogNormal prior and things just worked. Will investigate further.

PR #270, opened 2024-07-16
Avoid None encountered in jnp array warning

sbidari:
Add condition for padding `observed_infection` in `rtinfectionsrenewalmodel.py` to avoid padding the `observed_infection` when the `infection_obs_process_rv` is `None` and prevent the `None` encountered warning.

### DIFF
--- a/docs/source/tutorials/basic_renewal_model.qmd
+++ b/docs/source/tutorials/basic_renewal_model.qmd
@@ -268,6 +268,7 @@ az.plot_trace(
     coords={"Rt_dim_0": np.arange(10)},
     compact=False,
 )
+plt.show()
 ```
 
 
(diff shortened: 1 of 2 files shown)


damonbayer (changes requested):
Thanks for figuring this out @sbidari! I think we can remove the the `plot.show()` lines you added, though.

sbidari:
I think we either need `plot.show()` or a semi-colon on the end of the code on the cells with plots. Quarto didn't like it when there were multiple outputs from a single code cell (at least while rendering the docs locally on my computer)

damonbayer:
> edit: I realize that the addition of `plot.show()` is totally unrelated to the issue in this PR and probably would have been better as it's own PR. I noticed the images were not rendering properly while I was working on this and fixed it.

Interesting. Please create new issue and/or PR to describe and fix this problem.

damonbayer (approved):

PR #272, opened 2024-07-16
fix image rendering in basic renewal model tutorial

sbidari:
adds `plt.show()` to ensure that images render properly in the webpage


dylanhmorris (approved):
Checked the CI rendered version and it looks right to me. Code also looks right. Thanks for the fix, @sbidari!

Issue #188, opened 2024-06-13
Refactor and rename`RtRandomWalkProcess` and `SimpleRandomWalkProcess` to make init and step variance `RandomVariable`s


damonbayer:
I'm open to this being a separate issue and merging these changes now.

_Originally posted by @damonbayer in https://github.com/CDCgov/multisignal-epi-inference/pull/184#discussion_r1638451661_

>> Request: can we have a forecasting unit test that confirms stochastic processes are projected forward as we would expect (e.g. that forecast-period steps in a random walk are generated conditional on the posterior sampled value of the step s.d.)

> We do not currently support posterior updates of the step s.d., as far as I can tell. We definitely should, but I think that is out of scope for this PR.

_Originally posted by @damonbayer in https://github.com/CDCgov/multisignal-epi-inference/issues/241#issuecomment-2223916492_
            

After discussing with @dylanhmorris, I think we have settled on

- [x] Replace `SimpleRandomWalkProcess` with our own `GaussianRandomWalkProcess` which accepts `RandomVariable`s as arguments for the initial point and the step standard deviation.
- [x] I think just remove the concept of `RtRandomWalk`. We could perform the transformation in the `Model` or make it part of the above `GaussianRandomWalkProcess`.

PR #275, opened 2024-07-18
188 random walk refactor

dylanhmorris:
This addresses #188 and in the process attempts to move things in the direction of the proposed RV design spec from #266


damonbayer:
What is the idea behind `TransformedRandomVariable`? Why not just create a `DistributionalRV` with a [`TransformedDistribution`](https://num.pyro.ai/en/stable/distributions.html#transformeddistribution) as the dist? Or even us [`TransformedDistribution`](https://num.pyro.ai/en/stable/distributions.html#transformeddistribution) as the implementation for the `TransformedRandomVariable` if you are trying to make a different interface.

dylanhmorris:
> What is the idea behind `TransformedRandomVariable`? Why not just create a `DistributionalRV` with a [`TransformedDistribution`](https://num.pyro.ai/en/stable/distributions.html#transformeddistribution) as the dist? Or even use [`TransformedDistribution`](https://num.pyro.ai/en/stable/distributions.html#transformeddistribution) as the implementation for the `TransformedRandomVariable` if you are trying to make a different interface.

We may wish to apply transform the output of `RandomVariable.sample()` calls for RVs other than `DistributionalRV`s. Indeed, in the examples here we use `TransformedRandomVariable` to transform the output of `SimpleRandomWalk`, which is not implemented as a `numpyro.distributions.Distribution`

docs/source/tutorials/hospital_admissions_model.qmd
```diff
@@ -226,6 +250,7 @@ import numpy as np
 
 timeframe = 120
 
+
 with npro.handlers.seed(rng_seed=223):
```
dylanhmorris, line 254:
```suggestion
with numpyro.handlers.seed(rng_seed=223):
```

model/pyproject.toml
```diff
@@ -34,6 +34,7 @@ pytest-cov = "^5.0.0"
 pytest-mpl = "^0.17.0"
 numpydoc = "^1.7.0"
 arviz = "^0.18.0"
+quarto = "^0.1.0"
```
damonbayer, line 37: Why is this now required for `model`?

model/src/pyrenew/metaclass.py
```diff
@@ -209,45 +211,33 @@ def __call__(self, **kwargs):
...
-    ):
+        reparam: Reparam = None,
+    ) -> None:
         """
         Default constructor for DistributionalRV.
 
         Parameters
         ----------
-        dist : npro.distributions.Distribution
+        dist : dist.Distribution
```
damonbayer, line 231:
```suggestion
        dist : numpyro.dist.Distribution
```

model/pyproject.toml
dylanhmorris: Good catch. That should be reverted.

dylanhmorris, line 37:
```suggestion
```

model/src/pyrenew/metaclass.py
dylanhmorris: Fixed.

model/pyproject.toml
dylanhmorris: Fixed, thanks!

dylanhmorris: Now done.

model/src/pyrenew/metaclass.py
```diff
@@ -447,14 +445,14 @@ def print_summary(
...
     ) -> None:
         """
-        A wrapper of MCMC.print_summary
+        A wrapper of :meth:`numpyro.infer.MCMC.print_summary`
 
         Parameters
         ----------
         prob : float, optional
-            The acceptance probability of print_summary. Defaults to 0.9
+            The width of the credible interval to show. Default 0.9
```
damonbayer, line 453: a very strange oversight

model/src/pyrenew/metaclass.py
```diff
@@ -580,3 +581,119 @@ def prior_predictive(
...
+            n_entries = self.base_rv.sample_length()
+            if not n_transforms == n_entries:
+                raise ValueError(
+                    "There must be exactly as many transformations "
+                    "specified as entries self.transforms as there are "
+                    "entries in the tuple returned by "
+                    "self.base_rv.sample()."
+                    f"Got {n_transforms} transforms and {n_entries} "
+                    "entries"
+                )
```
damonbayer, line 699: Could we always accept single transforms and apply it to all `n_entries`? And error if `n_transforms > 1 & n_transforms != n_entries`?

dylanhmorris: I considered that and figured that in general I would want users to be as explicit about transforming multi-output RVs as possible, but I could be persuaded otherwise.

model/src/pyrenew/process/simplerandomwalk.py
```diff
@@ -76,11 +86,12 @@ def transition(x_prev, _):
             xs=jnp.arange(n_timepoints - 1),
         )
 
-        return (jnp.hstack([init, x]),)
+        return (jnp.hstack([init, x.flatten()]),)
```
damonbayer, line 89: Did the previous version lead to any problems?

model/src/pyrenew/process/simplerandomwalk.py
```diff
@@ -61,12 +72,11 @@ def sample(
...
         """
 
-        if init is None:
-            init = npro.sample(name + "_init", self.error_distribution)
+        init, *_ = self.init_rv.sample(**kwargs)
 
         def transition(x_prev, _):
             # numpydoc ignore=GL08
-            diff = npro.sample(name + "_diffs", self.error_distribution)
+            diff, *_ = self.step_rv.sample(**kwargs)
```
damonbayer, line 79:
```suggestion
            diff, *_ = self.step_rv(**kwargs)
```

model/src/pyrenew/process/simplerandomwalk.py
```diff
@@ -61,12 +72,11 @@ def sample(
             With a single array of shape (n_timepoints,).
         """
 
-        if init is None:
-            init = npro.sample(name + "_init", self.error_distribution)
+        init, *_ = self.init_rv.sample(**kwargs)
```
damonbayer, line 75:
```suggestion
        init, *_ = self.init_rv(**kwargs)
```

model/src/pyrenew/process/simplerandomwalk.py
dylanhmorris: Yes, `x` was ending up of shape `(1, n_timepoints)` rather than shape `(n_timepoints,)`

model/src/pyrenew/process/simplerandomwalk.py
```diff
@@ -76,11 +86,12 @@ def transition(x_prev, _):
...
         )
 
-        return (jnp.hstack([init, x]),)
+        return (jnp.hstack([init, x.flatten()]),)
 
     @staticmethod
     def validate():
         """
-        Validates inputted parameters, implementation pending.
+        Validates input parameters, implementation pending.
```
damonbayer, line 94: issue needed?

model/src/pyrenew/process/simplerandomwalk.py
dylanhmorris: It should. I admit to having forgotten about the r.v. `__call__` method when I was implementing this.

model/src/pyrenew/process/simplerandomwalk.py
dylanhmorris: Agree.

model/src/test/test_model_hospitalizations.py
```diff
@@ -21,10 +21,36 @@
...
+    RandomVariable,
+    TransformedRandomVariable,
+)
 from pyrenew.model import HospitalAdmissionsModel
 from pyrenew.observation import PoissonObservation
-from pyrenew.process import RtRandomWalkProcess
+from pyrenew.process import SimpleRandomWalkProcess
+
+
+def get_default_rt():
```
damonbayer, line 34: Should we share this across multiple files?

model/src/test/test_model_hospitalizations.py
```diff
@@ -361,7 +377,7 @@ def test_model_hosp_with_obs_model():
 
     # Sampling and fitting model 0 (with no obs for infections)
     np.random.seed(223)
-    with npro.handlers.seed(rng_seed=np.random.randint(1, 600)):
+    with numpyro.handlers.seed(rng_seed=np.random.randint(1, 600)):
```
damonbayer, line 380: not from this PR, but I think we should do away with this pattern. It is just as "random" as if we did with `numpyro.handlers.seed(rng_seed=223):`

docs/source/tutorials/hospital_admissions_model.qmd
```diff
@@ -179,11 +180,34 @@ I0 = InfectionInitializationProcess(
...
 # Generation interval and Rt
 gen_int = deterministic.DeterministicPMF(gen_int, name="gen_int")
-rtproc = process.RtRandomWalkProcess(
-    Rt0_dist=dist.TruncatedNormal(loc=1.2, scale=0.2, low=0),
-    Rt_transform=transformation.ExpTransform().inv,
-    Rt_rw_dist=dist.Normal(0, 0.025),
-)
+
+
+class MyRt(metaclass.RandomVariable):
```
damonbayer, line 185: Model description above should be updated to reflect this.

docs/source/tutorials/hospital_admissions_model.qmd
damonbayer, line 1: I don't like the log scale for the "Simulated Admissions". I think it is very hard to read.

docs/source/tutorials/basic_renewal_model.qmd
```diff
@@ -305,7 +332,7 @@ axes.plot(x_data, mean_Rt[0], color="C0", label="Mean")
 axes.legend()
 axes.set_title("Posterior Effective Reproduction Number", fontsize=10)
 axes.set_xlabel("Time", fontsize=10)
-axes.set_ylabel("$R_t$", fontsize=10)
+axes.set_ylabel("$\mathcal{R}(t)$", fontsize=10)
```
damonbayer, line 335: There is a warning on the website:
```pycon
<>:29: SyntaxWarning: invalid escape sequence '\m'
<>:29: SyntaxWarning: invalid escape sequence '\m'
/tmp/ipykernel_2398/2465682994.py:29: SyntaxWarning: invalid escape sequence '\m'
  axes.set_ylabel("$\mathcal{R}(t)$", fontsize=10)
```

damonbayer, line 335: This doesn't look right
image

damonbayer: I see above you are trying to ignore `nan` padding, but are we sure there are `nan`s? It looks like we can compute the mean.

docs/source/tutorials/basic_renewal_model.qmd
```diff
@@ -158,7 +186,7 @@ The following diagram summarizes how the modules interact via composition; notab
 flowchart TB
     genint["(1) gen_int\n(DetermnisticPMF)"]
     i0["(2) I0\n(InfectionInitializationProcess)"]
-    rt["(3) rt_proc\n(RtRandomWalkProcess)"]
+    rt["(3) rt_proc\n(TransformedRandomVariable)"]
```
damonbayer, line 189: Maybe we can indicate that it is a transformation of a simple random walk?

damonbayer (changes requested):
Nice work. Some issues with the tutorials remain.

model/src/pyrenew/process/simplerandomwalk.py
damonbayer: Can we write a test based on this?

model/src/test/test_model_hospitalizations.py
dylanhmorris: I agree. I removed it in a bunch of other places in this PR and the last.

docs/source/tutorials/basic_renewal_model.qmd
dylanhmorris: good catch, needs to be a double backslash to escape it.

docs/source/tutorials/basic_renewal_model.qmd
dylanhmorris: Addressed in 5869033

docs/source/tutorials/hospital_admissions_model.qmd
dylanhmorris: Addressed in 5869033

model/src/test/test_model_hospitalizations.py
dylanhmorris: Addressed in 1081ae2

model/src/test/test_model_hospitalizations.py
dylanhmorris: Possibly but I think that's beyond the scope of this PR.

model/src/pyrenew/process/simplerandomwalk.py
dylanhmorris: > issue needed?

Created #286

model/src/test/test_model_hospitalizations.py
dylanhmorris: Created #287

model/src/pyrenew/process/simplerandomwalk.py
damonbayer: The “issue needed” was about the “implementation pending” part

dylanhmorris: Oh yes, that also needs an issue.

model/src/pyrenew/process/simplerandomwalk.py
dylanhmorris: What did you have in mind here?

model/src/pyrenew/process/simplerandomwalk.py
dylanhmorris: Created #288

model/src/pyrenew/process/simplerandomwalk.py
dylanhmorris: Something about the change from 

```python
diff = npro.sample(name + "_diffs", self.error_distribution)
```

to 

```
diff, *_ = self.step_rv.sample(**kwargs)
```
Unclear why the latter has a different shape though. This also seems relevant to the question of non-tuple RV return types.

damonbayer: Ah, I wasn't accounting for that change. There was no error in the version before this PR?

dylanhmorris: Not that I saw, anyway.

docs/source/tutorials/hospital_admissions_model.qmd
dylanhmorris: Counterpoint: hard to show the full (wide) range of prior predictions on a linear scale, b/c a random walk renewal can grow by orders of magnitude or collapse toward zero.

damonbayer: Fair enough. I do think a scatter plot would be more appropriate and look better.

dylanhmorris: Oh sorry I was looking at the wrong plot. I thought you meant the `arviz` plot of the prior predictive check. I agree about the single simulated timeseries plot, and will change it.

dylanhmorris: Addressed in 8ee433f

model/src/pyrenew/process/simplerandomwalk.py
damonbayer: Then I'm fine without a test.

docs/source/tutorials/hospital_admissions_model.qmd
damonbayer, line 1: Line 120: `Taking a pick` should be `Taking a peek`

dylanhmorris: addressed in c1f1cc9

docs/source/tutorials/basic_renewal_model.qmd
dylanhmorris: Fixed in 0afa30e

Explicitly removed the `nan` values when plotting, rather than relying on internal removal.

In the process, I caught an issue with how the posterior means were being computed. It was computing posterior means chain-by-chain rather than across all chains, but then only using the first chain's means. It's easier to see (and avoid) the potential for this error when using built-in `InferenceData.mean()` or `.median()` methods (as opposed to `np.mean` or `np.median`). I have implemented this change.

dylanhmorris:
@damonbayer ready for re-review

docs/source/tutorials/hospital_admissions_model.qmd
```diff
@@ -179,11 +180,34 @@ I0 = InfectionInitializationProcess(
...
+
+    def validate(self):
+        pass
+
+    def sample(self, n_timepoints: int, **kwargs) -> tuple:
+        sd_rt = numpyro.sample("Rt_random_walk_sd", dist.HalfNormal(0.025))
```
damonbayer, line 191: Could we do this as RandomVariable? Even better if we could pass it as an argument to `MyRt()`.

model/src/pyrenew/process/simplerandomwalk.py
```diff
@@ -12,45 +10,58 @@ class SimpleRandomWalkProcess(RandomVariable):
...
-        init: float = None,
         **kwargs,
     ) -> tuple:
         """
-        Samples from the randomwalk
+        Sample from the random walk.
 
         Parameters
         ----------
         n_timepoints : int
```
damonbayer, line 63: `n_timepoints` feels a bit specific for a generic randomwalk.

damonbayer (approved):
Thanks @dylanhmorris! A few small comments. Nothing mandatory.

model/src/pyrenew/process/simplerandomwalk.py
dylanhmorris: `n_steps`?

docs/source/tutorials/hospital_admissions_model.qmd
dylanhmorris: For a first tutorial with a user-defined RV I was thinking of keeping things a bit more concrete, but we can / should absolutely point out to the user that they can achieve additional configurability at the cost of additional abstraction when they built `RandomVariables` and models generally:

Some examples in the present context:

A not-much-more-generic case in which we always want a Half-Normal prior on the Normal random walk step s.d. but want to set a fixed scale value for that Half-Normal at class instantiation:

```python
class MyRt(RandomVariable):
   def __init__(self, rw_step_sd_prior_halfnormal_scale: float):
```

A not-much-more-generic-than-that case in which the Half-Normal prior's scale is a `RandomVariable`:

```python
class MyRt(RandomVariable):
   def __init__(self, rw_step_sd_prior_halfnormal_scale: RandomVariable):
```

A case in which we pass in a pre-instantiated prior distribution for the random walk s.d.: 

```python
class MyRt(RandomVariable):
   def __init__(self, rw_step_sd_prior: numpyro.distirbutions.Distribution):
```

A case in which we completely defer the question of how to determine the step s.d. and make it is own `RandomVariable`:

```python
class MyRt(RandomVariable):
   def __init__(self, rw_step_sd_rv: RandomVariable):
```

And finally, a case in which we completely defer the question of how to determine the step _at all_ and make _that_ its own `RandomVariable`:

```python
class MyRt(RandomVariable):
   def __init__(self, rw_step_rv: RandomVariable):
```

All of these design choices are possible. We should think about which if any we want to encourage users to make.

damonbayer: It would make https://github.com/CDCgov/multisignal-epi-inference/issues/284 easier

damonbayer: @dylanhmorris I think it is best practice to make all priors configurable at the highest level possible. When specifying a model, you should not have to rewrite a class just to change the prior it uses.

model/src/pyrenew/process/simplerandomwalk.py
dylanhmorris: Done in f23adb5

docs/source/tutorials/hospital_admissions_model.qmd
dylanhmorris: I agree RE: production code but I also think that concreteness in examples / tutorials can help make them clearer. In any case separate PR but I agree that we should probably have a tutorial demonstrating what we believe to be a readable but also modular and re-configurable way of specifying a complete model.

PR #277, opened 2024-07-18
Update negativebinomial.py

dylanhmorris:
Negative binomial obs was returning `tuple[tuple[ArrayLike]]` when we want `tuple[ArrayLike]`


damonbayer (approved):
Thanks @dylanhmorris!

Issue #274, opened 2024-07-18
In tutorials, sample chains in parallel when possible


damonbayer:
Rather than forcing chains to sample sequentially (as in https://github.com/CDCgov/multisignal-epi-inference/pull/257), we should sample in parallel when possible, so that tutorials always render as fast as the hardware permits.

PR #278, opened 2024-07-18
Parallel MCMC in tutorials 

damonbayer:
Closes https://github.com/CDCgov/multisignal-epi-inference/issues/274.
Supposedly our Github runners have 4 cores, so we should be able to perform sampling in parallel. https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories


dylanhmorris (approved):
LGTM. Maybe just add a note for the user about the `set_host_device_count` calls and how they may wish to reconfigure on their own systems.

PR #300, opened 2024-07-23
Set Default `eps` Value To `jnp.finfo(float).eps` And Test

AFg6K7h4fhy2:
Previously the defaults for the `eps` value in `dist.Poisson(rate=mu + self.eps)` and `fn=dist.NegativeBinomial2(mean=mu + self.eps, concentration=concentration)` were set to `1e-8` and `1e-10` respectively, but instead can be set to the machine float minimum (see [here](https://numpy.org/doc/stable/reference/generated/numpy.finfo.html)) via `jnp.finfo(float).eps`. This needs to be tested, however, to make sure the distribution implementations of Poisson and NegativeBinomial still function properly.

### DIFF
--- a/model/src/pyrenew/observation/poisson.py
+++ b/model/src/pyrenew/observation/poisson.py
@@ -68,9 +68,13 @@ class PoissonObservation(RandomVariable):
         tuple
         """
 
+        clipped_mu = jnp.clip(
+            mu + self.eps, min=jnp.finfo(float).eps, max=jnp.inf
+        )
+
         poisson_sample = numpyro.sample(
             name=self.name,
-            fn=dist.Poisson(rate=mu + self.eps),
+            fn=dist.Poisson(rate=clipped_mu),
             obs=obs,
         )
         return (
--- a/model/src/pyrenew/observation/poisson.py
+++ b/model/src/pyrenew/observation/poisson.py
@@ -68,9 +68,7 @@ class PoissonObservation(RandomVariable):
         tuple
         """
 
-        clipped_mu = jnp.clip(
-            mu + self.eps, min=jnp.finfo(float).eps, max=jnp.inf
-        )
+        clipped_mu = jnp.clip(mu, min=self.eps, max=jnp.inf)
 
         poisson_sample = numpyro.sample(
             name=self.name,
--- a/model/src/pyrenew/observation/poisson.py
+++ b/model/src/pyrenew/observation/poisson.py
@@ -3,6 +3,7 @@
 
 from __future__ import annotations
 
+import jax.numpy as jnp
 import numpyro
 import numpyro.distributions as dist
 from jax.typing import ArrayLike
@@ -17,7 +18,7 @@ class PoissonObservation(RandomVariable):
     def __init__(
         self,
         name: str,
-        eps: float = 1e-8,
+        eps: float = jnp.finfo(float).eps,
     ) -> None:
         """
         Default Constructor
@@ -28,7 +29,7 @@ class PoissonObservation(RandomVariable):
             Passed to numpyro.sample.
         eps : float, optional
             Small value added to the rate parameter to avoid zero values.
-            Defaults to 1e-8.
+            Defaults to `jnp.finfo(float).eps` (previously 1e-8).
 
         Returns
         -------
--- a/model/src/test/test_observation_negativebinom.py
+++ b/model/src/test/test_observation_negativebinom.py
@@ -1,7 +1,6 @@
 # -*- coding: utf-8 -*-
 # numpydoc ignore=GL08
 
-import jax.numpy as jnp
 import numpy as np
 import numpy.testing as testing
 import numpyro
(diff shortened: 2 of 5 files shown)


AFg6K7h4fhy2:
Would appreciate some thoughts on tests here. I am thinking testing new default, testing previous values, and testing a "large" value, where "testing" means "sampling does not through an error" and "samples are equal to jnp.ceil(samples)".

damonbayer:
@AFg6K7h4fhy2 Testing should demonstrate that the naive case (`eps = 0`) fails in cases where using the new default `eps` succeeds.

model/src/test/test_rtperiodicdiff.py
damonbayer, line 1: This should probably be a separate PR. I'm also confused about why this fixes the tests.

AFg6K7h4fhy2: I am also confused why this fixes the tests as well. I think something is wrong with caching or awry with memory in `test_rtperiodicdiff.py`. I changed the `rng_seed` values in the PR for #307 to those in this PR and it failed. When I changed them to new values, the test passed. This ought to be investigated further.

For the separate PR point, I can delete the modified test in `test_observation_poisson.py` in this PR if you'd like. Or was the separate PR for investigating the situation in the previous paragraph?

damonbayer: Let's get the tests passing in a separate PR ASAP. Can make a separate issue to investigate why they weren't passing before.

damonbayer: Resolved in https://github.com/CDCgov/multisignal-epi-inference/pull/327. You can remove the changes here.

damonbayer:
This problem is probably better addressed via jax.numpy.clip  https://jax.readthedocs.io/en/latest/_autosummary/jax.numpy.clip.html

AFg6K7h4fhy2:
> This problem is probably better addressed via jax.numpy.clip https://jax.readthedocs.io/en/latest/_autosummary/jax.numpy.clip.html

Not quite sure I get this. Rather than have a machine-minimum, we want to clip `esp` to be between `jnp.finfo(float).eps` and some other value (upper bound)?

damonbayer:
> Not quite sure I get this. Rather than have a machine-minimum, we want to clip `esp` to be between `jnp.finfo(float).eps` and some other value (upper bound)?

The problem we are trying to avoid is that the mean argument to `sample`  maybe be too small. `clip`ping the value to be between (whatever the smallest valid value is) and infinity makes the `mean` argument valid while keeping it as close to the input `mean` value as possible.


With this approach, you may want to error if the `mean` argument is too negative.

AFg6K7h4fhy2:
Came back to this today. Can you take a look @damonbayer ?

model/src/pyrenew/observation/poisson.py
```diff
@@ -67,9 +68,13 @@ def sample(
         tuple
         """
 
+        clipped_mu = jnp.clip(
+            mu + self.eps, min=jnp.finfo(float).eps, max=jnp.inf
```
damonbayer, line 72: We should just use clips, no need to add eps.

model/src/test/test_observation_negativebinom.py
damonbayer, line 1: There should be a test that demonstrates the value of using `clip` or `eps`. We should to show a case where the naive method fails, but our approach succeeds.

damonbayer: The `eps` stuff was added before I joined the project, so I'm not completely sure what specific situation this is meant to remedy.

AFg6K7h4fhy2: Going to tag @dylanhmorris here before I remove `eps`.

dylanhmorris: If we can't make a reprex of failure it may be that things now work and we can remove `clip`-ing / `eps`-addition entirely. In the past allowing for literal 0 negative binomial / poisson means could break things, but maybe this has now been addressed.

AFg6K7h4fhy2: I think DHM's remark — — holds here.

The following test fails because the naive case here does not fail.

Note that this is what the clipping looks like in `sample` function of NegativeBinomialObservation:

```python
clipped_mu = jnp.clip(mu + self.eps, min=self.eps, max=jnp.inf)
```

Test that does not pass currently: 

```python
def test_negativebinom_naive_vs_clipped():
    """
    Test that demonstrates the failure of the naive method (eps=0) 
    and the success of the method with clipping or small eps.
    """

  
    concentration_rv = DeterministicVariable(name="concentration", value=10)

    negb_naive = NegativeBinomialObservation(
        "negbinom_rv_naive",
        concentration_rv=concentration_rv,
        eps=0.0
    )

    negb_clipped = NegativeBinomialObservation(
        "negbinom_rv_clipped",
        concentration_rv=concentration_rv,
        eps=jnp.finfo(float).eps
    )

    naive_failed = False
    try:
            negb_naive_sample = negb_naive(mu=0.0)
    except Exception as e:
        naive_failed = True
        print(f"Naive method failed as expected: {e}")
    assert naive_failed, "Naive method did not fail as expected."

    clipped_failed = False
    try:
            negb_clipped_sample = negb_clipped(mu=0.0)
    except Exception as e:
        clipped_failed = True
        print(f"Clipped method failed unexpectedly: {e}")
    assert not clipped_failed, "Clipped method failed unexpectedly."

    assert isinstance(negb_clipped_sample, tuple)
    assert isinstance(negb_clipped_sample[0].value, ArrayLike)

    mean_sample_value = jnp.mean(negb_clipped_sample[0].value)
    assert mean_sample_value >= jnp.finfo(float).eps, "Clipped sample mean is unexpectedly low."
```

AFg6K7h4fhy2: Further thoughts here @dylanhmorris ?

damonbayer: I think I would be willing to close this PR and accept a new one where we just set `eps=0` as the default. Having the option of `eps` seems like it could be useful in the future? If we find a use case for it, we can re-visit this issue.

dylanhmorris: Sounds good to me.

AFg6K7h4fhy2: My thinking is similar if you are also thinking of keeping the clipping in place. "Doesnt hurt might be useful" is where I am at. 

>  If we find a use case for it, we can re-visit this

+1

damonbayer: For now, I prefer no clipping, and eps of 0.

PR #306, opened 2024-07-24
remove default name arg

### LINKED ISSUE #198: Remove default name arguments for NumPyro sites from non-model classes

sbidari:
closes #198


damonbayer (approved):
Thanks @sbidari!

PR #316, opened 2024-07-25
Change Certain `numpyro`s To `npro`

AFg6K7h4fhy2:
There is, in some instances, the use of `import numpyro as npro`. This has not been a uniform behavior though, despite this syntax being more preferable to MSR developers than `import numpyro`. This PR changes all instances of `import numpyro` to `import numpyro as npro` and modifies the code correspondingly.

PR #318, opened 2024-07-25
Revert "Change Certain `numpyro`s To `npro`"

damonbayer:
Reverts CDCgov/multisignal-epi-inference#316


dylanhmorris:
I have examined the diffs for the landed commit, and I don't think it actually undid any of #317. I think it simply revealed that #317 didn't catch all cases. I think we can close this but should open a new issue/PR to fix the remaining cases of `numpyro.handlers.seed`

PR #324, opened 2024-07-25
Change Order Of The `name` Argument To Match Numpyro

AFg6K7h4fhy2:
As per DHM:

> `name` should be the first argument to the constructor of all `RandomVariables`, and should be mandatory.
> 
> This makes `RandomVariable` constructors follow the same pattern as calls to `numpyro.sample()`, which have the name first. This is relevant because the `name` attribute of a `RandomVariable` is used to construct site names for any `numpyro.sample` or `numpyro.deterministic` calls that occur within the `RandomVariable`'s `sample()` method.
> 
> Currently, some `RandomVariable`s have `name` as the first constructor argument, but others do not, and some have default values for `name` (which #306 should fix)

### DIFF
--- a/model/src/pyrenew/deterministic/deterministic.py
+++ b/model/src/pyrenew/deterministic/deterministic.py
@@ -27,7 +27,7 @@ class DeterministicVariable(RandomVariable):
         name : str
             A name to assign to the process.
         vars : ArrayLike
-            A tuple with arraylike objects.
+            An ArrayLike object.
 
         Returns
         -------
--- a/model/src/pyrenew/deterministic/deterministic.py
+++ b/model/src/pyrenew/deterministic/deterministic.py
@@ -17,26 +17,26 @@ class DeterministicVariable(RandomVariable):
 
     def __init__(
         self,
-        vars: ArrayLike,
         name: str,
+        vars: ArrayLike,
     ) -> None:
         """Default constructor
 
         Parameters
         ----------
-        vars : ArrayLike
-            A tuple with arraylike objects.
         name : str
             A name to assign to the process.
+        vars : ArrayLike
+            A tuple with arraylike objects.
 
         Returns
         -------
         None
         """
 
-        self.validate(vars)
-        self.vars = jnp.atleast_1d(vars)
         self.name = name
+        self.vars = jnp.atleast_1d(vars)
+        self.validate(vars)
 
         return None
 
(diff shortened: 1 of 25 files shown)


model/src/pyrenew/deterministic/deterministic.py
```diff
@@ -17,26 +17,26 @@ class DeterministicVariable(RandomVariable):
...
         """Default constructor
 
         Parameters
         ----------
-        vars : ArrayLike
-            A tuple with arraylike objects.
         name : str
             A name to assign to the process.
+        vars : ArrayLike
+            A tuple with arraylike objects.
```
dylanhmorris, line 30: Good opportunity to fix this.
```suggestion
            An ArrayLike object.
```

model/src/pyrenew/deterministic/deterministicpmf.py
```diff
@@ -29,10 +29,10 @@ def __init__(
 
         Parameters
         ----------
-        vars : tuple
-            A tuple with arraylike objects.
         name : str
             A name to assign to the process.
+        vars : tuple
+            A tuple with arraylike objects.
```
dylanhmorris, line 35:
```suggestion
            An ArrayLike object.
```

model/src/pyrenew/deterministic/deterministicpmf.py
```diff
@@ -29,10 +29,10 @@ def __init__(
 
         Parameters
         ----------
-        vars : tuple
-            A tuple with arraylike objects.
         name : str
             A name to assign to the process.
```
dylanhmorris, line 33:
```suggestion
            A name to assign to the variable.
```

model/src/pyrenew/deterministic/deterministicpmf.py
```diff
@@ -46,7 +46,7 @@ def __init__(
             tol=tol,
         )
 
-        self.basevar = DeterministicVariable(vars, name)
+        self.basevar = DeterministicVariable(name, vars)
```
dylanhmorris, line 49:
```suggestion
        self.basevar = DeterministicVariable(name=name, vars=vars)
```

dylanhmorris (commented):
Thanks, @AFg6K7h4fhy2! A couple very minor suggestions.

dylanhmorris (approved):
LGTM, thanks @AFg6K7h4fhy2

PR #330, opened 2024-07-25
fix code links formatting in tutorials

damonbayer:
I think this is a limitation of md -> rst conversion.

Before: CleanShot 2024-07-25 at 17 00 04@2x


dylanhmorris:
Another reason to move toward using `.md` documentation source via MyST?

dylanhmorris (approved):

PR #331, opened 2024-07-25
Rename Deterministic `vars` Attribute To `value`

AFg6K7h4fhy2:
As per DHM:

> This will make it more consistent with `SampledValue`, and indeed allow for creating new `DeterministicVariable` instances by unpacking `SampledValue` instances (see https://github.com/CDCgov/multisignal-epi-inference/pull/262#discussion_r1690441825)


damonbayer (approved):
Thank @AFg6K7h4fhy2!

PR #341, opened 2024-07-29
add test for assert sample and rtype

sbidari:
This PR attempts to add tests to `_assert_sample_and_rtype` function defined as
```{python}
def _assert_sample_and_rtype(
    rp: "RandomVariable", skip_if_none: bool = True
) -> None:
```
Tests included

1. return `None` if `(rp is None) and (skip_if_none)` 
2. raises error if `(rp is None) and (not skip_if_none)` 
3. check `rp` is a valid `RandomVariable`
4. check return type is a tuple and is annotated as such 
5. raises error when return type annotation is missing in sample

### DIFF
--- a/model/src/test/test_assert_sample_and_rtype.py
+++ b/model/src/test/test_assert_sample_and_rtype.py
@@ -2,10 +2,13 @@
 Tests for _assert_sample_and_rtype method
 """
 
-import numpyro
 import pytest
 from numpy.testing import assert_equal
-from pyrenew.metaclass import RandomVariable, SampledValue, _assert_sample_and_rtype
+from pyrenew.metaclass import (
+    RandomVariable,
+    SampledValue,
+    _assert_sample_and_rtype,
+)
 
 
 class RVreturnsTuple(RandomVariable):
@@ -25,7 +28,9 @@ class RVreturnsTuple(RandomVariable):
         )
         """
 
-        return (SampledValue(value=1, t_start=self.t_start, t_unit=self.t_unit),)
+        return (
+            SampledValue(value=1, t_start=self.t_start, t_unit=self.t_unit),
+        )
 
     def validate(self):
         """
@@ -55,7 +60,9 @@ class RVnoAnnotation(RandomVariable):
         )
         """
 
-        return (SampledValue(value=1, t_start=self.t_start, t_unit=self.t_unit),)
+        return (
+            SampledValue(value=1, t_start=self.t_start, t_unit=self.t_unit),
+        )
 
     def validate(self):
         """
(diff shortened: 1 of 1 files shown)


gvegayon:
This is a helper function, and I'm unsure how much it is used across the package. Should this be somehow enforced in every model/RV that received RVs?

sbidari:
> This is a helper function, and I'm unsure how much it is used across the package. 

It seems the usage is split between using `_assert_sample_and_rtype` and `asssert isinstance(some_rv, RandomVariable)` . I think it would be good idea to replace `assert isinstance(some_rv, RandomVariable)` with `_assert_sample_and_rtype` to check for tuple return inaddition to RV.

> Should this be somehow enforced in every model/RV that received RVs?

Deferring the more general question of enforcing this in every model/RV to @damonbayer @dylanhmorris

damonbayer:
@sbidari Agree on all points.

damonbayer:
Can we keep `assert isinstance(some_rv, RandomVariable)` but check for the proper return type whenever a ` RandomVariable` is instantiated?

sbidari:
Right now, we are using `assert isinstance(some_rv, RandomVariable)` in some codes (eg: in `hospitaladmissions.py`) and `_assert_sample_and_rtype(some_rv)` in others (eg: in `InfectionsWithFeedback` ) as validation when a `RandomVariable` is instantiated. 
We should probably use one of these throughout for uniformity. `_assert_sample_and_rtype(some_rv)` tests if the `input_rv` is a `RandomVariable` and the return type

damonbayer:
@sbidari I am suggesting that we separate "checking that something is a random variable" from "checking that the random variable is valid."

sbidari:
> @sbidari I am suggesting that we separate "checking that something is a random variable" from "checking that the random variable is valid."

ok! I will create a new issue for extending the use of `_assert_sample_and_rtype` but not remove `assert isinstance(some_rv, RandomVariable)` 

Marking the tests as ready for review.

model/src/test/test_assert_sample_and_rtype.py
```diff
@@ -0,0 +1,124 @@
...
+        DistributionalRV(name="rv2", dist=dist.Normal(0, 1)),
+    ]
+    not_rv = jnp.array([1])
+
+    for rv in valid_rv:
+        _assert_sample_and_rtype(rv)
+
+    with pytest.raises(
+        Exception,
+        match=re.escape(f"{not_rv} is not an instance of RandomVariable"),
```
damonbayer, line 106:
```suggestion
        match="is not an instance of RandomVariable",
```

This will work and you can get rid of the `re` import.

model/src/test/test_assert_sample_and_rtype.py
```diff
@@ -0,0 +1,124 @@
...
+    Test that RandomVariable has a sample method with return type tuple
+    """
+
+    rv3 = RVreturnsTuple()
+    _assert_sample_and_rtype(rv3)
+
+    rv4 = RVnoAnnotation()
+    with pytest.raises(
+        Exception,
+        match=f"The RandomVariable {rv4} does not have return type annotation",
```
damonbayer, line 122:
```suggestion
        match="does not have return type annotation",
```

Very optional

damonbayer (approved):
@sbidari let me know if you want to make the changes I suggested. They are very minor and this could be merged without the changes.

sbidari:
@damonbayer I tried following the pattern in other tests and do not feel strongly about the match strings. I think it's probably better not importing additional packages unless necessary, will change to remove importing `re`.

PR #342, opened 2024-07-30
Fix some docstrings and add validation testing for `deterministic.py`

dylanhmorris:


model/src/pyrenew/deterministic/deterministic.py
```diff
@@ -63,10 +63,10 @@ def validate(value: ArrayLike) -> None:
         Raises
         ------
         Exception
-            If the input value object is not a ArrayLike.
+            If the input value object is not an ArrayLike object.
         """
         if not isinstance(value, ArrayLike):
-            raise Exception("value is not a ArrayLike")
+            raise ValueError("value is not an ArrayLike object")
```
damonbayer, line 69: Can we print the object and its type?

damonbayer (approved):
One optional improvement.

model/src/pyrenew/deterministic/deterministic.py
dylanhmorris: addressed in 73711ce

damonbayer (approved):

model/src/pyrenew/deterministic/deterministic.py
```diff
@@ -63,10 +63,14 @@ def validate(value: ArrayLike) -> None:
...
-            If the input value object is not a ArrayLike.
+            If the input value object is not an ArrayLike object.
         """
         if not isinstance(value, ArrayLike):
-            raise Exception("value is not a ArrayLike")
+            raise ValueError(
+                f"value {value} passed to a DeterministicVariable "
+                f"is of type {type(value).__name__}, expected "
+                "an ArrayLike object"
+            )
```
damonbayer, line 73: We have this type of check error throughout the project. Could be good to write a function for it.

PR #344, opened 2024-07-30
Day of the week tutorial

gvegayon:
## Implementation

- The day-of-the-week [DOW] effect is implemented as a `RandomVariable` that returns a vector of length 7.
- The `RandomVariable` `latent.HospitalAdmissions` takes care of broadcasting the seven values to match the length of the data (latent infections, in this case).
- The PR replaces how `latent.HospitalAdmissions.sample()` is called; instead of passing a `latent_infections` array, it passes the corresponding `SampledValue` containing the latent infections.
- Since `SampledValues` has `t_start`, we can correct for the offset of the infections wrt the observed data.
- In addition, the constructor of `latent.HospitalAdmissions` receives a new argument indicating what's the first day of the week in the observed data. With the previous point, `latent.HospitalAdmissions.sample()` properly broadcasts the DOW effect.

The tutorial itself recycles the model built during the "Implementing a Hospital Admissions-only Model". 

## Draft Notes (not implemented, but here for future reference)

Current state:

- The day-of-the-week [DOW] effect is applied via `latent.HospitalAdmissions`.
- The sampling method of hospital admissions receives latent infections.
- Those latent infections are a time series equal to the number of observed data points **plus** the infection seeding process.
- Since the seeding process extends the observed time series to the left, the resulting broadcasting should account for that.

```mermaid
flowchart LR
   Seeding[Seeding\nprocess] -->|Input to| LatentI[Latent\nInfections]
   LatentI -->|Input to| HospitalAdmissions
   subgraph HospitalAdmissions["HospitalAdmissions.sample()"]
      dow["day_of_week_effect_rv(\n  n_timepoints=latent_hospital_admissions.size\n)"]
   end
```

To implement:

- We would need to pass on the information about `t_start` resulting from the seeding process.
- Currently, the seeding process has `t_start=self.t_start`, both `None` (see [here](https://github.com/CDCgov/multisignal-epi-inference/blob/cf38ef0d5d904d8f30e12111eaf3403714f6c429/model/src/pyrenew/latent/infections.py#L100-L106)). 
- The latent infections vector should have `t_start=-I0.size`.
- Instead of passing the raw array to `latent.HospitalAdmissions`, we should pass the `SampledValue` object with the `t_start=-I0.size`. 
- With that accessible, instead of passing a `n_timepoints` to the DOW RV, we could pass the `SampledValue` latent infections.
- The DOW RV sample method could then access the `t_start` and the number of time points to sample and adjust accordingly.
- The DOW RV object should know what day of the week the first observation of the sample is; this information could be passed during the construction of the DOW RV.

```mermaid
flowchart LR
   Seeding[Seeding\nprocess] -->|Input to| LatentI[Latent\nInfections]
   LatentI -->|Input to| HospitalAdmissions
   subgraph HospitalAdmissions["HospitalAdmissions.sample()"]
      dow["day_of_week_effect_rv(\n  data=latent_hospital_admissions\n)"]
   end
```


docs/source/tutorials/day_of_the_week.qmd
```diff
@@ -0,0 +1,309 @@
...
+title: Implementing a Day of the Week Effect
+format: gfm
+engine: jupyter
+---
+
+This document illustrates how to leverage the time-aware arrays to create a a day of the week effect. We use the same model designed in the hospital admissions-only tutorial.
+
+## Recap: Hospital Admissions Model
+
+In the "Fitting a hospital admissions-only model" tutorial, we built a fairly complicated model that included pre-defined random variables as well as a custom random variable representing the reproductive number. As this is not the focus of this tutorial, we will not go into detail about the model's components. Instead, we will focus on the day-of-the-week effect. Yet, we still need to re-define the model's components to build the day-of-the-week effect.
```
damonbayer, line 11: link to the tutorial

damonbayer, line 11: > In the "Fitting a hospital admissions-only model" tutorial

Link to the tutorial

docs/source/tutorials/day_of_the_week.qmd
```diff
@@ -0,0 +1,309 @@
...
+    ),
+)
+```
+
+
+## Round 2: Incorporating day-of-the-week effects
+
+We will re-use the infection to admission interval and infection to hospitalization rate from the previous model. But we will also add a day-of-the-week effect. To do this, we will add two additional arguments to the latent hospital admissions random variable: `day_of_the_week_rv` (a `RandomVariable`) and `obs_data_first_day_of_the_week` (an `int` mapping days of the week from 0:6, zero being Monday). The `day_of_the_week_rv`'s sample method should return a vector of length seven; those values are then broadcasted to match the length of the dataset. Moreover, since the observed data may start in a weekday other than Monday, the `obs_data_first_day_of_the_week` argument is used to offset the day-of-the-week effect.
+
+For this example, the effect will be passed as a truncated normal distribution with a mean of 1.0 and a standard deviation of 0.5. The distribution will be truncated between 0.1 and 10.0[^note-other-examples]:
```
damonbayer, line 216: This is a bad idea for identifiability reasons. I think we should use something based on a simplex instead.

docs/source/tutorials/day_of_the_week.qmd
```diff
@@ -0,0 +1,309 @@
...
+
+Now, by re-defining the latent hospital admissions random variable with the day-of-the-week effect, we can build a model that includes this effect:
+
+```{python}
+# | label: latent-hosp-weekday
+latent_hosp_wday_effect = latent.HospitalAdmissions(
+    infection_to_admission_interval_rv=inf_hosp_int,
+    infect_hosp_rate_rv=hosp_rate,
+    day_of_week_effect_rv=dayofweek_effect,
+    # As the dataset starts on October 30th, 2023 (Monday)
```
damonbayer, line 239: Would be nice to write this in a generic way that uses the date library to get the day of the week from the start date and compute the correct integer for the offset.

docs/source/tutorials/day_of_the_week.qmd
```diff
@@ -0,0 +1,309 @@
...
+# | fig-cap: Hospital Admissions posterior distribution without weekday effect
+import jax
+import numpy as np
+
+# Model without weekday effect
+hosp_model.run(
+    num_samples=2000,
+    num_warmup=2000,
+    data_observed_hosp_admissions=daily_hosp_admits,
+    rng_key=jax.random.PRNGKey(54),
```
damonbayer, line 195:
```suggestion
    rng_key=jax.random.key(54),
```
See https://github.com/CDCgov/multisignal-epi-inference/issues/370.

docs/source/tutorials/day_of_the_week.qmd
```diff
@@ -0,0 +1,309 @@
...
+
+```{python}
+# | label: fig-output-day-of-week
+# | fig-cap: Day of the week effect
+out = hosp_model_dow.plot_posterior(
+    var="dayofweek_effect", ylab="Day of the Week Effect", samples=500
+)
+```
+
+The new model with the day-of-the-week effect can be compared to the previous model without the effect. Let's start by reproducing the figure without the day-of-the-week effect, and then plotting the new model with the effect:
```
damonbayer, line 279: > Let's start by reproducing the figure without the day-of-the-week effect, and then plotting the new model with the effect:

Seems to suggest we will be doing more after plotting the figures, but that's all we do.

model/src/pyrenew/latent/hospitaladmissions.py
```diff
@@ -22,13 +21,18 @@ class HospitalAdmissionsSample(NamedTuple):
         The infection-to-hospitalization rate. Defaults to None.
     latent_hospital_admissions : SampledValue or None
         The computed number of hospital admissions. Defaults to None.
+    day_of_week_effect_mult : SampledValue or None
```
damonbayer, line 24: I think it makes more sense to just call this `multiplier`. It's not really specific to the day of the week effect.

model/src/pyrenew/latent/hospitaladmissions.py
```diff
@@ -171,7 +196,9 @@ def sample(
 
         infection_hosp_rate, *_ = self.infect_hosp_rate_rv(**kwargs)
 
-        infection_hosp_rate_t = infection_hosp_rate.value * latent_infections
+        infection_hosp_rate_t = (
```
damonbayer, line 199: Can we pick better names here? https://github.com/CDCgov/multisignal-epi-inference/issues/357

model/src/pyrenew/latent/hospitaladmissions.py
```diff
@@ -184,13 +211,33 @@ def sample(
...
+        # Identifying the offset
+        inf_offset = (
+            latent_infections.t_start % 7
+            if latent_infections.t_start is not None
+            else 0
+        ) + self.obs_data_first_day_of_the_week
+
+        # Replicating the day of the week effect to match the number of
+        # timepoints
+        dow_effect = jnp.tile(
```
damonbayer, line 235: Why aren't we using `tile_unitl_n`?

model/src/pyrenew/latent/hospitaladmissions.py
```diff
@@ -184,13 +211,33 @@ def sample(
...
+                "Day of the week effect should have 7 values. "
+                f"Got {dow_effect_sampled.value.size} instead."
+            )
+
+        # Identifying the offset
+        inf_offset = (
+            latent_infections.t_start % 7
+            if latent_infections.t_start is not None
+            else 0
+        ) + self.obs_data_first_day_of_the_week
```
damonbayer, line 231: Prefer a standard `if-else` ordering.

damonbayer, line 231: Not sure how to think of this… is this saying that we assume `t=0` to be a Monday?

damonbayer (changes requested):
Thanks @gvegayon! Some required changes and some things open for discussion.

model/src/pyrenew/latent/hospitaladmissions.py
```diff
@@ -67,6 +71,7 @@ def __init__(
         infect_hosp_rate_rv: RandomVariable,
```
damonbayer, line 71: These names could be made more clear.

- `infect_hosp_rate_rv` makes me think of something like a case-fataility-rate, but I think is actually the delay distribution?
- `hosp_report_prob_rv` is not more like an infection-hospitalization-ratio than a reporting probability, I think?

model/src/pyrenew/latent/hospitaladmissions.py
gvegayon: Not assuming. It is part of the design. See the documentation.

model/src/pyrenew/latent/hospitaladmissions.py
gvegayon: This implementation was prior to that function. Will fix.

model/src/pyrenew/latent/hospitaladmissions.py
gvegayon: This is standard in python, I would argue. What do you think, @dylanhmorris?

model/src/pyrenew/latent/hospitaladmissions.py
gvegayon: Sure

model/src/pyrenew/latent/hospitaladmissions.py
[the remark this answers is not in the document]
gvegayon: Not sure I'm following.

model/src/pyrenew/latent/hospitaladmissions.py
gvegayon: And drop the `_rv` suffix?

model/src/pyrenew/latent/hospitaladmissions.py
gvegayon: Even better, got rid of that.

docs/source/tutorials/day_of_the_week.qmd
gvegayon: Sure!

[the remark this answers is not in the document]
gvegayon: Corrected.

docs/source/tutorials/day_of_the_week.qmd
gvegayon: Sounds good!

docs/source/tutorials/day_of_the_week.qmd
```diff
@@ -0,0 +1,309 @@
...
+Running the model:
+
+```{python}
+# | label: model-2-run-weekday
+# Model with weekday effect
+hosp_model_dow.run(
+    num_samples=2000,
+    num_warmup=2000,
+    data_observed_hosp_admissions=daily_hosp_admits,
+    rng_key=jax.random.PRNGKey(54),
```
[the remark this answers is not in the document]
gvegayon: Sounds good

docs/source/tutorials/day_of_the_week.qmd
gvegayon: Fixed

docs/source/tutorials/day_of_the_week.qmd
gvegayon: done

docs/source/tutorials/day_of_the_week.qmd
gvegayon: Good point

model/src/pyrenew/latent/hospitaladmissions.py
gvegayon: Also, worth noticing `datetime` and `numpy.datetime64` both use Monday as anchor.

model/src/pyrenew/latent/hospitaladmissions.py
damonbayer: No, sorry. Keep the suffix.

model/src/pyrenew/latent/hospitaladmissions.py
damonbayer: Can you explain (either in GitHub or a code comment) the logic behind these lines? I find it hard to follow.

model/src/pyrenew/latent/hospitaladmissions.py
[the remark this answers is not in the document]
damonbayer: In this model, we assume that a length 7 vector represents a week. So the "fundamental time unit" is one day. Do we do enough to communicate or enforce this? Is that out of scope for this?

model/src/pyrenew/latent/hospitaladmissions.py
dylanhmorris: Yes, this is the correct syntax for tuple comprehension with a conditional https://stackoverflow.com/questions/4260280/if-else-in-a-list-comprehension

damonbayer: But this is not a tuple comprehension? It's just a matter of preference.

model/src/pyrenew/latent/hospitaladmissions.py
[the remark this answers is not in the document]
gvegayon: I think is out of scope.

model/src/pyrenew/latent/hospitaladmissions.py
gvegayon: It defines the offset in terms of days between 0-7, regardless of the starting date of the data. If the seeding process starts -4 days, `-4 % 7 = 3` that would be Sunday->Saturday->Friday->Thursday. Otherwise, if it is 4 then `4 % 7 = 4`, thus Friday. This would work with bigger numbers too. For instance, adding three weeks, we would have `-25 % 7 = 3`.

gvegayon: It defines the offset in terms of days between 0-6, regardless of the starting date of the data. If the seeding process starts -4 days, -4 % 7 = 3 that would be Sunday->Saturday->Friday->Thursday. Otherwise, if it is 4 then 4 % 7 = 4, thus Friday. This would work with bigger numbers too. For instance, adding three weeks, we would have -25 % 7 = 3.

docs/source/tutorials/day_of_the_week.qmd
```diff
@@ -0,0 +1,314 @@
...
+    ),
+)
+```
+
+
+## Round 2: Incorporating day-of-the-week effects
+
+We will re-use the infection to admission interval and infection to hospitalization rate from the previous model. But we will also add a day-of-the-week effect. To do this, we will add two additional arguments to the latent hospital admissions random variable: `day_of_the_week_rv` (a `RandomVariable`) and `obs_data_first_day_of_the_week` (an `int` mapping days of the week from 0:6, zero being Monday). The `day_of_the_week_rv`'s sample method should return a vector of length seven; those values are then broadcasted to match the length of the dataset. Moreover, since the observed data may start in a weekday other than Monday, the `obs_data_first_day_of_the_week` argument is used to offset the day-of-the-week effect.
+
+For this example, the effect will be passed as a Dirichlet distribution with a concentration parameter of 1/7: [^note-other-examples]:
```
damonbayer, line 216: Why concentration of 1/7? We should probably multiply the Dirichlet by 7, as in the WW model. You can do this with a `TransformedRandomVariable`:

```python
TransformedRandomVariable(
    "hosp_wday_effect",
    DistributionalRV(
        "hosp_wday_effect_raw", dist.Dirichlet(concentration=jnp.ones(7))
    ),
    transforms.AffineTransform(loc=0, scale=7),
)
```

gvegayon:
Ready for review, @damonbayer and @dylanhmorris.

damonbayer (approved):
Looks good. I think we just want to wait for https://github.com/CDCgov/multisignal-epi-inference/pull/387 so we can avoid having to define `MyDOWEffect`.

We could also merge as is and address that later. Your call @gvegayon.

gvegayon:
> Looks good. I think we just want to wait for #387 so we can avoid having to define `MyDOWEffect`.
> 
> We could also merge as is and address that later. Your call @gvegayon.

I'll wait

damonbayer:
> I'll wait

@gvegayon https://github.com/CDCgov/multisignal-epi-inference/pull/387 is merged

gvegayon:
Just addressed conflicts. Once tests pass, I'll merge.

damonbayer:
@gvegayon The point of waiting for https://github.com/CDCgov/PyRenew/pull/387 was so that your `MyDOWEffect` class could be replaced with a `TransformedRandomVariable`. Please make a new issue for this.

gvegayon:
> @gvegayon The point of waiting for #387 was so that your `MyDOWEffect` class could be replaced with a `TransformedRandomVariable`. Please make a new issue for this. I think it could/should be done in this sprint.

Yes! Sorry I forgot that. Fixing now

PR #346, opened 2024-07-30
add dependency check workflow

sbidari:
adds dependency check to CI

### DIFF
--- a/.github/workflows/deptry.yml
+++ b/.github/workflows/deptry.yml
@@ -1,0 +1,32 @@
+name: model
+
+on:
+  pull_request:
+  push:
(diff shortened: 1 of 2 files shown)


.github/workflows/deptry.yml
```diff
@@ -0,0 +1,32 @@
+name: model
```
dylanhmorris, line 1:
```suggestion
name: deptry-check
```

dylanhmorris (changes requested):
EDIT: misread.

dylanhmorris (approved):
Misread things above. LGTM.

PR #348, opened 2024-07-31
add a narrow dependency group for testing in CI

### LINKED ISSUE #337: Define a narrow `test` set of poetry dependencies and run CI testing in that context
This will catch dependency problems that can be obscured when running in a general `dev` context, see e.g. #336

sbidari:
closes #337

### DIFF
--- a/.github/workflows/deptry.yml
+++ b/.github/workflows/deptry.yml
@@ -25,7 +25,7 @@ jobs:
         run: pip install poetry
 
       - name: install package
-        run: poetry install --with dev,test
+        run: poetry install --with dev
 
       - name: run deptry
         run: |
--- a/.github/workflows/deptry.yml
+++ b/.github/workflows/deptry.yml
@@ -25,7 +25,7 @@ jobs:
         run: pip install poetry
 
       - name: install package
-        run: poetry install --with dev
+        run: poetry install --with docs
 
       - name: run deptry
         run: |
--- a/.github/workflows/deptry.yml
+++ b/.github/workflows/deptry.yml
@@ -25,7 +25,7 @@ jobs:
         run: pip install poetry
 
       - name: install package
-        run: poetry install --with docs,test
+        run: poetry install --with dev
 
       - name: run deptry
         run: |
--- a/.github/workflows/deptry.yml
+++ b/.github/workflows/deptry.yml
@@ -25,7 +25,7 @@ jobs:
         run: pip install poetry
 
       - name: install package
-        run: poetry install --with docs, test
+        run: poetry install --with docs,test
 
       - name: run deptry
         run: |
--- a/.github/workflows/deptry.yml
+++ b/.github/workflows/deptry.yml
@@ -1,4 +1,4 @@
-name: model
+name: dependency-check
 
 on:
   pull_request:
@@ -25,7 +25,7 @@ jobs:
         run: pip install poetry
 
       - name: install package
-        run: poetry install --with docs
+        run: poetry install --with docs, test
 
       - name: run deptry
         run: |
--- a/.github/workflows/deptry.yml
+++ b/.github/workflows/deptry.yml
@@ -29,4 +29,4 @@ jobs:
 
       - name: run deptry
         run: |
-          poetry run deptry . --per-rule-ignores "DEP001=pyrenew,DEP003=pytest"
+          poetry run deptry . --per-rule-ignores "DEP001=pyrenew,DEP001=pytest,DEP003=pytest"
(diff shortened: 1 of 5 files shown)


damonbayer (changes requested):
I think there should still be a `dev` group that includes things like `deptry`, `ipykernel`, and `ipywidgets` (maybe a few others, but I'm not sure). Those packages are not required for building the docs, but are used in development.

sbidari:
> I think there should still be a `dev` group that includes things like `deptry`, `ipykernel`, and `ipywidgets` (maybe a few others, but I'm not sure). Those packages are not required for building the docs, but are used in development.

`ipykernel` is also required for building the docs, not sure about `ipywidgets` 

I can separate what is currently `docs` into `dev` and `docs` as below. We would also need to install `dev` dependency for the website build but maybe we only need `dev` for development

```
[tool.poetry.group.dev.dependencies]
ipykernel = "^6.29.3"
nbconvert = "^7.16.4"
ipywidgets = "^8.1.3"
deptry = "^0.17.0"

[tool.poetry.group.docs.dependencies]
pillow = "^10.3.0"
pyyaml = "^6.0.0"
nbclient = "^0.10.0"
nbformat = "^5.10.0"
numpydoc = "^1.7.0"
arviz = "^0.18.0"
sphinx = "^7.2.6"
sphinxcontrib-mermaid = "^0.9.2"
sphinx-autodoc-typehints = "^2.1.0"
sphinx-book-theme = "^1.1.2"
```

damonbayer:
>  We would also need to install `dev` dependency for the website build but maybe we only need `dev` for development

Can we have
- `main` lets you use the package
- `main` + `test` lets you run tests
- `main` + `docs` lets you build the website
- everything else goes in `dev`

sbidari:
I am afraid `dev` might end up having just `deptry` but will run some tests to confirm

damonbayer:
> I am afraid `dev` might end up having just `deptry` but will run some tests to confirm 

That's totally fine with me. I just want each use case to have the minimum required dependencies.

dylanhmorris:
Agree with @damonbayer

.github/workflows/deptry.yml
```diff
@@ -25,7 +25,7 @@ jobs:
         run: pip install poetry
 
       - name: install package
-        run: poetry install --with dev
+        run: poetry install --with dev,test
```
damonbayer, line 28:
```suggestion
        run: poetry install --with dev
```

pyproject.toml
```diff
@@ -23,22 +23,32 @@ matplotlib = "^3.8.3"
...
+
+[tool.poetry.group.docs]
+optional = true
+
+[tool.poetry.group.docs.dependencies]
 ipykernel = "^6.29.3"
+pillow = "^10.3.0"
+pyyaml = "^6.0.0"
 nbclient = "^0.10.0"
 nbformat = "^5.10.0"
```
damonbayer, line 38: What are these used for? @CDCgov/multisignal-epi-inference-devs

damonbayer (approved):
Thanks @sbidari! I would like to discuss https://github.com/CDCgov/multisignal-epi-inference/pull/348#discussion_r1699131139 before merging.

pyproject.toml
dylanhmorris: - Is `pillow` perhaps used somewhere in handling the output images bundled with the quarto-rendered `.rst` tutorials? @gvegayon. Might be worth trying a fresh render of the docs without it and seeing if anything breaks.
 
- I would guess `pyyaml` had something to do with loading of `yaml` formatted configs from, e.g. https://github.com/cdcgov/wastewater-informed-covid-forecasting/ But as above I think we can try removing it and seeing if it breaks things.

damonbayer: I am also asking about `nbclient ` and `nbformat`.

gvegayon: Don't remember exactly what was the pillow for, but `nbclient` and `nbformat` are required for rendering the quarto using the jupyter engine.

dylanhmorris: Related: would be very nice to handle the dependency on the `quarto` CLI via `poetry` if possible. I don't think we currently do.

Note that this [official python interface to the `quarto` CLI](https://github.com/quarto-dev/quarto-python) lists `nbclient`, `nbformat`, and `pyyaml` as deps: 

https://github.com/quarto-dev/quarto-python/blob/4fcf29d33f8da520cdfd690d5dd53d556a21c09d/setup.py#L16

damonbayer: @sbidari Based on this conversation, can you try removing `pillow`? Seems like we need to keep the others.

gvegayon: There's a dependency for pillow somewhere, but I'm happy if that works :)!

sbidari: All the checks have passed even after removing pillow!

PR #351, opened 2024-08-01
add helper function for type error assertion

### LINKED ISSUE #345: DRYify testing of validation errors
> We have this type of check/error message throughout the project. Could be good to write a function for it.

_Originally posted by @damonbayer in https://github.com/CDCgov/multisignal-epi-inference/pull/342#discussion_r1697286580_

sbidari:
closes #345

### DIFF
--- a/model/src/pyrenew/metaclass.py
+++ b/model/src/pyrenew/metaclass.py
@@ -21,7 +21,7 @@ from pyrenew.mcmcutils import plot_posterior, spread_draws
 from pyrenew.transformation import Transform
 
 
-def _assert_type_error(arg_name: str, value, expected_type) -> None:
+def _assert_type(arg_name: str, value, expected_type) -> None:
     """
     Matches TypeError arising during validation
 
@@ -46,14 +46,11 @@ def _assert_type_error(arg_name: str, value, expected_type) -> None:
 
     if not isinstance(value, expected_type):
         raise TypeError(
-            f"{arg_name} must be an instance of {expected_type}. "
-            f"Got {type(value)}"
+            f"{arg_name} must be an instance of {expected_type}. " f"Got {type(value)}"
         )
 
 
-def _assert_sample_and_rtype(
-    rp: "RandomVariable", skip_if_none: bool = True
-) -> None:
+def _assert_sample_and_rtype(rp: "RandomVariable", skip_if_none: bool = True) -> None:
     """
     Return type-checking for RandomVariable's sample function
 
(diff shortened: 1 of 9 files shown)


model/src/pyrenew/metaclass.py
```diff
@@ -21,6 +21,36 @@
 from pyrenew.transformation import Transform
 
 
+def _assert_type_error(arg_name: str, value, expected_type) -> None:
```
dylanhmorris, line 24: I'd call this just `_assert_type` rather than `_assert_type_error` since it passes without error if the type is as expected 
 
```suggestion
def _assert_type(arg_name: str, value, expected_type) -> None:
```

dylanhmorris (changes requested):

damonbayer (changes requested):
Looks nice. We need tests for this function. We should use it on every function for input validation. Maybe not appropriate to do that all in a single PR.

model/src/pyrenew/metaclass.py
sbidari: changed in https://github.com/CDCgov/multisignal-epi-inference/pull/351/commits/7ad67f8b02315ffee008fa6d8bd44aae7867862a

sbidari:
> Looks nice. We need tests for this function. We should use it on every function for input validation. Maybe not appropriate to do that all in a single PR.

Opened new issues for this!

damonbayer:
@sbidari I should have been more clear. We should add the tests in this PR, but we can roll out input validation more gradually.

model/src/test/test_assert_type.py
```diff
@@ -0,0 +1,41 @@
...
+
+    values = [None] * 4
+    arg_names = ["input_int", "input_string", "input_tuple", "input_rv"]
+    input_types = [int, str, tuple, RandomVariable]
+
+    for arg, value, input in zip(arg_names, values, input_types):
+        with pytest.raises(
+            TypeError,
+            match=f"{arg} must be an instance of {input}. "
+            f"Got {type(value)}",
```
damonbayer, line 39: It's probably enough just to check if it raises a `TypeError`. Duplicating the entire error message seems unnecessary.

sbidari: fixed in https://github.com/CDCgov/multisignal-epi-inference/pull/351/commits/24c15941e49819b13cef04e48ae5341346a1bff3

dylanhmorris: I am weakly in favor of checking the error message but I'll defer to @damonbayer

damonbayer (approved):

Issue #354, opened 2024-08-01
Use `_assert_type` for input validation


sbidari:
https://github.com/CDCgov/multisignal-epi-inference/pull/351#pullrequestreview-2213566751

dylanhmorris:
Closing as not planned as we're moving to different patterns

PR #356, opened 2024-08-01
add validation using assert sample and rtype

sbidari:
This PR aims to adds validation of `RandomVariables` using `_assert_sample_and_rtype` function. The validation methods currently implemented varies between checking something is a random variable using `isinstance` and checking that the random variable is valid (i.e check `tuple` return type). 

Based on the comments here https://github.com/CDCgov/multisignal-epi-inference/pull/341#issuecomment-2259357623, I think @damonbayer would like to keep the current implementation of `assert isinstance` and maybe add further check with `_assert_sample_and_rtype`? I wanted to confirm my understanding before adding more to this PR

### DIFF
--- a/model/src/pyrenew/process/simplerandomwalk.py
+++ b/model/src/pyrenew/process/simplerandomwalk.py
@@ -3,7 +3,11 @@
 
 import jax.numpy as jnp
 from numpyro.contrib.control_flow import scan
-from pyrenew.metaclass import RandomVariable, SampledValue
+from pyrenew.metaclass import (
+    RandomVariable,
+    SampledValue,
+    _assert_sample_and_rtype,
+)
 
 
 class SimpleRandomWalkProcess(RandomVariable):
@@ -45,6 +49,7 @@ class SimpleRandomWalkProcess(RandomVariable):
         None
         """
         self.name = name
+        self.validate(step_rv=step_rv, init_rv=init_rv)
         self.step_rv = step_rv
         self.init_rv = init_rv
         self.t_start = t_start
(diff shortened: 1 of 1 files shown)


damonbayer:
I would like it so that any time we have some function `f(x: RandomVariable)`, we validate that `x` is a `RandomVariable` using `_assert_type`.

Separately, any time we instantiate a `RandomVariable`, we check that it is valid via something like `_assert_sample_and_rtype`.


So if you have code like this 

```python
my_non_rv = y
f(x=my_non_rv)
```
`f` will error because `x` is not a `RandomVariable`.

But if you have code like this
```python
my_rv = some_way_to_make_an_rv_with_the_wrong_return_type()
f(x=my_rv)
```
`some_way_to_make_an_rv_with_the_wrong_return_type` will error because it has the wrong return type and we won't even try to call `f`.

cc @dylanhmorris for thoughts.

sbidari:
> I would like it so that any time we have some function `f(x: RandomVariable)`, we validate that `x` is a `RandomVariable` using `_assert_type`.

This can be done in #354 

> Separately, any time we instantiate a `RandomVariable`, we check that it is valid via something like `_assert_sample_and_rtype`.

I don't think `_assert_sample_and_rtype` is useful for this. But happy to hear if you have ideas for it.

I think this PR and associated issue can be closed? @damonbayer

sbidari:
Note for clarification: shelving this for now after discussion with Damon. Potential design changes may affect this in the future.

damonbayer:
> Note for clarification: shelving this for now after discussion with Damon. Potential design changes may affect this in the future.

Closing for now.

PR #365, opened 2024-08-06
Fix placeholder `rst` hook bugs

### LINKED ISSUE #364: Update `.rst` placeholder generation
This will no longer cause the hook to run, because the tutorials are now in subdirectories of `docs/`, not of `model/docs`:

https://github.com/CDCgov/multisignal-epi-inference/blob/c2d01eafc04b599f3c7a934f3ad8029783f5235f/.pre-commit-config.yaml#L21

The hook also needs updating:

https://github.com/CDCgov/multisignal-epi-inference/blob/c2d01eafc04b599f3c7a934f3ad8029783f5235f/hook_scripts/pre-commit-rst-placeholder.sh#L19

It might also be nice to update the hook so that it fails if the`.rst` files exist but are not identical to the expected placeholders.

dylanhmorris:
Closes #364 but does not resolve this, which will require a larger rethink of the hook script:

> It might also be nice to update the hook so that it fails if the.rst files exist but are not identical to the expected placeholders.


hook_scripts/pre-commit-rst-placeholder.sh
```diff
@@ -16,10 +16,9 @@ for rst in $RST_FILES; do
     bname=$(basename $rst .rst)
 
     # Check if the corresponding qmd file exists
-    if [ ! -f model/docs/${bname}.qmd ]; then
+    if [ ! -f ${TUTORIALS_DIR}/${bname}.qmd ]; then
         echo "Removing $rst"
         rm $rst
-        COUNTER=$((COUNTER+1))
```
gvegayon, line 22: Why wouldn't you count this as a change for failing? So, if I understand correctly, if the file doesn't exist (b/c the tutorial was removed), then this should be changed, and pre-commit should fail, no?

docs/source/tutorials/.gitignore
```diff
@@ -8,4 +8,4 @@ _compiled
 
 *.ipynb
 *.py
-*.rst
```
gvegayon, line 11: This used to be a safeguard to avoid adding rst files resulting from rendering the qmd file.

gvegayon (commented):
Mostly one comment about counting changes when RSTs are removed. Before closing/merging this, I would also create a separate issue for improving this whole process (compare expected vs observed RST).

docs/source/tutorials/.gitignore
dylanhmorris: Right. But it needs to be changed b/c the pre-commit generated `.rst` (which we _do_ want to git track) now live in this directory.

hook_scripts/pre-commit-rst-placeholder.sh
dylanhmorris: Good catch; that should not have been removed. Will revert.

dylanhmorris: Reverted in 9127df3

gvegayon (approved):
LGTM

Issue #366, opened 2024-08-06
Improve placeholder tutorial file pre-commit hook


dylanhmorris:
From #364:
> It might also be nice to update the hook so that it fails if the `.rst` files exist but are not identical to the expected placeholders.

Was not done in #365, as it requires a larger refactor and might be worth waiting for other breaking changes such as #231

dylanhmorris:
I think we can close this as won't do because we should instead just do #459

PR #367, opened 2024-08-06
refactor website generation

dylanhmorris:
## Summary of changes

- `docs/source/tutorials/Makefile` now does not have `poetry` calls internally. Instead, to use within the poetry environment, the user should call `poetry run make <target>`
-  `docs/source/tutorials/Makefile` can be used to render `.rst` files. Note that users will have to be careful not to overwrite the placeholders with locally rendered `.rst` files. This should be addressed by #366, but I believe it is a worthwhile price to pay for making other workflows more robust.
- The `website` action now ruyns `poetry run make tutorials_clean` and `poetry run make tutorials_rst` to render and postprocess the `.rst` files. This simplifies `workflow.yaml`
- The website action renders tutorial `.rst` files from their original `docs/source/tutorials` directory rather than first moving them to `model/`. This should make pathing less prone to surprises. See https://github.com/CDCgov/multisignal-epi-inference/pull/367/files#r1705908307


.github/workflows/website.yaml
```diff
@@ -42,24 +36,16 @@ jobs:
       # which is then uploaded as an artifact.
       - name: Render documents
         run: |
-            poetry run quarto render README.md --to rst -o index2.rst
-            mv docs/source/tutorials model/
```
dylanhmorris, line 46: This line made it harder to do relative pathing within tutorials

damonbayer (approved):

PR #368, opened 2024-08-06
Fix typos and clarify sample site naming guidance in `developer_documentation.rst`

dylanhmorris:


docs/source/developer_documentation.rst
```diff
@@ -26,20 +26,22 @@ A variety of coding conventions are enforced by automated tools in continuous in
...
-   -  Use the ``observed_`` for the output of sample statements where ``obs`` is a ``data_`` prefixed object.
-   -  Thus, code which may reasonably written like ``infections = infections.sample(x, obs=infections)`` should instead be written ``observed_infections = infections_rv.sample(x, obs=data_infections)``.
 
--  Class conventions
+Class conventions
+-----------------
+-  Composing ``Model`` objects is discouraged. To modularize and reuse code across ``Model``s, create custom ``RandomVariable`` classes.
+-  Returning anything from ``Model.sample`` is discouraged. Instead, sample from models using ``Predictive`` or our ``prior_predictive`` or ``posterior_predictive`` functions.
+-  Using ``numpyro.deterministic`` within a ``RandomVariable.sample`` method is discouraged. Only use it in ``Model.model`` calls. If something might need to be recorded from a ``RandomVariable``, it should be returned from the ``RandomVariable.sample`` call so it can be recorded within the ``Model.model`` call.
+-  Using default site names in a ``RandomVariable.sample`` is discouraged. Only use default site names at the ``Model.model`` level.
```
damonbayer, line 42: I think there is a better term than “using.” This is meant to suggest that someone writing a random variable should not give their own random variable a default site name. We should say something closer to that and point out that the site name should be an argument to the RV constructor.

dylanhmorris: Tried something in 33063a7. Might be too verbose. Let me know what you think, @damonbayer

docs/source/developer_documentation.rst
```diff
@@ -26,20 +27,31 @@ A variety of coding conventions are enforced by automated tools in continuous in
 PyRenew Principles
 ------------------
 
--  Variable naming conventions
+Variable naming conventions
+---------------------------
+-  Use the ``data_`` prefix for (potentially) observed data.
+-  Use the ``_rv`` suffix for random variables.
+-  Use the ``observed_`` for the output of sample statements where ``obs`` is a ``data_`` prefixed object.
```
gvegayon, line 34: I am not entirely sure that these prefixes make complete sense. Particularly, I make an explicit distinction between observed and unobserved data. In principle, everything can be called data, so it is superfluous to use that as a prefix. Instead, I would say `sampled_` for *any* output (deterministic or random), leave the `observed_` to actually observed data, and keep the `_rv` suffix.

docs/source/developer_documentation.rst
```diff
@@ -26,20 +27,31 @@ A variety of coding conventions are enforced by automated tools in continuous in
...
+-  Thus, code which may reasonably written like ``infections = infections.sample(x, obs=infections)`` should instead be written ``observed_infections = infections_rv.sample(x, obs=data_infections)``.
+
+
+Class conventions
+-----------------
+-  Composing ``Model`` objects is discouraged. To modularize and reuse code across ``Model``s, create custom ``RandomVariable`` classes.
+-  Returning anything from ``Model.sample`` is discouraged. Instead, sample from models using ``Predictive`` or our ``prior_predictive`` or ``posterior_predictive`` functions.
```
gvegayon, line 41: But `sample` is what `__call__` calls, so it needs to be defined. I think this is not up to date?

dylanhmorris: `Model.sample`, not `RandomVariable.sample`. @damonbayer originally wrote this, so I don't want to speak for him, but as I understand it, the idea was that `Model.sample` (alias `Model.model`) should always be _defined_ and should make various `RandomVariable.sample()` calls internally (thus defining the actual model), but should itself be a void function (i.e. should return `None`). I believe the idea was to discourage the user from treating models as `RandomVariables` (see previous point about composition) and to encourage the user to use `Model.prior_predictive()` rather than `Model.sample()` for forward simulation.

gvegayon: Totally right. Missed the Model bit 🙃

docs/source/developer_documentation.rst
damonbayer: @gvegayon @dylanhmorris How about this? The `observed_` prefix is nice because it maps to the `obs` argument in `numpyro.sample`, though I understand why it is confusing say the `observed_` object may not actually be observed.
```
-  Use the ``observed_`` prefix for (potentially) observed data.
-  Use the ``_rv`` suffix for random variables.
-  Use the ``sampled_`` for the output of sample statements where ``obs`` is a ``obserced_`` prefixed object.
-  Thus, code which may reasonably written like ``infections = infections.sample(x, obs=infections)`` should instead be written ``sampled_infections = infections_rv.sample(x, obs=observed_infections)``.
```

dylanhmorris:
Closing pending a rethink of sample site naming etc in #407

PR #372, opened 2024-08-13
create and use default rt across tests

### LINKED ISSUE #287: Re-use simple Rt definition throughout testing suite
see e.g.:

> Should we share this across multiple files?

_Originally posted by @damonbayer in https://github.com/CDCgov/multisignal-epi-inference/pull/275#discussion_r1685019213_

sbidari:
closes #287

### DIFF
--- a/model/src/test/utils.py
+++ b/model/src/test/utils.py
@@ -10,10 +10,16 @@ from pyrenew.metaclass import DistributionalRV, TransformedRandomVariable
 from pyrenew.process import SimpleRandomWalkProcess
 
 
-def get_default_rt():
+def simple_rt(arg_name: str = "Rt_rv"):
     """
     Helper function to create a default Rt
     RandomVariable for testing.
+
+    Parameters
+    -----------
+    arg_name : str
+        Name assigned to the randonvariable.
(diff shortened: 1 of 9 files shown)


model/src/test/utils.py
```diff
@@ -0,0 +1,36 @@
...
+test utilities
+"""
+
+import numpyro.distributions as dist
+import pyrenew.transformation as t
+from pyrenew.metaclass import DistributionalRV, TransformedRandomVariable
+from pyrenew.process import SimpleRandomWalkProcess
+
+
+def get_default_rt():
```
damonbayer, line 13: suggest renaming. Something like "simple_rt()". Also might be useful to have arguments for names (with the same defaults as here). That way you could use the function twice in the same model if you needed to.

damonbayer (changes requested):
This will be really nice. Two suggestions.

model/src/test/utils.py
sbidari: changed in 875e28392516bc6ba06fb2de210368a4dfad68fe

damonbayer (approved):
Thanks @sbidari!

PR #373, opened 2024-08-13
replace jax.random.PRNGKey with jax.random.key

### LINKED ISSUE #370: Replace all uses of `jax.random.PRNGKey` with `jax.random.key`
From `jax.random.PRNGKey` [documentation](https://jax.readthedocs.io/en/latest/_autosummary/jax.random.PRNGKey.html):

> "This function produces old-style legacy PRNG keys, which are arrays of dtype uint32. For more, see the note in the [PRNG keys](https://jax.readthedocs.io/en/latest/jax.random.html#prng-keys) section. When possible, [jax.random.key()](https://jax.readthedocs.io/en/latest/_autosummary/jax.random.key.html#jax.random.key) is recommended for use instead.

sbidari:
closes #370

### DIFF
--- a/docs/source/tutorials/basic_renewal_model.qmd
+++ b/docs/source/tutorials/basic_renewal_model.qmd
@@ -247,7 +247,7 @@ model1.run(
     num_warmup=2000,
     num_samples=1000,
     data_observed_infections=sim_data.observed_infections.value,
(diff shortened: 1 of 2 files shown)


damonbayer (approved):
Thanks @sbidari!

PR #375, opened 2024-08-13
374 make predictor_values an argument in GLMPrediction

### LINKED ISSUE #374: Make predictor values an argument passed to `GLMPrediction.predict()` and `GLMPrediction.sample()`
Change `fixed_predictor_values` from `GLMPrediction` class attribute to a `predictor_values` argument. This will allow it to interact more appropriately with numpyro tools such as `Predictive`.

@AFg6K7h4fhy2, @sbidari is going to do this, so when it's ready you can use the fix in your cfaepim branch.

sbidari:
closes #374


model/src/pyrenew/regression.py
```diff
@@ -114,21 +107,31 @@ def predict(
...
             Array of transformed predictions.
         """
-        transformed_prediction = (
-            intercept + self.fixed_predictor_values @ coefficients
-        )
+        transformed_prediction = intercept + predictor_values @ coefficients
         return self.transform.inv(transformed_prediction)
 
-    def sample(self) -> dict:
+    def sample(self, **kwargs) -> dict:
```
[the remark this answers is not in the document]
dylanhmorris: I'd explicitly declare `predictor_values` and make it a mandatory positional argument

model/src/pyrenew/regression.py
```diff
@@ -114,21 +107,31 @@ def predict(
...
         """
         Sample generalized linear model
 
+        Parameters
+        -----------
+        **kwargs : dict
```
dylanhmorris, line 132:
```suggestion
        predictor_values : ArrayLike(n_predictors, n_observations)
            Matrix of predictor variables (covariates) for the
            regression problem. Each row should represent the
            predictor values corresponding to an observation;
            each column should represent a predictor variable.
            Do not include values of 1 for the intercept;
            these will be added automatically. Passed as the 
            `predictor_values` argument to 
            :meth:`GLMPrediction.predict()`

        **kwargs : dict
             Additional keyword arguments. Ignored.
```

dylanhmorris (changes requested):
This looks very good, thanks @sbidari. Just one small mandatory change.

Also, if you have the time, it could be in scope to fix the class docstring (which incorrectly refers to "`GLMObservation`" and "`observations`" (legacy of an older version) and/or to change the return type of `sample` from `dict` to `SampledValue`, to be consistent with other `RandomVariable`s. Either/both of these could also be separate PRs.

model/src/pyrenew/regression.py
```diff
@@ -141,7 +155,9 @@ def sample(self) -> dict:
         coefficients = numpyro.sample(
             self.name + self.coefficient_suffix, self.coefficient_priors
         )
-        prediction = self.predict(intercept, coefficients)
+
+        predictor_values = kwargs.get("predictor_values")
```
dylanhmorris, line 159:
```suggestion
```

dylanhmorris (changes requested):

sbidari:
@dylanhmorris ready for re-review!
The recent commit includes mandatory changes plus docstring fix and change to the `sample` return object from a `dict` to a collection of `SampledValues`

model/src/pyrenew/regression.py
```diff
@@ -114,38 +138,54 @@ def predict(
...
             Array of transformed predictions.
         """
-        transformed_prediction = (
-            intercept + self.fixed_predictor_values @ coefficients
-        )
+        transformed_prediction = intercept + predictor_values @ coefficients
         return self.transform.inv(transformed_prediction)
 
-    def sample(self) -> dict:
+    def sample(self, predictor_values: ArrayLike) -> dict:
```
dylanhmorris, line 157:
```suggestion
    def sample(self, predictor_values: ArrayLike) -> GLMPredictionSample:
```

model/src/pyrenew/regression.py
```diff
@@ -114,38 +138,54 @@ def predict(
...
-            coefficients=coefficients,
+        prediction = self.predict(intercept, coefficients, predictor_values)
+
+        return GLMPredictionSample(
+            prediction=SampledValue(prediction),
+            intercept=SampledValue(intercept),
+            coefficients=SampledValue(coefficients),
         )
 
     def __call__(self, **kwargs):
```
dylanhmorris, line 191:
```suggestion
    def __call__(self, *args, **kwargs):
```

dylanhmorris: And similarly for the call to `sample()` within

dylanhmorris (changes requested):
Looks very good. A few very minor tweaks. Thanks, @sbidari!

model/src/pyrenew/regression.py
sbidari: changed in cf3b96a7f790b891ec1df0141542bcb9863dd365

model/src/pyrenew/regression.py
sbidari: included in cf3b96a7f790b891ec1df0141542bcb9863dd365

PR #378, opened 2024-08-14
Explicitly specify domain for AffineTransform in ScaledLogitTransform

dylanhmorris:
This will prevent bugs if it is used to create `TransformedDistribution`s See https://github.com/pyro-ppl/numpyro/issues/1756


damonbayer (approved):

gvegayon:
Not sure how the constraints work exactly. I've been trying to set it to be within 0-1, but I still get weird values. I suggest creating an integration test where you call the random variable via model sampling.

dylanhmorris:
Was this comment supposed to be in the simplex issue? I think for the inference to work you'd need to constrain the domain of the transformation to a simplex, not just constrain the components of the input vector to the unit interval

PR #380, opened 2024-08-15
Refactor differenced processes, random walks, and AR processes

### LINKED ISSUE #328: Refactor AR process
We should clean up https://github.com/CDCgov/multisignal-epi-inference/blob/main/model/src/pyrenew/process/ar.py to reflect recent design decisions. This work should be similar to the work done in https://github.com/CDCgov/multisignal-epi-inference/pull/275.

We should also look around at other examples to see if we can learn any lessons:

- https://forum.pyro.ai/t/ar-p-in-numpyro/2671
- https://juanitorduz.github.io/arma_numpyro/
- https://num.pyro.ai/en/stable/examples/ar2.html

dylanhmorris:
Closes #328


model/src/pyrenew/process/randomwalk.py
```diff
@@ -101,3 +99,39 @@ def validate():
         """
         super().validate()
         return None
+
+
+class StandardNormal(RandomWalk):
+    """
+    A random walk with
+    standard Normal (mean = 0, standard deviation = 1)
+    steps, implmemented via the base RandomWalk class.
```
damonbayer, line 108:
```suggestion
    steps, implemented via the base RandomWalk class.
```

model/src/pyrenew/process/randomwalk.py
```diff
@@ -101,3 +99,39 @@ def validate():
         """
         super().validate()
         return None
+
+
+class StandardNormal(RandomWalk):
```
damonbayer, line 104: Suggest including RandomWalk in the name.

dylanhmorris: Yes, this was a mistake. Fixed in 8a4f6c6

model/src/pyrenew/process/randomwalk.py
dylanhmorris: Fixed in f39fa7280

model/src/pyrenew/deterministic/deterministic.py
damonbayer, line 1: Any concerns with this based on @gvegayon's comment here? https://github.com/CDCgov/multisignal-epi-inference/issues/386

dylanhmorris: No, relevant areas have all been changed to use `atleast_1d` coercion at the last possible moment, e.g. in the `InfectionInitializationMethod`s. https://github.com/CDCgov/multisignal-epi-inference/issues/386#issuecomment-2296789195

model/src/test/test_rtperiodicdiff.py
damonbayer, line 1: I don't think `RtPeriodicDiffProcess` is tested anywhere now?

model/src/pyrenew/process/iidrandomsequence.py
```diff
@@ -0,0 +1,145 @@
+# -*- coding: utf-8 -*-
+# numpydoc ignore=GL08
+
+import numpyro.distributions as dist
+from numpyro.contrib.control_flow import scan
+from pyrenew.metaclass import DistributionalRV, RandomVariable, SampledValue
+
+
+class IIDRamdomSequence(RandomVariable):
```
damonbayer, line 9:
```suggestion
class IIDRandomSequence(RandomVariable):
```
Present throughout, but only commenting here.

dylanhmorris: Oops! Will fix.

docs/source/tutorials/extending_pyrenew.qmd
```diff
@@ -44,7 +48,9 @@ The following code-chunk defines the model components. Notice that for both the
 gen_int_array = jnp.array([0.25, 0.5, 0.15, 0.1])
 
 gen_int = DeterministicPMF(name="gen_int", value=gen_int_array)
-feedback_strength = DeterministicVariable(name="feedback_strength", value=0.01)
+feedback_strength = DeterministicVariable(
+    name="feedback_strength", value=jnp.array([0.01])
```
damonbayer, line 52: Will `value=0.01` not work?

model/src/pyrenew/model/rtinfectionsrenewalmodel.py
```diff
@@ -214,34 +213,22 @@ def sample(
         )
 
         observed_infections, *_ = self.infection_obs_process_rv(
-            mu=post_initialization_latent_infections.value[padding:],
+            mu=post_initialization_latent_infections.value,
```
damonbayer, line 216: Why? is this fixing some bug?

model/src/pyrenew/model/rtinfectionsrenewalmodel.py
```diff
@@ -214,34 +213,22 @@ def sample(
...
-
-        if observed_infections is not None:
-            observed_infections = au.pad_x_to_match_y(
-                observed_infections.value,
-                all_latent_infections,
-                jnp.nan,
-                pad_direction="start",
-            )
-
-        Rt = au.pad_x_to_match_y(
```
damonbayer, line 235: This removal will make some figures have incorrect x-axes.

model/src/test/utils.py
```diff
@@ -6,37 +6,64 @@
...
+           Name assigned to the RandomVariable.
+           If None, then defaults to "Rt_rv"
+
+        Returns
+        -------
+        None
+        """
+        self.name = name
+        self.rt_rv_ = TransformedRandomVariable(
+            name=name + "log_rt_random_walk",
```
damonbayer, line 40:
```suggestion
            name=f"{name}_log_rt_random_walk",
```

model/src/test/utils.py
```diff
@@ -6,37 +6,64 @@
...
+                    name="rw_step_rv", distribution=dist.Normal(0, 0.025)
+                ),
             ),
-        ),
-        transforms=t.ExpTransform(),
-    )
+            transforms=t.ExpTransform(),
+        )
+        self.rt_init_rv_ = DistributionalRV(
+            name=name + "init_log_rt", distribution=dist.Normal(0, 0.2)
```
damonbayer, line 50:
```suggestion
            name=f"{name}_init_log_rt", distribution=dist.Normal(0, 0.2)
```

model/src/pyrenew/process/iidrandomsequence.py
dylanhmorris: Fixed in 0e08ee9

model/src/test/test_rtperiodicdiff.py
dylanhmorris: This was incorrectly deleted. Will add back.

model/src/test/utils.py
damonbayer: Actually, this seems like an inappropriate name for the transformed random variable. I think we should drop the `log` part.

model/src/test/test_rtperiodicdiff.py
dylanhmorris: Fixed in 2f304cb

docs/source/tutorials/extending_pyrenew.qmd
dylanhmorris: Good catch. It will as of 07a482e

model/src/test/test_iid_random_sequence.py
damonbayer, line 1: Test for `StandardNormalSequence`?

damonbayer, line 1: Test that the vectorized and non-vectorized approaches produce the same result?

model/src/pyrenew/model/rtinfectionsrenewalmodel.py
dylanhmorris: Incorrectly removed. Fixed in 0166ba0

model/src/test/test_iid_random_sequence.py
dylanhmorris: Oh good catch the vectorized result is implementation pending. I should either implement it now or cause it to raise an error.

model/src/pyrenew/process/differencedprocess.py
```diff
@@ -0,0 +1,228 @@
...
+from jax.typing import ArrayLike
+from numpyro.contrib.control_flow import scan
+from pyrenew.metaclass import RandomVariable, SampledValue
+
+
+class DifferencedProcess(RandomVariable):
+    """
+    Class for differenced stochastic process X(t),
+    constructed by placing a fundamental stochastic
+    process on the first differences (rates of change).
```
damonbayer, line 16: Left over from previous description?

model/src/pyrenew/process/differencedprocess.py
```diff
@@ -0,0 +1,228 @@
...
+        if not isinstance(differencing_order, int):
+            raise ValueError(
+                "differcing_order must be an integer."
+                f"got {type(differencing_order)}"
+            )
+        if not differencing_order >= 1:
+            raise ValueError(
+                "differcing_order must be an integer "
+                "greater than or equal to 1. Got "
+                f"{differencing_order}"
```
damonbayer, line 227: Can combine these into a single `if` and with a common error message.

model/src/pyrenew/process/differencedprocess.py
```diff
@@ -0,0 +1,228 @@
...
+    def integrate(
+        self, init_diff_vals: ArrayLike, highest_order_diff_vals: ArrayLike
+    ):
+        """
+        Integrate (de-difference) the differenced process,
+        obtaining the process values X(t=0), X(t=1), ... X(t)
+        from the n-th differences and a set of initial process /
+        difference values X(t=0), X^1(t=1), X^2(t=2), ...
+        X^(n-1)(t=n-1), where X^k(t) is the value of the n-th
+        differenc at index t of the process.
```
damonbayer, line 78:
```suggestion
        difference at index t of the process.
```

damonbayer: Better to use math formatting here.

damonbayer: Some precise definition of the nth difference could be helpful.

model/src/test/test_differenced_process.py
```diff
@@ -0,0 +1,122 @@
...
+        ValueError, match="exactly as many initial difference values"
+    ):
+        proc.integrate(inits_long, diffs)
+    proc.integrate(inits_correct, diffs)
+
+
+@pytest.mark.parametrize(
+    ["order", "n_diffs"], [[1, 250], [2, 40], [3, 10], [4, 10], [5, 5]]
+)
+def test_integrator_correctness(order, n_diffs):
```
damonbayer, line 106: A non-random example computed by hand would be illustrative.

model/src/pyrenew/process/rtperiodicdiff.py
damonbayer, line 1: Should there be some reference in the classes to the fact that this is a first difference process?

dylanhmorris: Yes I think so. Can add.

model/src/test/test_model_basic_renewal.py
```diff
@@ -125,6 +135,7 @@ def test_model_basicrenewal_no_obs_model():
     # Sampling and fitting model 0 (with no obs for infections)
     with numpyro.handlers.seed(rng_seed=223):
         model0_samp = model0.sample(n_datapoints=30)
+    print(model0_samp)
```
damonbayer, line 138: Remove?
```suggestion
```

model/src/test/test_model_basic_renewal.py
```diff
@@ -33,16 +33,21 @@ def test_model_basicrenewal_no_timepoints_or_observations():
         name="gen_int", value=jnp.array([0.25, 0.25, 0.25, 0.25])
     )
 
-    I0 = DistributionalRV(name="I0", distribution=dist.LogNormal(0, 1))
+    I0_init_rv = InfectionInitializationProcess(
+        "I0_initialization",
+        DistributionalRV(name="I0", distribution=dist.LogNormal(0, 1)),
+        InitializeInfectionsZeroPad(n_timepoints=gen_int.basevar.value.size),
```
damonbayer, line 39: Why this change throughout?

dylanhmorris: I'm not sure why/how it was working before (if it was) but it shouldn't have, since `gen_int` is a `DeterministicVariable` and therefore doesn't have a defined `.size`.

model/src/pyrenew/process/ar.py
```diff
@@ -4,96 +4,84 @@
...
+        self.noise_rv_ = StandardNormalSequence(name=name + noise_rv_suffix)
 
     def sample(
         self,
-        duration: int,
-        inits: ArrayLike = None,
+        n: int,
+        autoreg: ArrayLike,
+        init_vals: ArrayLike,
+        noise_sd: ArrayLike,
```
damonbayer, line 42: Wouldn't this always be scalar?

model/src/test/test_ar_process.py
```diff
@@ -11,33 +11,44 @@ def test_ar_can_be_sampled():
...
     with numpyro.handlers.seed(rng_seed=62):
-        # can sample with and without inits
-        ar1(duration=3532, inits=jnp.array([50.0]))
-        ar1(duration=5023)
+        # can sample
+        ar1(
+            n=3532,
+            init_vals=jnp.array([50.0]),
+            autoreg=jnp.array([0.95]),
+            noise_sd=jnp.array([0.5]),
```
damonbayer, line 21: I think this could just be  `noise_sd=0.5,` throughout?

model/src/test/test_model_basic_renewal.py
dylanhmorris: Can be reverted. Didn't realize `DeterministicPMF` variables had a `size()` method. Whether they should is a different question, but it's beyond the scope of this PR.

model/src/pyrenew/process/differencedprocess.py
dylanhmorris: Will change.

model/src/test/test_model_basic_renewal.py
```diff
@@ -222,38 +238,35 @@ def test_model_basicrenewal_padding() -> None:  # numpydoc ignore=GL08
...
     with numpyro.handlers.seed(rng_seed=223):
-        model1_samp = model1.sample(n_datapoints=30, padding=pad_size)
+        model1_samp = model1.sample(n_datapoints=30)
 
     model1.run(
         num_warmup=500,
         num_samples=500,
         rng_key=jr.key(22),
         data_observed_infections=model1_samp.observed_infections.value,
-        padding=5,
```
damonbayer, line 256: Why is this removed?

dylanhmorris: restored in fe18feb

model/src/test/test_model_basic_renewal.py
dylanhmorris, line 39:
```suggestion
        InitializeInfectionsZeroPad(n_timepoints=gen_int.size()),
```

model/src/test/test_model_basic_renewal.py
```diff
@@ -64,16 +69,21 @@ def test_model_basicrenewal_both_timepoints_and_observations():
         value=jnp.array([0.25, 0.25, 0.25, 0.25]),
     )
 
-    I0 = DistributionalRV(name="I0", distribution=dist.LogNormal(0, 1))
+    I0_init_rv = InfectionInitializationProcess(
+        "I0_initialization",
+        DistributionalRV(name="I0", distribution=dist.LogNormal(0, 1)),
+        InitializeInfectionsZeroPad(n_timepoints=gen_int.basevar.value.size),
```
dylanhmorris, line 75:
```suggestion
        InitializeInfectionsZeroPad(n_timepoints=gen_int.size()),
```

model/src/test/test_model_basic_renewal.py
```diff
@@ -100,22 +110,22 @@ def test_model_basicrenewal_no_obs_model():
...
     with pytest.raises(ValueError):
-        I0 = DistributionalRV(name="I0", distribution=1)
+        _ = DistributionalRV(name="I0", distribution=1)
 
-    I0 = InfectionInitializationProcess(
+    I0_init_rv = InfectionInitializationProcess(
         "I0_initialization",
         DistributionalRV(name="I0", distribution=dist.LogNormal(0, 1)),
-        InitializeInfectionsZeroPad(n_timepoints=gen_int.size()),
+        InitializeInfectionsZeroPad(n_timepoints=gen_int.basevar.value.size),
```
dylanhmorris, line 118:
```suggestion
        InitializeInfectionsZeroPad(n_timepoints=gen_int.size()),
```

model/src/test/test_model_basic_renewal.py
```diff
@@ -173,21 +184,21 @@ def test_model_basicrenewal_with_obs_model():
         name="gen_int", value=jnp.array([0.25, 0.25, 0.25, 0.25])
     )
 
-    I0 = InfectionInitializationProcess(
+    I0_init_rv = InfectionInitializationProcess(
         "I0_initialization",
         DistributionalRV(name="I0", distribution=dist.LogNormal(0, 1)),
-        InitializeInfectionsZeroPad(n_timepoints=gen_int.size()),
+        InitializeInfectionsZeroPad(n_timepoints=gen_int.basevar.value.size),
```
dylanhmorris, line 190:
```suggestion
        InitializeInfectionsZeroPad(n_timepoints=gen_int.size()),
```

model/src/test/test_model_basic_renewal.py
```diff
@@ -222,29 +238,29 @@ def test_model_basicrenewal_padding() -> None:  # numpydoc ignore=GL08
...
```
dylanhmorris, line 244:
```suggestion
        InitializeInfectionsZeroPad(n_timepoints=gen_int.size()),
```

dylanhmorris (commented):
Flagging `gen_int.basevar.value.size` references for reversion

model/src/test/test_model_basic_renewal.py
dylanhmorris: Fixed in 1b34490

model/src/test/test_ar_process.py
dylanhmorris: Fixed in e21cce6

damonbayer (changes requested):
My initial review is done. I need to spend some more time with differencedprocess.py before approval.

model/src/pyrenew/process/rtperiodicdiff.py
dylanhmorris: addressed in 59cfbfe

gvegayon:
If needed, I can review after the conflicts have been solved.

model/src/pyrenew/process/differencedprocess.py
dylanhmorris: addressed in 5bceb27

model/src/pyrenew/process/differencedprocess.py
dylanhmorris: addressed in 5bceb27173

model/src/test/test_iid_random_sequence.py
dylanhmorris: added the `expand()` method to distributional rvs in  d1f5c5e

Test for the use in vectorized IID sequences pending.

dylanhmorris: Addressed as of 76539ef

dylanhmorris: Addressed in 8789f3b

Also changed from using `expand` to using `expand_by`, as that will produce results more consistent with `scan` if the original distribution is already vector-valued.

model/src/pyrenew/process/differencedprocess.py
dylanhmorris: Don't want to compare to `1` unless it's an `int`, to avoid an uninformative error:

```python
if isinstance("test", int) or "test" >= 1:
    print("this should not print")
```

```python
TypeError: '>=' not supported between instances of 'str' and 'int'
```

model/src/pyrenew/process/ar.py
dylanhmorris: In the future we might want to enable a vector-valued `noise_sd` argument, but as the process is currently implemented, that wouldn't make sense. Changing the type hint accordingly.

dylanhmorris: addressed in 3790b0c

model/src/pyrenew/process/differencedprocess.py
damonbayer: Just do `if not (isinstance(differencing_order, int) and differencing_order > 1):`

model/src/test/test_differenced_process.py
dylanhmorris: addressed in 0bfc0d3

src/pyrenew/metaclass.py
```diff
@@ -305,6 +305,11 @@ def __init__(
             If not None, reparameterize sampling
             from the distribution according to the
             given numpyro reparameterizer
+        expand_by_shape : tuple, optional
+            If not None, call :meth:`expand_by()` on the
```
damonbayer, line 309:
```suggestion
            If not None, call :math:`expand_by()` on the
```

dylanhmorris: No, this is `:meth:` for "method"
https://stackoverflow.com/questions/34848253/whats-the-difference-in-behaviour-between-func-and-meth-roles-in-python-sph

dylanhmorris:
Ready for re-review @damonbayer

damonbayer:
@dylanhmorris looks like your latest commit caused a test failure

dylanhmorris:
> @dylanhmorris looks like your latest commit caused a test failure

Should be fixed now. Issue was that `rtperiodicdiffar.py` wants to be able to sample differenced processes for durations shorter than the differencing order. I go back and forth on whether we should allow this, but if we're going to, it should just return the appropriate initial values.

src/pyrenew/process/differencedprocess.py
```diff
@@ -104,9 +104,21 @@ def integrate(
...
         order of the process.
         """
-        init_diff_vals = jnp.atleast_1d(init_diff_vals)
-        highest_order_diff_vals = jnp.atleast_1d(highest_order_diff_vals)
-        n_inits = init_diff_vals.shape[0]
+        init_arr = jnp.atleast_1d(init_diff_vals)
+        diff_arr = jnp.atleast_1d(highest_order_diff_vals)
+        if not init_arr.ndim == 1:
+            raise ValueError(
+                "init_diff_vals must be 1-dimensional "
```
damonbayer, line 111: same as other comment.

src/pyrenew/process/differencedprocess.py
```diff
@@ -104,9 +104,21 @@ def integrate(
...
+                "array or coercible to 1D array. "
+                f"Got {init_diff_vals}"
+            )
+        if not diff_arr.ndim == 1:
+            raise ValueError(
+                "highest_order_diff_vals must be a "
+                "1-dimensional array or coercible to "
```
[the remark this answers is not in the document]
dylanhmorris: Fixed by ccacc9d

src/pyrenew/process/differencedprocess.py
dylanhmorris: Fixed by ccacc9d

src/pyrenew/process/differencedprocess.py
```diff
@@ -198,15 +193,23 @@ def sample(
             Whose value entry is a single array representing the
             undifferenced timeseries
         """
-        diffs, *_ = self.fundamental_process.sample(
-            *args,
-            n=(n - self.differencing_order),
-            init_vals=fundamental_process_init_vals,
-            **kwargs,
-        )
+        if n < 1:
```
damonbayer, line 196: also check for integer?

src/pyrenew/process/differencedprocess.py
```diff
@@ -198,15 +193,23 @@ def sample(
...
+                init_vals=fundamental_process_init_vals,
+                **kwargs,
+            )
+            diffs = diff_samp.value
+        else:
+            diffs = jnp.array([])
         return (
             SampledValue(
-                value=self.integrate(init_vals, diffs.value),
+                value=self.integrate(init_vals, diffs)[:n],
```
damonbayer, line 212: slight preference to compute self.integrate and store it to make the return statement more readable.

dylanhmorris: addressed in 755fc39

src/pyrenew/process/differencedprocess.py
dylanhmorris: addressed in 755fc39

My only question is whether we'd like to allow `n=jnp.array(5)`, but I think we can make this less strict if and when that becomes a probelm.

src/pyrenew/process/differencedprocess.py
```diff
@@ -0,0 +1,256 @@
...
+from pyrenew.metaclass import RandomVariable, SampledValue
+
+
+class DifferencedProcess(RandomVariable):
+    """
+    Class for differenced stochastic process X(t),
+    constructed by placing a fundamental stochastic
+    process on the :math:`n^{th}` differences
+    (rates of change). See
+    https://otexts.com/fpp2/stationarity.html
```
damonbayer, line 19:
```suggestion
    https://otexts.com/fpp3/stationarity.html
```

newer version

damonbayer:
```suggestion
    https://otexts.com/fpp3/stationarity.html#differencing
```

or to be more specific

dylanhmorris: I think it makes more sense to start from the top, since the actual chapter title is "Stationarity and differencing"

src/pyrenew/process/differencedprocess.py
```diff
@@ -0,0 +1,256 @@
...
+        Notes
+        -----
+        The order of differencing is the discrete
+        analogue of the order of a derivative in single
+        variable calculus. A first difference (derivative)
+        represents a rate of change. A second difference
+        (derivative) represents the rate of change of that
+        rate of change, et cetera.
+        """
+        self.assert_valid_differencing_order(differencing_order)
```
damonbayer, line 69: I think we have many checks for "integer greater than 1" throughout the codebase. Maybe a common function would be helpful.

pyproject.toml
```diff
@@ -41,6 +41,7 @@ sphinxcontrib-mermaid = "^0.9.2"
 sphinx-autodoc-typehints = "^2.1.0"
 sphinx-book-theme = "^1.1.2"
 ipywidgets = "^8.1.3"
+jupyter = "^1.0.0"
```
damonbayer, line 44: Why is this added now?

dylanhmorris: Can't render tutorials without it, at least on my machine. Possible consequence of #388?

src/pyrenew/process/differencedprocess.py
dylanhmorris: Agree. Separate issue I think.

pyproject.toml
damonbayer: Tutorials are still being rendered in CI without this change, though?

src/pyrenew/process/differencedprocess.py
```diff
@@ -0,0 +1,256 @@
...
+        """
+        Default constructor
+
+        Parameters
+        ----------
+        name : str
+            Name of the stochastic process
+        fundamental_process : RandomVariable
+            Stochastic process for the
+           differences. Should accept an
```
damonbayer, line 40:
```suggestion
            differences. Should accept an
```

pyproject.toml
dylanhmorris: Perhaps installed by the `quarto` installation action? Not sure. But in any case I would argue it should be an explicit docs dep since quarto uses it as the execution engine for the tutorials.

src/pyrenew/process/iidrandomsequence.py
```diff
@@ -0,0 +1,150 @@
...
+    """
+    Class for constructing random sequence of
+    independent and identically distributed elements
+    given an arbitrary RandomVariable representing
+    those elements.
+    """
+
+    def __init__(
+        self,
+        name: str,
```
damonbayer, line 20: I don't think this `name` actually gets used? Wouldn't you need a `numpyro.deterministic` call in `sample` to do that?

dylanhmorris: True, I was including it on the grounds that we might in the future move toward all RVs being mandatorily named / optionally recorded. But happy to remove for now if you want.

damonbayer: I think I prefer removing, but am somewhat interested in an optional `name` and/or optional `record`.

dylanhmorris: Will remove.

damonbayer: related: https://github.com/CDCgov/PyRenew/issues/402

src/test/test_iid_random_sequence.py
```diff
@@ -0,0 +1,81 @@
...
+def test_iidrandomsequence_with_dist_rv(distribution, n):
+    """
+    Check that an IIDRandomSequence can be
+    initialized and sampled from when the element_rv is
+    a distributional RV, including with array-valued
+    distributions
+    """
+    element_rv = DistributionalRV("el_rv", distribution=distribution)
+    rseq = IIDRandomSequence(name="randseq", element_rv=element_rv)
+    expected_shape = distribution.batch_shape
```
damonbayer, line 34: since `expected_shape` is always changed in the `if-else` block below, I think a different name or just re-using `distribution.batch_shape` below is more clear.

src/pyrenew/process/iidrandomsequence.py
dylanhmorris: Removed in 7cc5377

src/test/test_iid_random_sequence.py
dylanhmorris: fixed in e84f975

src/test/test_random_walk.py
```diff
@@ -21,25 +22,16 @@ def test_rw_can_be_sampled():
...
+        DistributionalRV(name="rw_step_rv", distribution=dist.Normal(0, 0.27)),
     )
 
     with numpyro.handlers.seed(rng_seed=62):
         # can sample with a fixed init
         # and with a random init
-        ans_rand = rw_init_rand(n_steps=3532)
-        ans_fixed = rw_init_fixed(n_steps=5023)
-
+        ans_rand = rw(n=3532, init_vals=init_rv_rand()[0].value)
```
damonbayer, line 33: I think the aspects of this test which rely on a random init are unnecessary. Isn't it really a test of whether `init_rv_rand()[0].value` is the size?

src/test/test_random_walk.py
```diff
@@ -49,51 +41,55 @@ def test_rw_can_be_sampled():
...
+@pytest.mark.parametrize(
+    ["step_mean", "step_sd"],
+    [
+        [0, 1],
+        [2.253, 0.025],
+        [-3.2521, 1],
+        [1052, 3],
+        [1e-6, 0.02],
+    ],
+)
```
damonbayer, line 53: This is fine. Just wondering what kinds of failure modes might be identified by trying these different combinations?

dylanhmorris: - Failures of integer coercion to float (which should work)
- Failures to work with negative step means (which should work)
-  Very large and very small values don't cause problems (they shouldn't)

src/test/test_latent_admissions.py
```diff
@@ -91,58 +75,3 @@ def test_admissions_sample():
...
-        with testing.assert_raises(ValueError):
-            hosp3b(latent_infections=inf_sampled2).multiplier.value
-
-    testing.assert_array_equal(
-        sim_hosp_2a[2 : (sim_hosp_2b.size - 2)],
-        sim_hosp_2b[: (sim_hosp_2b.size - 4)],
-    )
-
-
-test_admissions_sample()
```
damonbayer, line 148: this line should be removed

src/test/test_random_walk.py
damonbayer: I don't really see how we could introduce any of those issues, but I suppose its fine to test. These are just straightforwardly passed to `dist.Normal`, so none of our code attempts to check / process these arguments.

src/test/test_latent_admissions.py
```diff
@@ -91,58 +75,3 @@ def test_admissions_sample():
         sim_hosp_1.latent_hospital_admissions.value,
         inf_sampled1[0].value,
     )
-
-    # Testing the offset in the observed data
```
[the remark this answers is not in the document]
dylanhmorris: Incorrectly removed. Restored in 8c0f435

src/test/test_random_walk.py
dylanhmorris: Fair. This was a holdover from a previous implementation of `RandomWalk` that optionally sampled inits automagically. Can remove.

src/pyrenew/process/iidrandomsequence.py
damonbayer: @dylanhmorris You removed names in a different section than the one referenced here.

src/pyrenew/process/ar.py
```diff
@@ -4,97 +4,92 @@
...
 
         Parameters
         ----------
-        duration: int
+        n: int
             Length of the sequence.
-        inits : ArrayLike, optional
-            Initial points, if None, then these are sampled.
-            Defaults to None.
+        autoreg: ArrayLike
```
damonbayer, line 53: Seems to be written as though it will work with a multidimensional `autoreg` and `init_vals`. If this is the case, we should have a test for it. If not, add more checks regarding the dimensionality.

src/test/test_random_walk.py
dylanhmorris: Addressed in 75f14fb

src/pyrenew/process/ar.py
damonbayer, line 1: Without a `name`, I wonder if this needs to even be a class. In the future, I think we will have `name` as a part of all `RandomVariable`s, so probably safe to keep as a class.

src/test/test_ar_process.py
```diff
@@ -12,33 +13,73 @@ def test_ar_can_be_sampled():
...
+            init_vals=jnp.array([50.0, 49.9, 48.2]),
+            autoreg=jnp.array([0.05, 0.025, 0.025]),
+            noise_sd=[0.25],
+        )
+        ar3(
+            n=1230,
+            init_vals=jnp.array([50.0, 49.9, 48.2]),
+            autoreg=jnp.array([0.05, 0.025, 0.025]),
+            noise_sd=jnp.array([0.25]),
+        )
```
damonbayer, line 47: opportunities to DRY here

src/pyrenew/process/rtperiodicdiffar.py
```diff
@@ -179,17 +155,25 @@ def sample(
...
+        ar_diff = DifferencedProcess(
+            fundamental_process=ARProcess(name="first_diff_log_rt_ar"),
+            differencing_order=1,
+        )
+
+        log_rt = ar_diff(
+            n=n_periods,
+            init_vals=jnp.array([log_rt_rv[1]]),
+            autoreg=b,
+            noise_sd=s_r,
```
damonbayer, line 167: `b` and `s_r` could have better names

src/pyrenew/process/rtperiodicdiffar.py
```diff
@@ -179,17 +155,25 @@ def sample(
         n_periods = (duration + self.period_size - 1) // self.period_size
 
         # Running the process
-        ar_diff = FirstDifferenceARProcess(self.name, autoreg=b, noise_sd=s_r)
-        log_rt = ar_diff.sample(
-            duration=n_periods,
-            init_val=log_rt_rv[1],
-            init_rate_of_change=log_rt_rv[1] - log_rt_rv[0],
+        ar_diff = DifferencedProcess(
+            fundamental_process=ARProcess(name="first_diff_log_rt_ar"),
```
damonbayer, line 159: `name` should be configurable

src/test/test_rtperiodicdiff.py
```diff
@@ -1,16 +1,15 @@
...
 from jax.typing import ArrayLike
-from numpy.testing import assert_array_almost_equal, assert_array_equal
+from numpy.testing import assert_array_equal
 
 from pyrenew.deterministic import DeterministicVariable
-from pyrenew.process import RtWeeklyDiffProcess
+from pyrenew.process import RtWeeklyDiffARProcess
 
 
 def _manual_rt_weekly_diff(
```
damonbayer, line 15: This doesn't get used

src/pyrenew/process/rtperiodicdiffar.py
```diff
@@ -179,17 +155,25 @@ def sample(
...
+            differencing_order=1,
+        )
+
+        log_rt = ar_diff(
+            n=n_periods,
+            init_vals=jnp.array([log_rt_rv[1]]),
```
damonbayer, line 165:
```suggestion
            init_vals=jnp.array([log_rt_rv[0]]),
```

I think

src/pyrenew/process/ar.py
dylanhmorris: I think it can be made to work, but as written probably would not work "as expected". So for now I've made things stricter about 1D arrays and added tests that the errors get raised (in e0273ae)

src/test/test_ar_process.py
dylanhmorris: Agree but I think it can be a separate PR.

src/pyrenew/process/rtperiodicdiffar.py
dylanhmorris: Out of scope for this PR imo, since this was an existing class.

src/test/test_ar_process.py
dylanhmorris: Created #405

src/pyrenew/process/rtperiodicdiffar.py
dylanhmorris: correct. addressed in 8345a4f

src/pyrenew/process/rtperiodicdiffar.py
dylanhmorris: addressed in 8345a4f

src/test/test_rtperiodicdiff.py
dylanhmorris: Suggest a separate PR.

damonbayer: Per conversation on teams, let's just remove this function?

dylanhmorris: Removed in 7c62db6

src/pyrenew/process/iidrandomsequence.py
dylanhmorris: addressed in 27f0c52

damonbayer:
@dylanhmorris This is very close. Documentation needs a further look. For example, `:fun:` is not valid in Sphinx.

dylanhmorris:
> @dylanhmorris This is very close. Documentation needs a further look. For example, `:fun:` is not valid in Sphinx.

Fixed in 94131ce

damonbayer (approved):
Great work @dylanhmorris!

PR #387, opened 2024-08-16
record TrandformedVariable sample

### LINKED ISSUE #377: Record `TransformedRandomVariable`s.
The name argument to `TransformedRandomVariable` is unused. It should be used to record the sample.

Discovered here: https://github.com/CDCgov/multisignal-epi-inference/issues/376#issuecomment-2289977317

sbidari:
closes #377


model/src/pyrenew/metaclass.py
```diff
@@ -721,6 +720,10 @@ def sample(self, **kwargs) -> tuple:
             for t, uv in zip(self.transforms, untransformed_values)
         )
 
+        numpyro.deterministic(self.name, transformed_values[0].value)
```
dylanhmorris, line 723: 1. Make optional (following the same `if record` pattern used in `DeterministicVariable`
2. Iterate over tuple entries in some sensible way and name the `numpyro.deterministic` sites accordingly?

dylanhmorris (changes requested):

model/src/pyrenew/metaclass.py
sbidari: made these changes in https://github.com/CDCgov/multisignal-epi-inference/pull/387/commits/4b48a379b2b938f307e04a921c97c50bc29c1dde
Taking suggestions for better naming inside the loop (if any)!

model/src/pyrenew/metaclass.py
```diff
@@ -693,14 +693,17 @@ def __init__(
         self.transforms = transforms
         self.validate()
 
-    def sample(self, **kwargs) -> tuple:
+    def sample(self, record=True, **kwargs) -> tuple:
```
dylanhmorris, line 696: I'd suggest default `False`. I had thought we had an issue to change `DeterministicVariable` to that as well, but I couldn't find one, so I created it #389

dylanhmorris (changes requested):
One remaining change @sbidari. Apologies; I was unclear before.

model/src/pyrenew/metaclass.py
```diff
@@ -721,6 +723,12 @@ def sample(self, **kwargs) -> tuple:
             for t, uv in zip(self.transforms, untransformed_values)
         )
 
+        if record:
+            for i, tv in enumerate(transformed_values):
+                numpyro.deterministic(f"{self.name}_{i}", tv.value)
```
damonbayer, line 728: If the original `RandomVariable` `sample` is a `namedtuple`, can we re-use the name from the `namedtuple`?

sbidari: Do you mean use the same name as the untransformed `RandomVariable` instead of the `name` argument for the `TransformedRandonVariable`?

damonbayer: It's a bit confusing. `RandomVariable`s do not necessarily have a `name`.

`RandomVariable.sample()` always returns a tuple of `SampleValue`s. This tuple is sometimes a namedtuple.  This is the tuple that you are looping over in the code we are commenting on. I am wondering if we can use the names in the tuple as suffixes for the name in the `numpyro.deterministic` call.

sbidari: Ok, you mean use the `base_rv` name as a suffix! Will try

damonbayer (changes requested):
@sbidari Can you add tests to demonstrate that your naming code works as expected?

model/src/test/test_transformed_rv_class.py
```diff
@@ -61,6 +65,66 @@ def validate(self):
...
+    """
+
+    rv1: SampledValue | None = None
+    rv2: SampledValue | None = None
+
+    def __repr__(self):
+        return f"RVSamples(rv1={self.rv1},rv2={self.rv2})"
+
+
+class NamedBaseRV(RandomVariable):
```
damonbayer, line 80: Test for an un-named BaseRV?

sbidari: just added in https://github.com/CDCgov/multisignal-epi-inference/pull/387/commits/8dc1fdefc95cf36ef333cdbce20d3cb77a555957

sbidari: I realized I need to fix something here, will ping when ready for review

damonbayer: Test for a RandomVariable that returns a tuple of length > 1, but it is unnamed?

sbidari: I didn't find an example of a `RandomVariable` class that return an unnamed tuple length >1 in the codebase. If the tuple doesn't have `field` attribute, I recorded the first entry of the tuple with no suffix. Would you like for me to change that behavior?

damonbayer: I think we should handle this case, even if it is not currently present in the codebase.

model/src/pyrenew/metaclass.py
```diff
@@ -887,6 +889,19 @@ def sample(self, **kwargs) -> tuple:
...
+                    suffix = untransformed_values._fields[i]
+                    numpyro.deterministic(f"{self.name}_{suffix}", tv.value)
+            elif len(untransformed_values) == 1:
+                numpyro.deterministic(self.name, transformed_values[0].value)
+            else:
+                for i, tv in enumerate(transformed_values):
+                    numpyro.deterministic(f"{self.name}_{i}", tv.value)
+
+        return transformed_values
+
```
damonbayer, line 904: You can make this more compact by computing the `suffix` as either `untransformed_values._fields[i]` or `i` depending on ` hasattr(untransformed_values, "_fields")`, and then use a single `for i, tv in enumerate(transformed_values):` block.

damonbayer (approved):
Thanks @sbidari!

PR #388, opened 2024-08-16
Moving model to top level

gvegayon:
- Merges `./README.md` with `model/README.md`.
- Moves all the content under `model` to `.`.
- Adjust the paths for GHA to ensure they run under the new project structure.


README.md
```diff
@@ -1,21 +1,63 @@
-# Multisignal Renewal Project
+# PyRenew: A Package for Bayesian Renewal Modeling with JAX and Numpyro.
```
damonbayer, line 1:
```suggestion
# PyRenew: A Package for Bayesian Renewal Modeling with JAX and NumPyro.
```

.pre-commit-config.yaml
```diff
@@ -59,7 +59,6 @@ repos:
             "--baseline",
             ".secrets.baseline",
             "--exclude-files",
-            "model/docs/*_cache",
```
damonbayer, line 62: What does this change do?

damonbayer (changes requested):
@gvegayon Can you also update the installation instructions in `getting_started`?

.pre-commit-config.yaml
gvegayon: That was to exclude cache data generated by quarto. It was relevant when we had the tutorials under model/docs.

gvegayon:
> @gvegayon Can you also update the installation instructions in `getting_started`?

Good catch! Just did. Should be good now.

gvegayon:
Codecov is not uploading, so I'm guessing it's an issue on their side. All other test are passing.

damonbayer:
@gvegayon Can you remove the `pipeline` github actions?

PR #391, opened 2024-08-16
Dynamic and static distributional rvs

dylanhmorris:
This separates `DistributionalRV` into `StaticDistributionalRV`, in which the distribution is already instantiated (has fixed parameter values) and `DynamicDistributionalRV`, in which the distribution is instantiated when `sample()` is called with keyword arguments taken from the `sample()` call. The latter is designed to serve as a wrapper class for working with distributions with inferred parameters.

It also 
- Renames the `dist` keyword arg to `DistributionalRV` to `distribution`, to avoid clashes given the standard `numpyro` pattern of `import numpyro.distributions as dist`. 
- adds some basic tests for both subclasses

### DIFF
--- a/model/src/pyrenew/metaclass.py
+++ b/model/src/pyrenew/metaclass.py
@@ -5,7 +5,7 @@ pyrenew helper classes
 """
 
 from abc import ABCMeta, abstractmethod
-from typing import NamedTuple, get_type_hints
+from typing import Callable, NamedTuple, get_type_hints
 
 import jax
 import jax.numpy as jnp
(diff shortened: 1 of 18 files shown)


model/src/pyrenew/metaclass.py
```diff
@@ -135,7 +136,8 @@ class SampledValue(NamedTuple):
     t_start : int, optional
         The start time of the value.
     t_unit : int, optional
-        The unit of time relative to the model's fundamental (smallest) time unit.
+        The unit of time relative to the model's fundamental
+        (smallest) time unit.
```
damonbayer, line 140: Is this reformatting automated?

damonbayer (approved):
Really nice idea. Thanks @dylanhmorris!

PR #398, opened 2024-08-20
Test convolve mode in hospitaladmissionspy

### LINKED ISSUE #385: Incorrect convolve mode in `hospitaladmissions.py`
We should use `mode="valid"` to ensure the output reflects the entire `infection_to_admission_interval` being used.

https://github.com/CDCgov/multisignal-epi-inference/blob/main/model/src/pyrenew/latent/hospitaladmissions.py#L189-L193

I believe these lines should be replaced with

```python
        latent_hospital_admissions = jnp.convolve(
            latent_hospital_admissions_raw,
            infection_to_admission_interval.value,
            mode="valid",
        )
```

sbidari:
changes included:
- create a helper function for convolve `compute_incidence_observed_with_delay`
-  adds test for the helper function
-  update hospital admissions tutorial and DOW tutorials to use convolve `mode="valid"`

closes #385


src/pyrenew/metaclass.py
```diff
@@ -126,6 +126,38 @@ def _assert_sample_and_rtype(
     return None
 
 
+def compute_incidence_observed_with_delay(
```
damonbayer, line 129: `observed` isn't quite the right word. The result of this function is often something that is still `latent`.

src/pyrenew/metaclass.py
```diff
@@ -126,6 +126,38 @@ def _assert_sample_and_rtype(
...
+    incidence_to_observation_rate: ArrayLike,
```
damonbayer, line 130: I think `ratio` is more clear than `rate`.

src/pyrenew/metaclass.py
sbidari: Will use `compute_delay_ascertained_incidence` for now, we can rename if we come up with something better

src/pyrenew/metaclass.py
dylanhmorris: I disagree because of the x/y versus y/x problem. If you say "x_y_ratio", the baseline assumption is x/y. Here we typically represent things in units of observations per incidence (e.g. probability of hospital admissions given infection). If you call it "`ncidence_to_observation_ratio`, that implies incidence per observation, i.e. the reciprocal.

sbidari: will refrain from renaming based on @dylanhmorris's comment above.

damonbayer: I think `incidence_to_observation_rate` is easily confused with the concept of a reporting delay (the rate referring to the time to reporting rather than the proportion of reporting).

What about `ascertainment_ratio`, `reporting_ratio`, or `underreporting_ratio`? I'd also be fine with `ascertainment_rate`, `reporting_rate`, or `underreporting_rate`.

dylanhmorris:
@sbidari can you resolve the conflicts here?

src/pyrenew/metaclass.py
sbidari: I can change it to `ascertainment rate` to avoid possible confusion with `hospitalization_reporting_rate`.

It seems `rate/ratio` is used interchangeably in the codebase (e.g `latentadmissions.py`), we should probably change that for consistency.

dylanhmorris: Yes, can you make an issue for making this consistent?

Proposal: since these are Bayesian models, why don't we avoid the potential confusions associated with rates/ratios altogether and use conditional probability language:

`p_<x>_given_<y>`

sbidari: created #410

pyrenew/metaclass.py
```diff
@@ -125,6 +126,38 @@ def _assert_sample_and_rtype(
...
+        The rate at which latent incidences are observed.
+    latent_incidence: ArrayLike
+        Incidence values based on the true underlying process.
+    incidence_to_observation_delay_interval: ArrayLike
+        Pmf of delay interval between incidence to observation.
+
+    Returns
+    --------
+    ArrayLike
+        The incidence after the observation delay.
```
dylanhmorris, line 151:
```suggestion
        The predicted timeseries of delayed observations.
```

pyrenew/metaclass.py
```diff
@@ -125,6 +126,38 @@ def _assert_sample_and_rtype(
     return None
 
 
+def compute_delay_ascertained_incidence(
```
dylanhmorris, line 129: Would suggest moving this out of `metaclass.py`. Why not put it in `convolve.py`

test/test_model_hosp_admissions.py
```diff
@@ -522,12 +514,26 @@ def test_model_hosp_with_obs_model_weekday_phosp():
...
+    )
+
+    latent_infections = Infections()
+    Rt_process = SimpleRt()
+
+    observed_admissions = PoissonObservation("poisson_rv")
+
     # Other random components
-    total_length = n_obs_to_generate + pad_size + gen_int.size()
+    total_length = n_obs_to_generate + pad_size + 1  # gen_int.size()
```
dylanhmorris, line 536:
```suggestion
    total_length = n_obs_to_generate + pad_size + 1
```

dylanhmorris (dismissed):
Thanks, @sbidari. A few minor tweaks to docstrings and one major suggestion of function placement within the codebase. CC @damonbayer

src/pyrenew/metaclass.py
dylanhmorris: Thanks!

docs/source/tutorials/hospital_admissions_model.qmd
```diff
@@ -499,7 +496,7 @@ def compute_eti(dataset, eti_prob):
 
 fig, axes = plt.subplots(figsize=(6, 5))
 az.plot_hdi(
-    idata.prior_predictive["negbinom_rv_dim_0"] + gen_int.size(),
+    idata.prior_predictive["negbinom_rv_dim_0"],
```
damonbayer, line 499: I think these should have `+ n_initialization_points`. This is to make the time points across the plots agree. Alternatively, you could start with a negative time on the other plots.

sbidari: I removed the shift in x-axis so the plot with forecast has same time points as the posterior predictive plot. Will change the `latent_infection` plots to start from `- n_initialization_points`

pyrenew/metaclass.py
sbidari: Good point! Will move this function to `convolve.py`

docs/source/tutorials/hospital_admissions_model.qmd
sbidari: Actually, it's easier to have axis start from `n_initialization_points` than `-n_initialization_points` when plotting individual drwas using the `plot_posterior` method.

sbidari: I removed one of the plots with posterior draws of `latent_infections` in favor of uniformizing x-axis across all the figures but I can bring it back if we feel strongly about it.

damonbayer (changes requested):
Now that we made a decision around https://github.com/CDCgov/PyRenew/issues/410, can you adopt those changes for the new names in this PR?

pyrenew/convolve.py
```diff
@@ -165,3 +165,45 @@ def _new_scanner(
         return latest, (new_val, m_net1)
 
     return _new_scanner
+
+
+def compute_delay_ascertained_incidence(
+    p_observed_given_incident: ArrayLike,
+    latent_incidence: ArrayLike,
+    delay_incidence_to_observation_delay_pmf: ArrayLike,
```
damonbayer, line 173:
```suggestion
    delay_incidence_to_observation_pmf: ArrayLike,
```
throughout

pyrenew/convolve.py
```diff
@@ -168,9 +168,9 @@ def _new_scanner(
 
 
 def compute_delay_ascertained_incidence(
-    incidence_to_observation_rate: ArrayLike,
+    p_observed_given_incident: ArrayLike,
     latent_incidence: ArrayLike,
-    incidence_to_observation_delay_interval: ArrayLike,
+    delay_incidence_to_observation_delay_pmf: ArrayLike,
```
dylanhmorris, line 173: Can strike the second `delay`
```suggestion
    delay_incidence_to_observation_pmf: ArrayLike,
```

pyrenew/convolve.py
sbidari: sorry! I will go through it again

sbidari: fixed in https://github.com/CDCgov/PyRenew/pull/398/commits/4bc1736d0e1fd2485975cb9d18a44e021896fa8d

pyrenew/convolve.py
sbidari: fixed in https://github.com/CDCgov/PyRenew/pull/398/commits/4bc1736d0e1fd2485975cb9d18a44e021896fa8d

pyrenew/convolve.py
```diff
@@ -165,3 +165,45 @@ def _new_scanner(
...
+    on a delay interval.
+
+    Parameters
+    ----------
+    p_observed_given_incident: ArrayLike
+        The rate at which latent incident counts translate into observed counts.
+        For example, setting ``p_observed_given_incident=0.001``
+        when the incident counts are infections and the observed counts are
+        reported hospital admissions could be used to model disease and population
+        for which the probability (reported) hospital.admission given infection is
```
damonbayer, line 187:
```suggestion
        for which the probability of a latent infection leading to a reported hospital admission is
```

damonbayer: may need to fix line break after

pyrenew/convolve.py
```diff
@@ -165,3 +165,45 @@ def _new_scanner(
...
+        0.001.
+    latent_incidence: ArrayLike
+        Incidence values based on the true underlying process.
+    delay_incidence_to_observation_pmf: ArrayLike
```
damonbayer, line 191: Is this description accurate? Conceptually, I think it makes sense that there could be no delay from incidence to reporting. cc @dylanhmorris

dylanhmorris: True. I am in favor of allowing zero delay provided we document this clearly (and also document that we do _not_ allow zero-length generations, so generation interval PMFs are 1-indexed). The only thing this affects one way or the other is how the output vector of predicted observations is indexed in model time.

dylanhmorris: So yes, I favor rewording the docs here accordingly.

sbidari: I am not familiar with the concept of 0/1-indexed pmf. Did I miss anything in the edited docstring below? @dylanhmorris @damonbayer 
```
:math`i^{th}` entry represents a delay of :math:`i` 
time units, i.e. ``delay_incidence_to_observation_pmf[0]`` represents 
the fraction of observations that are not delayed (i.e delayed 0 time unit), 
``delay_incidence_to_observation_pmf[1]`` represents 
the fraction that are delayed 1 time unit, et cetera.
```

dylanhmorris: Yes, that looks good to me! thanks!

sbidari:
I realized `n_initialization_points` needed to be `max(gen_int_array.size, inf_hosp_int_array.size) - 1` instead of `max(gen_int_array.size, inf_hosp_int_array.size)` and made that change in https://github.com/CDCgov/PyRenew/pull/398/commits/5c4395ffc5d7a5b4d58afc1525e21028287f5449

pyrenew/convolve.py
```diff
@@ -165,3 +165,45 @@ def _new_scanner(
...
+        For example, setting ``p_observed_given_incident=0.001``
+        when the incident counts are infections and the observed counts are
+        reported hospital admissions could be used to model disease and population
+        for which the probability of a latent infection leading to a reported
+        hospital admission is 0.001.
+    latent_incidence: ArrayLike
+        Incidence values based on the true underlying process.
+    delay_incidence_to_observation_pmf: ArrayLike
+        Probability mass function of delay interval from incidence to observation,
+        where the :math`i^{th}` entry represents a delay of :math:`i`
```
damonbayer, line 193:
```suggestion
        where the :math:`i`th entry represents a delay of :math:`i`
```

damonbayer (approved):

pyrenew/convolve.py
```diff
@@ -165,3 +165,45 @@ def _new_scanner(
...
+        For example, setting ``p_observed_given_incident=0.001``
+        when the incident counts are infections and the observed counts are
+        reported hospital admissions could be used to model disease and population
+        for which the probability of a latent infection leading to a reported
+        hospital admission is 0.001.
+    latent_incidence: ArrayLike
+        Incidence values based on the true underlying process.
+    delay_incidence_to_observation_pmf: ArrayLike
+        Probability mass function of delay interval from incidence to observation,
+        where the :math:`i`th entry represents a delay of :math:`i`
```
damonbayer, line 193:
```suggestion
        where the :math:`i`\th entry represents a delay of :math:`i`
```

PR #406, opened 2024-08-21
Using TransformedRandomVariable

gvegayon:
- Uses `TransformedRandomVariable` in the day-of-the-week tutorial.
- Adds `record=True` in the call to `day_of_the_week_rv.sample()` in `latent.HospitalAdmissions`.


damonbayer (approved):
Thanks @gvegayon!

PR #416, opened 2024-08-28
Fix README numpyro link formatting in RST version

damonbayer:
RST doesn't let you link `code`-formatted text.


dylanhmorris (approved):
LGTM

PR #419, opened 2024-08-28
Cleanup `rtperiodicdiffar`

damonbayer:
- [x] Renames internal variables
- [ ] Removes `RtWeeklyDiffARProcess` entirely
- [ ] Makes  `log_rt_init_rv`, `autoreg_rv`, and `periodic_diff_sd_rv` into `non-RandomVariable` arguments to `RtPeriodicDiffARProcess`.

### DIFF
--- a/docs/source/tutorials/periodic_effects.qmd
+++ b/docs/source/tutorials/periodic_effects.qmd
@@ -27,7 +27,7 @@ from pyrenew import process, deterministic
 rt_proc = process.RtWeeklyDiffARProcess(
     name="rt_weekly_diff",
     offset=0,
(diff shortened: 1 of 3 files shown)


dylanhmorris:
> Could be good to have a test that makes use of both the vectorized differenced process and the vectorized AR process to make sure they work together. Perhaps as a test for `rtperiodicdiffar`. Could be done here or as part of https://github.com/CDCgov/PyRenew/pull/419

_Originally posted by @damonbayer in https://github.com/CDCgov/PyRenew/issues/446#issuecomment-2359268665_

PR #421, opened 2024-08-29
Use dev version of numpyro

damonbayer:
Use dev version of numpyro while we wait for https://github.com/pyro-ppl/numpyro/pull/1843 to make it to a released version


dylanhmorris (approved):

dylanhmorris (approved):
LGTM

PR #422, opened 2024-08-29
Remove UTF-8 Encoding Headers

AFg6K7h4fhy2:
Python3 uses UTF-8 encoding by default, so `# -*- coding: utf-8 -*-` in Python file headers is not needed. For clarity and minimalism of code, these are removed in this PR.


damonbayer (approved):
Thanks @AFg6K7h4fhy2!

PR #428, opened 2024-09-04
create a censored normal distribution

### LINKED ISSUE #427: Create a CensoredNormal distribution
The observation process for the log viral genome concentration in wastewater follow a censored normal distribution with concentration below the level of detection (LOD) censored. 
We can use @dylanhmorris's implementation of the censored normal distribution [here](https://github.com/dylanhmorris/host-viral-determinants/blob/42d4b47ad9e7d774bb8a9891ee78086eaaf1409a/src/distributions.py#L89-L144).

I am wondering if the censored normal distribution in `PyRenew` should inherit from the `RandomVariable` class instead of `numpyro.distributions.Distribution` so that it can be "time-aware" like all other randomvariables. Thoughts? @damonbayer @dylanhmorris @AFg6K7h4fhy2

sbidari:
closes #427

### DIFF
--- a/pyrenew/distributions/censorednormal.py
+++ b/pyrenew/distributions/censorednormal.py
@@ -61,7 +61,7 @@ class CensoredNormal(numpyro.distributions.Distribution):
         self.loc, self.scale = promote_shapes(loc, scale)
         self.lower_limit = lower_limit
         self.upper_limit = upper_limit
-
+self._support = constraints.interval(self.lower_limit, self.upper_limit)
         batch_shape = jax.lax.broadcast_shapes(
             jnp.shape(loc), jnp.shape(scale)
         )
--- a/pyrenew/distributions/censorednormal.py
+++ b/pyrenew/distributions/censorednormal.py
@@ -1,3 +1,5 @@
+# numpydoc ignore=GL08
+
 import jax
 import jax.numpy as jnp
 import numpyro
@@ -54,7 +56,9 @@ class CensoredNormal(numpyro.distributions.Distribution):
         self.lower_limit = lower_limit
         self.upper_limit = upper_limit
 
-        batch_shape = jax.lax.broadcast_shapes(jnp.shape(loc), jnp.shape(scale))
+        batch_shape = jax.lax.broadcast_shapes(
+            jnp.shape(loc), jnp.shape(scale)
+        )
         self.normal_ = numpyro.distributions.Normal(
             loc=loc, scale=scale, validate_args=validate_args
         )
@@ -95,7 +99,9 @@ class CensoredNormal(numpyro.distributions.Distribution):
         # we exploit the fact that for the
         # standard normal, P(x > a) = P(-x < a)
         # to compute the log complementary CDF
-        inbounds = jnp.logical_and(value > self.lower_limit, value < self.upper_limit)
+        inbounds = jnp.logical_and(
+            value > self.lower_limit, value < self.upper_limit
+        )
         result = jnp.where(inbounds, self.normal_.log_prob(value), lim_val)
 
         return result
--- a/pyrenew/distributions/censorednormal.py
+++ b/pyrenew/distributions/censorednormal.py
@@ -1,0 +1,101 @@
+import jax
+import jax.numpy as jnp
+import numpyro
+import numpyro.util
+from numpyro.distributions import constraints
+from numpyro.distributions.util import promote_shapes, validate_sample
+
+
+class CensoredNormal(numpyro.distributions.Distribution):
+    """
+    Censored normal distribution under which samples
+    are truncated to lie within a specified interval.
+    This implementation is adapted from
+    https://github.com/dylanhmorris/host-viral-determinants/blob/main/src/distributions.py
+    """
+
+    arg_constraints = {"loc": constraints.real, "scale": constraints.positive}
+    support = constraints.real
+
+    def __init__(
+        self,
+        loc=0,
+        scale=1,
+        lower_limit=-jnp.inf,
+        upper_limit=jnp.inf,
+        validate_args=None,
+    ):
+        """
+        Default constructor
+
+        Parameters
+        ----------
+        loc : ArrayLike or float, optional
+            The mean of the normal distribution.
+            Defaults to 0.
+        scale : ArrayLike or float, optional
+            The standard deviation of the normal
+            distribution. Must be positive. Defaults to 1.
+        lower_limit : float, optional
+            The lower bound of the interval for censoring.
+            Defaults to -inf (no lower bound).
+        upper_limit : float, optional
+            The upper bound of the interval for censoring.
+            Defaults to inf (no upper bound).
+        validate_args : bool, optional
+            If True, checks if parameters are valid.
+            Defaults to None.
+
+        Returns
+        -------
+        None
+        """
+        self.loc, self.scale = promote_shapes(loc, scale)
+        self.lower_limit = lower_limit
+        self.upper_limit = upper_limit
+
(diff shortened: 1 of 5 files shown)


SamuelBrand1:
I think we should call this _truncated_ normal; censoring has a special meaning in statistics so this could get confusing to users.

Typical usage is:

- _Truncated_ distribution: The conditional distribution of another distribution which has a wider support.
- Distribution _censoring_: Knowing the outcome of a random sample from the distribution only up to some censoring bounds.

I've got stuff really confused before, so I'm sensitive now! Wiki is good on this https://en.wikipedia.org/wiki/Censoring_(statistics)

SamuelBrand1:
Also, I think its worth reusing existing `jax` functions where possible for reducing code length and maybe getting optimization hacks.

Is it not possible to reuse https://jax.readthedocs.io/en/latest/_autosummary/jax.scipy.stats.truncnorm.logpdf.html and https://jax.readthedocs.io/en/latest/_autosummary/jax.random.truncated_normal.html#jax.random.truncated_normal ?

SamuelBrand1:
Ooops! Reread the issue, you are indeed doing censoring!

pyrenew/distributions/censorednormal.py
```diff
@@ -0,0 +1,107 @@
...
+class CensoredNormal(numpyro.distributions.Distribution):
+    """
+    Censored normal distribution under which samples
+    are truncated to lie within a specified interval.
+    This implementation is adapted from
+    https://github.com/dylanhmorris/host-viral-determinants/blob/main/src/distributions.py
+    """
+
+    arg_constraints = {"loc": constraints.real, "scale": constraints.positive}
+    support = constraints.real
```
dylanhmorris, line 20: This is my fault since it's inexact in the reference implementation, but we should have the support specified dynamically at initialization, since strictly speaking it depends on the `lower_limit` and `upper_limit` values.

I'd follow the pattern used for the `Uniform`
https://num.pyro.ai/en/stable/_modules/numpyro/distributions/continuous.html#Uniform

1. define a `self._support` attribute at initialization
2. Define a property method
```python
    @constraints.dependent_property(is_discrete=False, event_dim=0)
    def support(self):
        return self._support
```

We probably also want to declare `lower_limit`, `upper_limit`, and `support` as `pytree_data_fields`.

dylanhmorris (changes requested):
One quick thing to address before a full review.

damonbayer:
Does this differ in any substantial way from the example in the NumPyro docs: https://num.pyro.ai/en/latest/tutorials/censoring.html ? I have not reviewed this PR or the doc yet, but am curious if there are any major differences.

dylanhmorris:
> Does this differ in any substantial way from the example in the NumPyro docs: https://num.pyro.ai/en/latest/tutorials/censoring.html ? I have not reviewed this PR or the doc yet, but am curious if there are any major differences.

1. Docs do it as a modeling problem using only bundled distributions; this creates a `numpyro.distributions.Distribution` subclass to handle it directly. I strongly favor the second approach where possible for modularity / texting / reusability.
2. Does not use a boolean array to indicate which values are censored and instead treats any values at or outside the given censoring bounds as censored.

Would be a nice feature for numpyro to have semi-automated creation of censored distributions from base distributions (as it currently has has for truncated distributions)

pyrenew/distributions/censorednormal.py
sbidari: > We probably also want to declare `lower_limit`, `upper_limit`, and `support` as `pytree_data_fields`.

shouldn't `loc` and `scale` also be declared as `pytree_data_fields` if we are using it?

pyrenew/distributions/censorednormal.py
```diff
@@ -0,0 +1,117 @@
...
+            Defaults to None.
+
+        Returns
+        -------
+        None
+        """
+        self.loc, self.scale = promote_shapes(loc, scale)
+        self.lower_limit = lower_limit
+        self.upper_limit = upper_limit
+
```
dylanhmorris, line 64:
```suggestion
self._support = constraints.interval(self.lower_limit, self.upper_limit)
```

dylanhmorris (changes requested):
Missing the line where we define `self._support`

pyrenew/distributions/censorednormal.py
sbidari: my bad! added in https://github.com/CDCgov/PyRenew/pull/428/commits/7d4e7aef7e74c22d044a32a5a1ef1c573bd93d8f

pyrenew/distributions/censorednormal.py
```diff
@@ -0,0 +1,119 @@
...
+            jnp.shape(loc), jnp.shape(scale)
+        )
+        self.normal_ = numpyro.distributions.Normal(
+            loc=loc, scale=scale, validate_args=validate_args
+        )
+        super().__init__(batch_shape=batch_shape, validate_args=validate_args)
+
+    @constraints.dependent_property(is_discrete=False, event_dim=0)
+    def support(self):  # numpydoc ignore=GL08
+        return self._support
```
dylanhmorris, line 77: Add a unit test of this? Would've caught the previous potential bug. Notice that codecov flags it.

dylanhmorris (changes requested):
Unit test for the support property?

pyrenew/distributions/censorednormal.py
sbidari: added more tests in https://github.com/CDCgov/PyRenew/pull/428/commits/c336e54023b2c832bac2c884345d3213961b3d36

test/test_censorednormal.py
```diff
@@ -0,0 +1,168 @@
...
+    )
+
+    # test log prob of values lower than the limit
+    assert_array_almost_equal(
+        censored_dist.log_prob(l_val),
+        jax.scipy.special.log_ndtr((lower_limit - loc) / scale),
+    )
+
+    # test support of the distribution
+    assert_equal(censored_dist.support.lower_bound, lower_limit)
```
dylanhmorris, line 127:
```suggestion
    assert_equal(censored_dist.support.lower_bound, lower_limit)
    assert censored_dist.support.upper_bound >= jnp.inf
```

test/test_censorednormal.py
```diff
@@ -0,0 +1,168 @@
...
+    )
+
+    # test log prob of values higher than the limit
+    assert_array_almost_equal(
+        censored_dist.log_prob(h_val),
+        jax.scipy.special.log_ndtr(-(upper_limit - loc) / scale),
+    )
+
+    # test support of the distribution
+    assert_equal(censored_dist.support.upper_bound, upper_limit)
```
dylanhmorris, line 168:
```suggestion
    assert_equal(censored_dist.support.upper_bound, upper_limit)
    assert censored_dist.support.lower_bound <= -jnp.inf
```

dylanhmorris (changes requested):
A couple suggested additions to the support testing, otherwise looks good. Thanks @sbidari!

test/test_censorednormal.py
sbidari: I think the greater/less than signs are flipped here. Will add these comparison tests

sbidari: Why not use `assert censored_dist.support.upper_bound == jnp.inf`? The `>= jnp.inf` doesn't make sense to me since it is never going to be greater than infinity

dylanhmorris: Sure.

dylanhmorris (approved):
LGTM, thanks @sbidari!

PR #431, opened 2024-09-05
Further Prune PostProcessing Code, Specifically plot_posterior And spread_draws

AFg6K7h4fhy2:
This PR:

> In an effort to leave post-processing to the user, DB + others have determined that `plot_posterior` and `spread_draws`, stemming from the mcmcutils.py and being used partially in metaclass.py can be pruned.


dylanhmorris:
```python
 pyproject.toml: DEP002 'matplotlib' defined as a dependency but not used in the codebase
```
@AFg6K7h4fhy2 can you remove?

damonbayer:
@dylanhmorris it is used in the tutorials, which I don't think deptry recognizes

dylanhmorris:
Remove from the main group and add to `docs` then?

docs/source/tutorials/basic_renewal_model.qmd
```diff
@@ -252,7 +252,38 @@ Now, let's investigate the output, particularly the posterior distribution of th
...
+import arviz as az
+
+idata = az.from_numpyro(
+    posterior=model1.mcmc,
+)
+
+# Extract Rt signal
+rt = idata.posterior["Rt"]
+rt = rt.mean(dim="chain")
+rt = rt.T
```
damonbayer, line 264: I don't understand why you take the mean here?

docs/source/tutorials/day_of_the_week.qmd
```diff
@@ -197,16 +199,56 @@ hosp_model.run(
...
+    posterior_predictive=ppc_samples,
+)
 
-# Plotting the posterior
-out = hosp_model.plot_posterior(
-    var="latent_hospital_admissions",
-    ylab="Hospital Admissions",
-    obs_signal=daily_hosp_admits.astype(float),
+# Convert hospital admissions from discrete to continuous
+idata.posterior_predictive["negbinom_rv"] = idata.posterior_predictive[
```
damonbayer, line 222: I assume this is necessary, but it's surprising to me.

docs/source/tutorials/basic_renewal_model.qmd
AFg6K7h4fhy2: I could have used `rt.squeeze()` if there was a single chain but there were 2 chains.

docs/source/tutorials/day_of_the_week.qmd
```diff
@@ -197,16 +199,56 @@ hosp_model.run(
...
+    "negbinom_rv"
+].astype(float)
+
+# Use a time series plot (plot_ts) from arviz for plotting
+axes = az.plot_ts(
+    idata,
+    y="negbinom_rv",
+    y_hat="negbinom_rv",
+    num_samples=150,
+    y_kwargs={"color": "red", "linewidth": 2.5},
```
damonbayer, line 232: The color scheme is a little intense.

docs/source/tutorials/day_of_the_week.qmd
```diff
@@ -280,34 +322,99 @@ As a result, we can see the posterior distribution of our novel day-of-the-week
...
+dow_idata = az.from_numpyro(
+    posterior=hosp_model_dow.mcmc,
 )
 
-sp = hosp_model_dow.spread_draws(["dayofweek_effect_raw"])
-# dayofweek_effect is not recorded
+# Extract day of week effect (DOW)
+dow_effect_raw = dow_idata.posterior["dayofweek_effect_raw"]
+dow_effect_raw = dow_effect_raw.squeeze()
+dow_effect_raw = dow_effect_raw.T
```
damonbayer, line 333: just chain them all together `dow_idata.posterior["dayofweek_effect_raw"].squeeze().T`

docs/source/tutorials/day_of_the_week.qmd
AFg6K7h4fhy2: Screenshot 2024-10-02 at 15 47 22

This is what it looks like (not sure if you were looking at the code after running, so attaching image).

docs/source/tutorials/day_of_the_week.qmd
AFg6K7h4fhy2: Will do.

docs/source/tutorials/day_of_the_week.qmd
damonbayer: I also prefer for observed values to be dots rather than lines, but that's optional.

docs/source/tutorials/day_of_the_week.qmd
AFg6K7h4fhy2: Need to check again actually. This might be an artifact from having first used `plot_ppc`, which differentiates plots based on whether the data received are integers or floats.

docs/source/tutorials/day_of_the_week.qmd
```diff
@@ -280,34 +322,99 @@ As a result, we can see the posterior distribution of our novel day-of-the-week
...
+indices = np.random.choice(dow_effect_raw.shape[1], size=200, replace=False)
+dow_plot_samples = dow_effect_raw[:, indices]
+fig, ax = plt.subplots(1, 1, figsize=(8, 6))
```
damonbayer, line 336: Is there a reason this isn't done with `plot_ts` like the other plots?

docs/source/tutorials/basic_renewal_model.qmd
damonbayer: but why do you want to take the mean? The legend makes it look like you want to plot individual samples.

docs/source/tutorials/day_of_the_week.qmd
AFg6K7h4fhy2: > I also prefer for observed values to be dots rather than lines, but that's optional.

I didn't know this, but will change it.

damonbayer: I suggest using one of the arviz color schemes https://python.arviz.org/en/stable/examples/styles.html

docs/source/tutorials/basic_renewal_model.qmd
AFg6K7h4fhy2: Okay, then would you like me to choose chain 0 or 1? Or, would you like me to update the legend?

dylanhmorris: In general for this sort of spaghetti plot you:
- want to plot individual samples as individual pieces of spaghetti
- in that context, want to treat chain 0 samples, chain 1 samples, etc equivalently, and plot (typically a random subset of) samples from all chains.

There are cases in which you might want to plot samples from different chains in distinguishable ways (e.g. for non-convergence debugging), but those are beyond the scope of this example.

damonbayer: Agree with Dylan. You can randomly sample from all samples with `az.extract(idata, num_samples=100)`.

docs/source/tutorials/day_of_the_week.qmd
AFg6K7h4fhy2: Yes, looking over [this link](https://python.arviz.org/en/latest/api/generated/arviz.plot_ts.html) is a semi-prerequisite for the below. 

The `y` parameter operates from variables in the xarray dataset `observed_data`, which is standard to `InferenceData` objects. All instances of creating `idata` using `az.from_numpyro()` have had the variables `negbinom_rv` and its dimensions in the the `observed_data` dataset. 

Some data variables we want from the posterior (e.g. `Rt` or `dayofweek_effect_raw`) are not in the `observed_data` dataset, but rather in the xarray dataset `idata.posterior`. 

How then may be connect `idata.posterior["variable_we_want_to_plot"]` to `az.plot_ts(...)`? The method I opted for was to use `idata.assign(observed_data=<xarray_dataset_for_the_posterior>` (using `plot_ts()` in a way I presume was not intended). This leads to further issues, since the dimensions of the `observed_data` dataset will include the original dimensions. An example: Dimensions:  (negbinom_rv_dim_0: 90, chain: 1, draw: 2000, dayofweek_effect_raw_dim_0: 7). This "multidimensional" `y` requires then the use of `plot_dim` and `x`, which are usually manual. 

I opt for brevity to not use `plot_ts()` in instances like this.

docs/source/tutorials/basic_renewal_model.qmd
AFg6K7h4fhy2: Okay, I like this solution, will do.

damonbayer: > There are cases in which you might want to plot samples from different chains in distinguishable ways (e.g. for non-convergence debugging), but those are beyond the scope of this example.

Yes. You may sometimes want to
1. compute a summary (e.g. mean) across all samples in each chain to verify the different chains produce similar results
2. compute a summary across all samples and all chains to see the overall posterior

Computing a summary across all chains only (not the samples within a chain) doesn't sound immediately useful to me.

AFg6K7h4fhy2: The mean is as though you concatenated vertically the samples in both chains. But yes, I see that extracting N samples from across the chains and plotting them individually is what is desired.

docs/source/tutorials/day_of_the_week.qmd
AFg6K7h4fhy2: Some plot styles (for variables not in `observed_data`) have styles I've chosen. For these plots, what would you like besides "a seemingly harsh" color scheme?

AFg6K7h4fhy2: This sounds good.

damonbayer: I think you could specify `"color": "skyblue" ` and it would look decent https://matplotlib.org/stable/gallery/color/named_colors.html

docs/source/tutorials/day_of_the_week.qmd
```diff
@@ -280,34 +322,105 @@ As a result, we can see the posterior distribution of our novel day-of-the-week
...
+    np.arange(dow_effect_raw.shape[0]),
+    dow_plot_samples,
+    color="gray",
+    alpha=0.25,
+)
+ax.plot([], [], color="gray", alpha=0.25, label="DOW Posterior Samples")
+ax.plot(
+    np.arange(dow_effect_raw.shape[0]),
+    dow_plot_samples.mean(dim="draw"),
+    color="red",
```
damonbayer, line 345: this plot still has the old colors

damonbayer (changes requested):
one more plot to fix, then it will be ready to merge! Thanks @AFg6K7h4fhy2

damonbayer (approved):
Nice work @AFg6K7h4fhy2

PR #432, opened 2024-09-05
Reworks `InitializeInfectionsExponentialGrowth` to work with `numpyro.plate`

damonbayer:


scratch/plate_init_demo.py
```diff
@@ -0,0 +1,75 @@
...
+import numpyro.distributions as dist
+import polars as pl
+from jax import random
+from numpyro.infer import MCMC, NUTS
+
+from pyrenew.latent import InitializeInfectionsExponentialGrowth
+from pyrenew.randomvariable import DistributionalVariable
+
+n_groups = 3  # works fine
+# n_groups = 4 # totally fails
```
damonbayer, line 17: @dylanhmorris @sbidari I am getting some very strange behavior where this demo works perfectly with <= 3 groups, but inference completely fails with >=4 groups. Do you have any insights?

sbidari: I am able to run it with n_groups = 4 and 6

damonbayer: It runs but the posterior inference is way off. Compare the output of `mcmc.print_summary()` with the true values of `rates` and `i0s`.

## For 3 Groups:

```pycon
                mean       std    median      5.0%     95.0%     n_eff     r_hat
     i0[0]      1.00      0.06      1.00      0.90      1.08    291.76      1.01
     i0[1]      2.00      0.07      2.00      1.87      2.11    359.97      1.00
     i0[2]      2.99      0.06      2.99      2.88      3.08    466.37      1.00
   rate[0]      1.10      0.01      1.10      1.09      1.11    289.96      1.01
   rate[1]      1.01      0.00      1.01      1.00      1.02    373.89      1.00
   rate[2]      1.00      0.00      1.00      1.00      1.00    465.28      1.00

Number of divergences: 0
```

Compared to true values
```pycon
i0s: Array([1, 2, 3], dtype=int32)
rates: Array([1.1  , 1.01 , 1.001], dtype=float32)
```

## For 4 Groups:

```pycon
                mean       std    median      5.0%     95.0%     n_eff     r_hat
     i0[0]      1.21      0.00      1.21      1.21      1.21      0.50      1.00
     i0[1]      4.86      0.00      4.86      4.86      4.86      0.50      1.00
     i0[2]      0.27      0.00      0.27      0.27      0.27      0.50      1.00
     i0[3]      0.73      0.00      0.73      0.73      0.73      0.50      1.00
   rate[0]      1.25      0.00      1.25      1.25      1.25      0.50      1.00
   rate[1]      1.01      0.00      1.01      1.01      1.01      2.56      1.75
   rate[2]      1.23      0.00      1.23      1.23      1.23      0.50      1.00
   rate[3]      0.33      0.00      0.33      0.33      0.33      0.50      1.00

Number of divergences: 0
```

Compared to true values
```pycon
i0s: Array([1, 2, 3, 4], dtype=int32)
rates: Array([1.1   , 1.01  , 1.001 , 1.0001], dtype=float32)
```

sbidari: ops, I didn't look at the values earlier, confirming I get the similar values

damonbayer:
@dylanhmorris @sbidari docs are not rendering. Probably worth a broader discussion about what components of PyRenew we expect to "just work" in plates before I attempt a fix.

dylanhmorris:
> @dylanhmorris @sbidari docs are not rendering. The problem is that the shape of the output of the initialization method for a non-plated context has changed from `(x,)` to `(x,1)` (in a plated context is will be `(x, n_plates)`). Probably worth discussing a general plan for handling this type of change throughout PyRenew before I attempt a fix.

Is there a reason not to use [`jnp.squeeze`](https://jax.readthedocs.io/en/latest/_autosummary/jax.numpy.squeeze.html) in contexts like this?

dylanhmorris (changes requested):
Given the content of the change, could we have tests both for the case in which `rate` is a scalar and for the case in which it is an array?

dylanhmorris (approved):
LGTM. Tests could probably be DRY-ified via `pytest.mark.parametrize` but that can be a separate PR.

damonbayer:
> LGTM. Tests could probably be DRY-ified via `pytest.mark.parametrize` but that can be a separate PR.

I thought about that, but the logic to determine the manual computations seemed nearly as verbose as the fully manual solution, while being less readable.

PR #439, opened 2024-09-10
Vectorize `ARProcess()`

dylanhmorris:
This will allow it to play better with `numpyro.plate` (still not seamlessly, [see here](https://num.pyro.ai/en/stable/primitives.html#scan)), among other things

- This follows the convention that whenever a multi-dim array is a timeseries, time is the first dimension. One reason for that convention (@sbidari maybe worth noting in #443): By default, `numpyro.scan` builds by augmenting the first dimension. So since we most often scan over time, this convention will make `scan`s over time have sensible default output.
- Adds a bunch of tests for correctness
- Adds informative error messages when shapes cannot be broadcast properly
- Adds tests for the errors.
- Note for PRs such as #443: clearest way to take dot products along only certain axes of a multidimensional array of unknown dimensions appears to be via `jnp.einsum`. I think we should use it wherever that's the task https://github.com/CDCgov/PyRenew/pull/439#discussion_r1757223924


pyrenew/process/ar.py
```diff
@@ -52,60 +48,113 @@ def sample(
...
         Returns
         -------
         ArrayLike
+            of shape (n,) + init_vals.shape[1:].
         """
-        noise_sd_arr = jnp.atleast_1d(noise_sd)
-        if not noise_sd_arr.shape == (1,):
-            raise ValueError("noise_sd must be a scalar. " f"Got {noise_sd}")
         autoreg = jnp.atleast_1d(autoreg)
+        noise_sd = jnp.array(noise_sd)
```
damonbayer, line 62: Why are these not all `jnp.atleast_1d` or all `jnp.array`?

pyrenew/process/ar.py
```diff
@@ -52,60 +48,113 @@ def sample(
...
+                "to the expected shape of the process "
+                f"history (shape {history_shape}). "
+                "History shape is determined by the "
+                "shapes of the init_vals, autoreg, and "
+                "noise_sd arrays. See ARProcess "
+                "documentation for details"
+            ) from e
+
+        inits_flipped = jnp.flip(inits_broadcast, axis=0)
+        assert jnp.shape(inits_flipped) == history_shape
```
damonbayer, line 110: Appreciate the informative error messages in this PR, but this one appears to be overlooked.

pyrenew/process/ar.py
```diff
@@ -52,60 +48,113 @@ def sample(
...
+            with numpyro.handlers.reparam(
+                config={noise_name: LocScaleReparam(0)}
+            ):
+                next_noise = numpyro.sample(
+                    noise_name,
+                    numpyro.distributions.Normal(
+                        loc=jnp.zeros(noise_shape), scale=noise_sd
+                    ),
+                )
+                assert jnp.shape(next_noise) == noise_shape
```
damonbayer, line 122: Better error message?

pyrenew/process/ar.py
```diff
@@ -52,60 +48,113 @@ def sample(
...
+        assert jnp.shape(inits_flipped) == history_shape
+
+        def transition(recent_vals, _):  # numpydoc ignore=GL08
+            with numpyro.handlers.reparam(
+                config={noise_name: LocScaleReparam(0)}
+            ):
+                next_noise = numpyro.sample(
+                    noise_name,
+                    numpyro.distributions.Normal(
+                        loc=jnp.zeros(noise_shape), scale=noise_sd
```
damonbayer, line 119:
```suggestion
                        scale=noise_sd
```

`loc` is automatically set to `0` of the correct dimension if not provided.

pyrenew/process/ar.py
```diff
@@ -52,60 +48,113 @@ def sample(
...
+                    ),
+                )
+                assert jnp.shape(next_noise) == noise_shape
+
+            dot_prod = jnp.einsum("i...,i...->...", autoreg, recent_vals)
```
damonbayer, line 124: Haven't seen this before. Should we be looking for other places to use it?

pyrenew/process/ar.py
```diff
@@ -52,60 +48,113 @@ def sample(
...
+
+            new_term = dot_prod + next_noise
+
+            assert new_term.shape == noise_shape
```
damonbayer, line 128: Better error message? In general I'm wondering when you think these asserts might fail. Can the right shapes be guaranteed by checking only the inputs to our functions, rather than the outputs?

pyrenew/process/ar.py
```diff
@@ -52,60 +48,113 @@ def sample(
...
+                axis=0,
+            )[:order]
+
+            assert new_recent_vals.shape == history_shape
             return new_recent_vals, new_term
 
-        last, ts = scan(transition, init_vals, noise)
-        return jnp.hstack([init_vals, ts])
+        if n > order:
+            last, ts = scan(
```
damonbayer, line 144: don't need to store `last`.

pyrenew/process/ar.py
```diff
@@ -52,60 +48,113 @@ def sample(
...
+                length=(n - order),
+            )
+
+            ts_with_inits = jnp.concatenate(
+                [inits_broadcast, ts],
+                axis=0,
+            )
+        else:
+            ts_with_inits = inits_broadcast
+        return ts_with_inits[:n]
```
damonbayer, line 157: light preference for `else: ts = None` and always doing  `ts_with_inits  
 = jnp.concatenate([inits_broadcast, ts], axis=0)`.

pyrenew/process/ar.py
dylanhmorris: Yes, thanks. Meant to flag this. I think it's the best way to get what we need with dot products but only along certain axes. Will flag in the main PR message.

pyrenew/process/ar.py
dylanhmorris: It's not currently set anywhere though. Are you suggesting it would be clearer to set via `.expand()` than via the shape of the `loc`?

pyrenew/process/ar.py
damonbayer: Thinking a bit more.

`autoreg` can be a vector for each `plate` element, while `noise_sd` should be scalar for each `plate` element, so I understand there might be a difference for how they are handled. Not grasping why this is the difference, though.

dylanhmorris: `noise_sd` can be a scalar, but the other two need a first dimension (time), even if it's length 1, because that first dimension length gives the order of the process.

dylanhmorris: clarifying comment added to code in 3023a3d

pyrenew/process/ar.py
damonbayer: Understood. I think I prefer to take them out.

pyrenew/process/ar.py
dylanhmorris: added a clarifying comment on this point to the code in 3023a3d

pyrenew/process/ar.py
dylanhmorris: Addressed in f64c839

pyrenew/process/ar.py
dylanhmorris: Addressed in 461e024

pyrenew/process/ar.py
dylanhmorris: Addressed in 461e024

pyrenew/process/ar.py
dylanhmorris: Done in 461e024

pyrenew/process/ar.py
dylanhmorris: @damonbayer this is the only question and then should be review for re-review

damonbayer: I think I misunderstood initially. I thought `noise_shape` was the shape of `noise_sd`, but it is something else.

dylanhmorris:
> Since one of the intended use cases of this extension is compatibility with `numpyro.plate` I wonder if we should do some tests with `plate`s (and more complicated setups like nested `plate`s. I'm not immediately sure that all this will work).

Good point will add

damonbayer:
> Good point will add

I might even suggest that `plate` be the only way we do multi-dimensional testing.

dylanhmorris:
> > Good point will add
> 
> I might even suggest that `plate` be the only way we do multi-dimensional testing.

I wouldn't go that far; we want vectorizations / multi-dim behavior to be as expected outside of `plate` contexts too, no?

damonbayer (approved):

AFg6K7h4fhy2 (commented):
Confirming that I have seen all changes made in this PR and (mostly) have absorbed the updates.

Issue #189, opened 2024-06-13
Additional tests for convolve scanner factory functions


damonbayer:
Mentioned as "out of scope" here: https://github.com/CDCgov/multisignal-epi-inference/pull/161

Issue #394, opened 2024-08-19
Tests for `InfectionsWithFeedback` class


dylanhmorris:
All existing tests were able to pass with a bug in the `InfectionsWithFeedback` class. Only attempting to render the `extending_pyrenew.qmd` tutorial revealed it. We should have:
1. A test confirming that `InfectionsWithFeedback` can be instantiated and sampled.
2. A test comparing infections with feedback computed using the class to a manual computation.

dylanhmorris:
This is a lesson to me always to write a reprex, since the tests you're pointing to are almost exactly what I described and evidently didn't catch the bug. Let me investigate.

damonbayer:
@dylanhmorris Were you able to investigate?

PR #440, opened 2024-09-10
update infectionswithfeedback process 

sbidari:
This PR modifies the following functions/files such that these now work when `I0` and `Rt` are 2D arrays as is the case for site-level hierarchical model.

- convolve.py : `new_convolve_scanner` and `new_double_convolve_scanner`
- arrayutils.py : added a function `pad_edges_to_match` that pads multidimensional arrays using speficied edge values
-  `Infections` and `InfectionWithFeedback` classes in latent module
- add tests of these functions using 2D arrays where appropriate
-  add more tests for convolve scanner functions (#189)
- add additional tests for `Infection` and `InfectionWithFeedback` classes

**Out of scope**
- multidimensional generational intervals


pyrenew/arrayutils.py
```diff
@@ -8,6 +8,66 @@
 from jax.typing import ArrayLike
 
 
+def pad_edges_to_match(
```
damonbayer, line 11: Could this be combined with `pad_to_match` to be a more flexible extension to `jnp.pad`?

damonbayer: It looks like the only current use of `pad_x_to_match_y` is in `extending_pyrenew.qmd`. That use could be replaced with `pad_edges_to_match`, and `pad_x_to_match_y` could be removed.

damonbayer: `pad_to_match` also appears to be unused. Let's remove `pad_to_match` and `pad_x_to_match_y` and replace with `pad_edges_to_match`.

test/test_infectionsrtfeedback.py
```diff
@@ -39,32 +40,53 @@ def _infection_w_feedback_alt(
...
-
-        I_vec[t + len_gen] = Rt_adj[t] * np.dot(
-            I_vec[t : t + len_gen], np.flip(gen_int)
-        )
-
-    return {"post_initialization_infections": I_vec[I0.size :], "rt": Rt_adj}
-
-
-def test_infectionsrtfeedback():
+    I_vec = np.concatenate([I0.reshape(T, -1), np.zeros(Rt.shape)])
```
damonbayer, line 48: Probably faster and more readable to use`jnp.pad`.

test/test_infectionsrtfeedback.py
```diff
@@ -39,32 +40,53 @@ def _infection_w_feedback_alt(
...
+    Rt_adj = np.zeros(Rt.shape)
+    inf_feedback_strength = np.array(inf_feedback_strength).reshape(T, -1)
+
+    for n in range(Rt.shape[1]):
+        for t in range(Rt.shape[0]):
```
damonbayer, line 53: Prefer some functional programming or list comprehension type thing rather than initializing a 0 `Rt_adj` and filling it in later.

test/test_infectionsrtfeedback.py
damonbayer: Actually, prefer not to pad `I_vec` at all. See later comments.

damonbayer: Also, can we have a better name than `I_vec`?

test/test_infectionsrtfeedback.py
```diff
@@ -39,32 +40,53 @@ def _infection_w_feedback_alt(
...
+                inf_feedback_strength[t, n]
+                * np.dot(I_vec[t : t + len_gen, n], np.flip(inf_feedback_pmf))
+            )
+
+            I_vec[t + len_gen, n] = Rt_adj[t, n] * np.dot(
+                I_vec[t : t + len_gen, n], np.flip(gen_int)
+            )
+
+    return {
+        "post_initialization_infections": np.squeeze(I_vec[I0.shape[0] :]),
```
damonbayer, line 64: Can we just construct the `I_vec[I0.shape[0] :]` portion without padding `I_vec` and then overwriting it?

damonbayer (changes requested):
Looks nice @sbidari, but I think there are some opportunities to improve performance and clarity.

damonbayer:
@sbidari Can you also look into addressing https://github.com/CDCgov/PyRenew/issues/394 in this PR?

test/test_infectionsrtfeedback.py
sbidari: I can't think of a straightforward way of doing this since we use sliding window of values from a dynamically constructed array $[I_{- \text{init time}} ,\cdots,I_1,I_2,\cdots]$. Let me know if you have ideas.
Also, given this is meant to be a 'primitive' method to test values computed using a more efficient method, it may not be worth spending a lot of time here?

damonbayer: Good point about this being part of tests. I didn't think of that when I was reviewing.

test/test_infectionsrtfeedback.py
sbidari: I tried using list comprehension rather than initialization and filling in in https://github.com/CDCgov/PyRenew/pull/440/commits/81970595f222ad498753f9974e68b649bd39b3c2  but it doesn't seem to work and the fix is not obvious to me.
Again since this is just for testing, maybe gain in efficiency is not a priority here?

test/test_infectionsrtfeedback.py
sbidari: > Probably faster and more readable to use `jnp.pad`.

changed to use `np.pad`

sbidari: > Also, can we have a better name than `I_vec`?

renamed in https://github.com/CDCgov/PyRenew/pull/440/commits/97d19d05ed295cd7547376ff4958ea347c743bba

sbidari:
> @sbidari Can you also look into addressing #394 and #189 in this PR?

Added more tests for the convolve factory functions in https://github.com/CDCgov/PyRenew/pull/440/commits/d933df457b666addc862dee4f301ac11252f24e8

damonbayer:
@sbidari Can we also have tests that try to use `numpyro.plate` and make sure things work as expected?

pyrenew/latent/infectionswithfeedback.py
```diff
@@ -167,6 +167,10 @@ def sample(
                 **kwargs,
             )
         )
+
+        if inf_feedback_strength.ndim < Rt.ndim:
+            inf_feedback_strength = jnp.expand_dims(inf_feedback_strength, 0)
```
damonbayer, line 172: Will this work if `Rt.ndim - inf_feedback_strength.ndim > 1`?

sbidari: work in the sense, these particular lines of code will just add a dimension to `inf_feedback_strength` along `axis=0`.
The code does not work and will raise an error because of this line further down
```
inf_feedback_strength.size != Rt.size
```

damonbayer: In that case, I think the check here should just be 
```python
 if inf_feedback_strength.ndim ==  Rt.ndim -1:
```
so you can catch issues earlier.

test/test_convolve_scanners.py
```diff
@@ -39,10 +52,27 @@ def transform_a(x: any):
         """
         return 4 * x + 0.025
 
+    def transform_ones_like(x: any):
```
damonbayer, line 55: Why is this function needed? Can't we just replace its use with `jnp.ones_like(x)`?

damonbayer (changes requested):
A few open questions. I would also like @dylanhmorris to take a look, if he can, since he just did similar work on the AR class.

pyrenew/latent/infectionswithfeedback.py
sbidari: Good point! changed in https://github.com/CDCgov/PyRenew/pull/440/commits/cc61d0b2f3a973109808ffa2c222aaeb319be3c3

pyrenew/latent/infectionswithfeedback.py
```diff
@@ -143,11 +143,18 @@ def sample(
         InfectionsWithFeedback
             Named tuple with "infections".
         """
-        if I0.size < gen_int.size:
+        if I0.shape[0] < gen_int.size:
```
dylanhmorris, line 146: Multi-dim generation intervals now or out of scope?

test/test_convolve_scanners.py
sbidari: It's not! It was constructed without much thought to replace a lambda expression. 
replaced with `jnp.ones_like` in https://github.com/CDCgov/PyRenew/pull/440/commits/cc61d0b2f3a973109808ffa2c222aaeb319be3c3

pyrenew/latent/infectionswithfeedback.py
```diff
@@ -143,11 +143,18 @@ def sample(
...
+            )
+
+        if I0.ndim != Rt.ndim:
+            raise ValueError(
+                "Initial infections and Rt must have the same dimensions. "
+                f"Got initial infections of shape {I0.shape} "
+                f"and Rt of shape {Rt.shape}."
             )
 
         gen_int_rev = jnp.flip(gen_int)
```
dylanhmorris, line 160: If make multi-dim, specify flip on `axis=0`.

test/test_arrayutils.py
```diff
@@ -8,53 +8,67 @@
...
-
-    x = jnp.array([1, 2])
-    y = jnp.array([1, 2, 3])
+    # Padding along axis 0
+    axis = 0
+    x_pad, y_pad = au.pad_edges_to_match(x, y, axis=axis, pad_direction="end")
 
-    x_pad = au.pad_x_to_match_y(x, y)
+    assert jnp.array_equal(x_pad.shape[axis], y_pad.shape[axis])
+    assert jnp.array_equal(y_pad[-1], y[-1])
```
dylanhmorris, line 69: can we have an assertion on values of `x_pad` as well?

test/test_arrayutils.py
```diff
@@ -8,53 +8,67 @@
...
 
-    assert x_pad.size == 3
+    # padding along axis 1
+    axis = 1
+    x_pad, y_pad = au.pad_edges_to_match(x, y, axis=axis, pad_direction="end")
+    assert jnp.array_equal(x_pad.shape[axis], y_pad.shape[axis])
+    assert jnp.array_equal(y[:, -1], y_pad[:, -1])
```
dylanhmorris, line 74: as above, assertion on the values of `x_pad` as well?

pyrenew/latent/infectionswithfeedback.py
sbidari: I also made checks for shapes of `inf_feedback_strength` and `Rt` independent of other conditions!

test/test_convolve_scanners.py
```diff
@@ -39,10 +52,27 @@ def transform_a(x: any):
...
+            An array of ones with the same shape as the input value `x`.
+        """
+        return jnp.ones_like(x)
+
     scanner_a = pc.new_convolve_scanner(to_scan_a, transform_a)
 
     double_scanner_a = pc.new_double_convolve_scanner(
-        (jnp.array([523, 2, -0.5233]), to_scan_a), (lambda x: 1, transform_a)
+        (jnp.array([523, 2, -0.5233]), to_scan_a),
+        (transform_ones_like, transform_a),
```
dylanhmorris:
```suggestion
        (jnp.ones_like, transform_a),
```

test/test_convolve_scanners.py
dylanhmorris: Agree have suggested below.

dylanhmorris (commented):
This looks generally very good. A very questions about scope that on which I will defer to you and @damonbayer, and a suggestion about dot products that will be useful for future-proofing.

pyrenew/convolve.py
```diff
@@ -80,7 +80,9 @@ def _new_scanner(
         new_val = transform(
             multiplier * jnp.dot(array_to_convolve, history_subset)
```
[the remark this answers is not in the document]
dylanhmorris: See https://github.com/CDCgov/PyRenew/blob/ade84ce180df8fd102306630a97fb1e695e3ea39/pyrenew/process/ar.py#L153

pyrenew/latent/infectionswithfeedback.py
sbidari: I would say out of scope for now as it's not required for site level wastewater model.

test/test_arrayutils.py
sbidari: added in https://github.com/CDCgov/PyRenew/pull/440/commits/0cd56ad169a4d39941a24f5885fc9e34b06bc5f0

test/test_arrayutils.py
sbidari: added in https://github.com/CDCgov/PyRenew/pull/440/commits/0cd56ad169a4d39941a24f5885fc9e34b06bc5f0

docs/source/tutorials/extending_pyrenew.qmd
```diff
@@ -224,11 +224,10 @@ class InfFeedback(RandomVariable):
 
         inf_feedback_strength = jnp.atleast_1d(inf_feedback_strength)
 
-        inf_feedback_strength = au.pad_x_to_match_y(
+        inf_feedback_strength = au.pad_edges_to_match(
```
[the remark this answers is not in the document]
sbidari: changed in https://github.com/CDCgov/PyRenew/pull/440/commits/0cd56ad169a4d39941a24f5885fc9e34b06bc5f0

docs/source/tutorials/extending_pyrenew.qmd
```diff
@@ -224,11 +224,10 @@ class InfFeedback(RandomVariable):
...
             x=inf_feedback_strength,
             y=Rt,
-            fill_value=inf_feedback_strength[0],
-        )
+        )[0]
```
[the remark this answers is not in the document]
sbidari: fixed in https://github.com/CDCgov/PyRenew/pull/440/commits/0cd56ad169a4d39941a24f5885fc9e34b06bc5f0

pyrenew/convolve.py
```diff
@@ -80,7 +80,9 @@ def _new_scanner(
...
         )
-        latest = jnp.hstack([history_subset[1:], new_val])
+        latest = jnp.concatenate(
+            [history_subset[1:], jnp.expand_dims(new_val, axis=0)]
+        )
```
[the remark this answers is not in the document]
sbidari: Yes, assumed default value of `axis = 0` but like the suggestion of making it explicit!

pyrenew/convolve.py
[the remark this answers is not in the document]
sbidari: done in https://github.com/CDCgov/PyRenew/pull/440/commits/0f3e2fda4fde306a5a271e7a30150dbb68dc8df0

pyrenew/convolve.py
```diff
@@ -160,7 +162,9 @@ def _new_scanner(
         m1, m2 = multipliers
         m_net1 = t1(m1 * jnp.dot(arr1, history_subset))
         new_val = t2(m2 * m_net1 * jnp.dot(arr2, history_subset))
-        latest = jnp.hstack([history_subset[1:], new_val])
+        latest = jnp.concatenate(
+            [history_subset[1:], jnp.expand_dims(new_val, axis=0)]
+        )
```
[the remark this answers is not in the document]
sbidari: https://github.com/CDCgov/PyRenew/pull/440/commits/0f3e2fda4fde306a5a271e7a30150dbb68dc8df0

pyrenew/convolve.py
```diff
@@ -160,7 +162,9 @@ def _new_scanner(
         m1, m2 = multipliers
         m_net1 = t1(m1 * jnp.dot(arr1, history_subset))
```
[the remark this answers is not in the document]
sbidari: changed in https://github.com/CDCgov/PyRenew/pull/440/commits/0f3e2fda4fde306a5a271e7a30150dbb68dc8df0

sbidari:
Thanks @dylanhmorris
I think multidimensional `gen_int` is out of scope for this PR (I will make a note in the PR description) and this is ready for final review.

pyrenew/latent/infections.py
```diff
@@ -80,14 +80,21 @@ def sample(
...
                 "Initial infections vector must be at least as long as "
                 "the generation interval. "
-                f"Initial infections vector length: {I0.size}, "
+                f"Initial infections vector length: {I0.shape[0]}, "
                 f"generation interval length: {gen_int.size}."
             )
 
+        if I0.ndim != Rt.ndim:
+            raise ValueError(
+                "Initial infections and Rt must have the same dimensions. "
```
dylanhmorris, line 93:
```suggestion
                "Initial infections and Rt must have the same number of dimensions. "
```

pyrenew/latent/infectionswithfeedback.py
```diff
@@ -143,11 +143,18 @@ def sample(
...
-                f"and {gen_int.size} generation interval."
+                f"generation interval. Got initial infections length {I0.shape[0]}"
+                f"and generation interval length {gen_int.size}."
+            )
+
+        if I0.ndim != Rt.ndim:
+            raise ValueError(
+                "Initial infections and Rt must have the same dimensions. "
+                f"Got initial infections of shape {I0.shape} "
+                f"and Rt of shape {Rt.shape}."
```
dylanhmorris, line 157: Same as above

pyrenew/latent/infections.py
```diff
@@ -80,14 +80,21 @@ def sample(
...
+                f"Got initial infections of shape {I0.shape} "
+                f"and Rt of shape {Rt.shape}."
+            )
```
dylanhmorris, line 96: Error message (about shapes) and check (about dimensional count) are slightly inconsistent.

But is this really the check we want? One could maybe enforce `I0.shape[1:] == Rt.shape[1:]`, but with `I0.shape[0]` and `Rt.shape[0]` potentially different. 

Even that might be too strong: for example, one might imagine applying a single `Rt` timeseries of shape `(20,)` with an `I0` of shape `(10, 5)` (5 different initial timeseries, each of length 10 timesteps).

dylanhmorris (commented):
A key remaining question about the dimensionality check

pyrenew/latent/infections.py
sbidari: The checks on the shapes of `I0` and `Rt` is not required in `infections.py` but helps ensure that `Rt` and `R_adjustment` are compatible for broadcasting in `infectionswithfeedback.py`.

I agree enforcing equal batch dimension with `I0.shape[1:] == Rt.shape[1:]` is a better check.

dylanhmorris: Sounds good. LMK when ready for re-review.

pyrenew/convolve.py
```diff
@@ -78,9 +78,12 @@ def _new_scanner(
         history_subset: ArrayLike, multiplier: float
     ) -> tuple[ArrayLike, float]:  # numpydoc ignore=GL08
         new_val = transform(
-            multiplier * jnp.dot(array_to_convolve, history_subset)
+            multiplier
+            * jnp.einsum("i...,i...->...", array_to_convolve, history_subset)
+        )
+        latest = jnp.concatenate(
+            [history_subset[1:], jnp.expand_dims(new_val, axis=0)], axis=0
```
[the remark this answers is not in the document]
sbidari: changed in https://github.com/CDCgov/PyRenew/pull/440/commits/542ad84548055207f1b825589881ce83c886b077

test/test_infectionwithfeedback_plate_compatibility.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+def test_infections_with_feedback_plate_compatibility():
+    """
+    Test the InfectionsWithFeedback matching the Infections class.
+    """
+    I0 = jnp.array(
+        np.array([0.0, 0.0, 0.0, 0.5, 0.6, 0.7, 0.8] * 5).reshape(-1, 5)
+    )
+    Rt = jnp.ones((10, 5))
```
dylanhmorris, line 22: Can we also test that a scalar `Rt` gets successfully broadcast?

dylanhmorris (changes requested):
One final thing @sbidari, otherwise looks ready!

test/test_infectionwithfeedback_plate_compatibility.py
sbidari: Not sure I understand! 1-dimensional `Rt` does not get successfully broadcast in the `plate` context with multidimensional `I0`.

dylanhmorris: I just tried it out and I see what you mean. It should be doable but we'll need to do a bit more refactor (including in how we broadcast the feedback strength. But that's out of scope for this PR. So I think this can land and we can make a new issue for further generalization / vectorization and associated testing.

dylanhmorris (approved):
LGTM

dylanhmorris:
@damonbayer can you confirm your concerns were resolved?

pyrenew/latent/infectionswithfeedback.py
```diff
@@ -160,19 +167,23 @@ def sample(
                 **kwargs,
             )
         )
+
+        if inf_feedback_strength.ndim == Rt.ndim - 1:
+            inf_feedback_strength = jnp.expand_dims(inf_feedback_strength, 0)
```
damonbayer, line 172: I think this needs to be modified based on @dylanhmorris's [comment](https://github.com/CDCgov/PyRenew/pull/440#discussion_r1759557819)

test/test_infection_and_infectionwithfeedback.py
damonbayer, line 1: Why do we test `InfectionsWithFeedback` here and in `test/test_infectionsrtfeedback.py`?

sbidari: This file includes tests for input shapes for both the classes. I basically followed the same practice as `test/test_infectionsrtfeedback.py` which contained tests for both non-feedback infections and infections-with-feedback.

damonbayer: If both of the files test `infections` and `infectionsWithFeedback`, I think maybe there should just be one test file.

sbidari: combined into one file in https://github.com/CDCgov/PyRenew/pull/440/commits/bafe3b04021d389fe626e972297228cc586e99ce

pyrenew/latent/infectionswithfeedback.py
sbidari: replaced `jnp.expand_dims` with `jnp.newaxis` in https://github.com/CDCgov/PyRenew/pull/440/commits/bafe3b04021d389fe626e972297228cc586e99ce

damonbayer (approved):
Thanks @sbidari!

PR #446, opened 2024-09-16
Vectorized differenced process that passes all relevant existing tests

dylanhmorris:


damonbayer:
> # Out of scope but planned
> * Helper functions for broadcasting / shape checking / etc for multi-dim array dimensions after the first, where the first dimension is time and is scanned along

Is there an issue for this?

dylanhmorris:
Created #451

pyrenew/process/differencedprocess.py
```diff
@@ -193,7 +118,22 @@ def sample(
...
+        if not n_inits == self.differencing_order:
+            raise ValueError(
+                "Must have exactly as many "
+                "initial difference values as "
+                "the differencing order, given "
+                "in the sequence :math:`X(t=0), X^1(t=1),` "
+                "et cetera. "
+                f"Got {n_inits} values "
+                "for a process of order "
+                f"{self.differencing_order}"
```
damonbayer, line 133:
```suggestion
                f"{self.differencing_order}."
```

damonbayer:
Could be good to have a test that makes use of both the vectorized differenced process and the vectorized AR process to make sure they work together. Perhaps as a test for `rtperiodicdiffar`. Could be done here or as part of https://github.com/CDCgov/PyRenew/pull/419

damonbayer (approved):

PR #470, opened 2024-10-03
refactor infectionswithfeedback.py to allow shared infection feedback strength across sites

sbidari:
Used `jnp.broadcast` instead of padding to get desired shape for `infection_feedback_strength`. This allows flexibility of having shared or individual infection feedback strength across different sites/sub populations which the previous version did not.

This change also makes `pad_edges_to_match` function from the `arrayutils` unnecessary but I left it for now.


dylanhmorris (approved):
This is a much better solution, thanks @sbidari!

PR #473, opened 2024-10-03
Use `ruff format` for `qmd`s

damonbayer:
Closes https://github.com/CDCgov/PyRenew/issues/472.

I still need to look into publishing an action.

I also added ruff config options to `pyproject.toml` mimic the pre-commit options.


dylanhmorris (approved):
LGTM!

Issue #479, opened 2025-01-07
Simplify infection initialization API 


dylanhmorris:
in particular, it should be easy to retrieve the number of initialization points.

Right now, you have to do this:

https://github.com/CDCgov/pyrenew-hew/blob/c9e8cacae591bb2d1bd87be7656097bcaa0f0232/pyrenew_hew/pyrenew_hew_model.py#L381

PR #482, opened 2025-02-11
Add stable age distribution initialization

### LINKED ISSUE #481: Add infection initialization at stable age distribution given R

dylanhmorris:
Closes #481 

Also collates all tests for the `math.py` module into one file.

This approximator should plug straight into `IntitializeInfectionsExponentialGrowth`. Deferring building any additional infrastructure to #479


pyrenew/math.py
```diff
@@ -6,12 +6,110 @@
...
+    jnp.ndarray
+        The resulting array ``[1, ..., n]``.
+    """
+    return jnp.arange(1, jnp.size(vec) + 1)
+
+
+def _neg_MGF(r: float, w: ArrayLike) -> float:
+    """
+    Compute the negative moment generating function (MGF)
+    for a given rate ``r`` and weights ``w``.
```
sbidari, line 36: Add the expression for MGF and little more reference, something like [this](https://cdcgov.github.io/Rt-without-renewal/dev/lib/EpiInfModels/public/#EpiAware.EpiInfModels.R_to_r-Union{Tuple{T},%20Tuple{Any,%20Vector{T}}}%20where%20T%3C:AbstractFloat) in the docstring here?

pyrenew/math.py
```diff
@@ -6,12 +6,110 @@
...
+    ----------
+    vec: ArrayLike
+        The template array
+
+    Returns
+    -------
+    jnp.ndarray
+        The resulting array ``[1, ..., n]``.
+    """
+    return jnp.arange(1, jnp.size(vec) + 1)
```
damonbayer, line 30: any concern about `vec` being multidimensional?

pyrenew/math.py
```diff
@@ -6,12 +6,110 @@
...
+    float
+        The approximate value of ``r``.
+    """
+    mean_gi = jnp.dot(G, _positive_ints_like(G))
+    init_r = (R - 1) / (R * mean_gi)
+
+    def _r_next(_iter, r):  # numpydoc ignore=GL08
+        return r - ((R * _neg_MGF(r, G) - 1) / (R * _neg_MGF_del_r(r, G)))
+
+    return fori_loop(1, n_newton_steps, _r_next, init_r)
```
damonbayer, line 110: Why `fori_loop`? Per [jax docs](https://docs.jax.dev/en/latest/_autosummary/jax.lax.fori_loop.html), won't this be implemented as a scan?

damonbayer (approved):
I have some very minor concerns, but I think this is mergable.

pyrenew/math.py
dylanhmorris: Good idea. Addressed in db47c80

pyrenew/math.py
dylanhmorris: Thought fori_loop might be a bit more readable? But I think a direct `scan` is fine. Done in b3b522f

pyrenew/math.py
dylanhmorris: As implemented, the logic here obtains a scalar $r$ from from scalar $\mathcal{R}$, mostly because I didn't want to think about array dimensions and broadcasting. Generalizing it should be relatively straightforward, but may need some tests. I think I'd prefer to spin that off.

PR #483, opened 2025-02-12
Add weekly_to_daily functionality, move time handling functions to their own module

dylanhmorris:
## Summary
- Adds weekly-to-daily value broadcasting functions + documentation and tests
- Moves timeseries handling to its own module
- Tweaks how input/output data first day of week handling (see breaking changes)

## Breaking changes
This will introduce two breaking changes:
-`from pyrenew.convolve import daily_to_weekly` will need to become `from pyrenew.time import daily_to_weekly`. 
- `daily_to_weekly` now defaults `input_data_first_dow` to the specified `week_start_dow` rather than to `0` (Monday) if  `input_data_first_dow` is not specified. I think this is less surprising. The alternative would be to force the user always to specify `input_data_first_dow`.
We could go through a deprecation cycle but I think Pyrenew is a bit young for that. @damonbayer / @sbidari lmk if you disagree.


sbidari:
Why isn't the `weekly_values` adjusted before repeating to create daily values in `weekly_to_daily`?

dylanhmorris:
> Why isn't the weekly_values adjusted before repeating to create daily values in weekly_to_daily?

For `weekly_to_daily`, all weekly values have to be assumed complete. So the key thing is to ensure that any partial starting week is only partially tiled, which the final `[offset: ]` operation accomplishes. If the user wishes to trim days off the ending week, they must do so manually. I suppose we could add an `output_data_last_dow`, but I'm not sure whether or not it would make the API clearer. Thoughts?

sbidari:
The implementation looks okay to me.
However, I am missing the use case of this `weekly_to_daily` function. Why are the weekly values converted to daily by simply repeating instead of for example, diving by number of days in a week?

dylanhmorris:
> The implementation looks okay to me. However, I am missing the use case of this `weekly_to_daily` function. Why are the weekly values converted to daily by simply repeating instead of for example, diving by number of days in a week?

weekly $\mathcal{R}(t)$, for one. This PR is in part motivated by https://github.com/CDCgov/pyrenew-hew/issues/344

sbidari (approved):
Thanks @dylanhmorris for clarification, LGTM!

pyrenew/time.py
```diff
@@ -0,0 +1,258 @@
...
+    validate_dow(input_data_first_dow, "input_data_first_dow")
+    validate_dow(week_start_dow, "week_start_dow")
+
+    offset = (week_start_dow - input_data_first_dow) % 7
+    daily_values = daily_values[offset:]
+
+    if len(daily_values) < 7:
+        raise ValueError("No complete weekly values available")
+
+    weekly_values = jnp.convolve(daily_values, jnp.ones(7), mode="valid")[::7]
```
damonbayer, line 114:
```suggestion
    trimmed = daily_values[:len(daily_values) // 7 * 7]
    weekly_values = trimmed.reshape(-1, 7).sum(axis=1)
```
Seems to be about 5x faster

dylanhmorris: Nice, thanks! As written this function only accepts 1D inputs and will error with a greater than 1D input (because `jnp.convolve` does). The new version also assumes 1D but doesn't verify. We can verify 1D input but even better would be to generalize.

damonbayer:
```suggestion
    n_weeks = daily_values.shape[0] // 7
    trimmed = daily_values[:n_weeks * 7] 
    return trimmed.reshape(n_weeks, 7, *daily_values.shape[1:]).sum(axis=1)
```

This works, assuming the 0th dimension is "time."

dylanhmorris: Thanks! Would you be up for contributing an edit to the tests to check the multi-dim (1st dim time) case?

damonbayer: @dylanhmorris pushed directly to this branch [`7a85518` (#483)](https://github.com/CDCgov/PyRenew/pull/483/commits/7a85518e143c77e584306fa46274fb579b6ed8ed)

dylanhmorris: thanks!

PR #493, opened 2025-03-19
Add offset computation to compute_delay_ascertained_incidence

dylanhmorris:
Helps address https://github.com/CDCgov/pyrenew-hew/issues/392 and will result in cleaner code.
Fully backward compatible for now, but open to the suggestion that it should always return the tuple (i.e. that `return_offset` should be eliminated. 

I doubt we will use `return_padded` frequently in production, but I think it will be useful for writing tests etc


pyrenew/convolve.py
```diff
@@ -181,7 +181,9 @@ def compute_delay_ascertained_incidence(
     latent_incidence: ArrayLike,
     delay_incidence_to_observation_pmf: ArrayLike,
     p_observed_given_incident: ArrayLike = 1,
-) -> ArrayLike:
+    return_padded: bool = False,
+    return_offset: bool = False,
+) -> ArrayLike | tuple[ArrayLike, int]:
```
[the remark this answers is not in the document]
damonbayer: Then remove the `return_offset` argument, and change `return_padded` to `pad`?

test/test_convolve.py
```diff
@@ -212,3 +211,208 @@ def test_double_convolve_scanner(arr1, arr2, history, m1, m2, transforms):
...
+    [
+        [
+            jnp.array([10, 20, 30, 40, 50, 60, 70, 80]),
+            0.25,
+            jnp.array([0.1, 0.2, 0.3, 0.4]),
+            # does not enforce normed PMF
+        ],
+        [
+            jnp.array([10, 0, 0, 0, 0]),
+            2,  # does not actually force p_obs to be proper
```
sbidari, line 231: Is there a reason to not enforce these?

pyrenew/convolve.py
```diff
@@ -190,22 +192,41 @@ def compute_delay_ascertained_incidence(
...
     latent_incidence: ArrayLike
         Incidence values based on the true underlying process.
+
     delay_incidence_to_observation_pmf: ArrayLike
-        Probability mass function of delay interval from incidence to observation,
-        where the :math:`i`\th entry represents a delay of :math:`i`
-        time units, i.e. ``delay_incidence_to_observation_pmf[0]`` represents
+        Probability mass function of delay interval from incidence to
+        observation with support on the interval 0 to the length of the
+        array's first dimensino. The :math:`i`\th entry represents the
```
sbidari, line 208:
```suggestion
        array's first dimension. The :math:`i`\th entry represents the
```

test/test_convolve.py
dylanhmorris: One can imagine observed events for which there is more than one per incident event. But perhaps that's rare enough and likely enough to be unintended / a bug that we should force the user to write their own convolve in that case. @damonbayer thoughts? Similarly, we don't force the delay distribution to be a proper PMF, but perhaps we should.

pyrenew/convolve.py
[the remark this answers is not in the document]
dylanhmorris: I was already leaning that way, so with one more vote I'm inclined to go for it.

test/test_convolve.py
sbidari: Was referring to both `pmf` and `p_obs`. Would lean towards enforcing proper probability and normed pmf rules to avoid bugs.

pyrenew/convolve.py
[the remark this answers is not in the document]
dylanhmorris: Addressed in bfad3bd

test/test_convolve.py
```diff
@@ -212,3 +213,204 @@ def test_double_convolve_scanner(arr1, arr2, history, m1, m2, transforms):
...
+)
+def test_compute_delay_ascertained_incidence(
+    latent_incidence,
+    p_observed_given_incident,
+    unnormed_pmf,
+):
+    """
+    Basic test that compute_delay_ascertained_incidence
+    agrees with a manual reimplementation and returns
+    the offset if and only if it is asked to.
```
sbidari, line 237:
```suggestion
    agrees with a manual reimplementation.
```

test/test_convolve.py
```diff
@@ -212,3 +213,204 @@ def test_double_convolve_scanner(arr1, arr2, history, m1, m2, transforms):
...
+def test_compute_delay_ascertained_incidence_err(
+    latent_incidence,
+    p_observed_given_incident,
+    delay_incidence_to_observation_pmf,
+    error_type,
+    error_match,
+):
+    """
+    Test that compute_delay_ascertained_incidence
+    errors as expected should.
```
sbidari, line 306:
```suggestion
    errors as expected.
```

sbidari (approved):
LGTM! 
Some docstring edits that I think you missed.

PR #502, opened 2025-04-24
Migrate from Poetry to UV

damonbayer:
closes https://github.com/CDCgov/PyRenew/issues/494

### DIFF
--- a/docs/source/developer_documentation.md
+++ b/docs/source/developer_documentation.md
@@ -13,7 +13,7 @@
 
 ## Installation for Developers
 
-- `poetry install --with dev`
+- `uv pip install -r pyproject.toml --extra dev`
 - `pre-commit install`
 
 ## Coding Conventions
--- a/uv.lock
+++ b/uv.lock
@@ -1,0 +1,2157 @@
+version = 1
+revision = 2
+requires-python = ">=3.13"
+
+[[package]]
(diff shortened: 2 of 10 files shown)


uv.lock
dylanhmorris, line 1: I'm personally a weak no on committing lockfiles (note our now-obselete `poetry.lock` was `.gitignore`d) but I am open to counterarguments

dylanhmorris (commented):
Thanks, @damonbayer! Question about committing lockfiles.

damonbayer:
> Thanks, @damonbayer! Question about committing lockfiles.

https://docs.astral.sh/uv/guides/projects/#uvlock

> This file should be checked into version control, allowing for consistent and reproducible installations across machines.

dylanhmorris (approved):
LGTM, thanks!

docs/source/developer_documentation.md
```diff
@@ -13,7 +13,7 @@
 
 ## Installation for Developers
 
-- `poetry install --with dev`
+- `uv pip install -r pyproject.toml --extra dev`
```
natemcintosh, line 16: I notice this command used here and there in this PR. I think it could be replaced by `uv sync --all-groups` which creates the `uv.lock` and syncs the environment with all primary dependencies and all dependency groups.

PR #566, opened 2025-09-05
basic tutorial - align comments and code

cdc-mitzimorris:
For step 3 in the tutorial, the code comment says:   "The random walk on log Rt, with an inferred s.d.", but the transformed variable is constructed with fixed s.d. of 0.025, and the variable `sd_rt` is unused.   The change that makes the most sense to me is to plug in the variable `sd_rt` to the constructor for the DistributionVariable.


damonbayer (approved):
Thanks for the careful read through!

PR #600, opened 2025-09-29
time helper functions for reporting data

cdc-mitzimorris:
Moving functions from package `pyrenew-hew` file `pyrenew_hew_data.py` which are generally useful for all reporting signals.


SamuelBrand1:
Supporting both `np.datetime64` and `datetime.date` date objects seems to be causing alot of boilerplate?

SamuelBrand1 (approved):
This looks like it works!

I think there is some unnecessary repetition e.g. in `get_observation_indices` where I think it would be cleaner to have an `int` pass into a function rather than if `freq == "mmwr_weekly"` then that `int` is 5 etc.

However, maybe lets make that a small refactor issue.

SamuelBrand1:
@cdc-mitzimorris this is good for you to merge now.

PR #644, opened 2025-12-23
Add generic observation processes which combine the convolution with the noise model.

cdc-mitzimorris:
This PR adds work that was done in https://github.com/cdcent/cfa-pyrenew-hierarchical/pull/4   to PyRenew.

It adds the base observation process class, concrete implementations for Count processes and the abstract base class for Measurement processes,  together with unit tests and two new tutorials for count and measurement observation processes respectively.

Once this PR and the work done in https://github.com/cdcent/cfa-pyrenew-hierarchical/pull/5 have been added to PyRenew, subsequent PRs will deprecate unused features and harmonize the documentation and tutorials.

### DIFF
--- a/docs/tutorials/observation_processes_counts.qmd
+++ b/docs/tutorials/observation_processes_counts.qmd
@@ -525,7 +525,7 @@ for conc_val in concentration_values:
             )
 
         # Use relative days
-        for i, admit in enumerate(hosp_temp.observed[day_one:]):
+        for i, admit in enumerate(hosp_temp.observed):
             conc_results.append(
                 {
                     "day": i,


cdc-mitzimorris:
thanks @dylanhmorris  - all suggested changes made.

pyrenew/observation/noise.py
```diff
@@ -0,0 +1,372 @@
...
+"""
+Noise models for observation processes.
+
+Provides composable noise strategies for count and measurement observations,
+separating the noise distribution from the observation structure.
+
+Count Noise
+-----------
+- ``PoissonNoise``: Equidispersed counts (variance = mean). No parameters.
+- ``NegativeBinomialNoise``: Overdispersed counts (variance > mean).
```
dylanhmorris:
```suggestion
- ``NegativeBinomialNoise``: Overdispersed counts relative to Poisson (variance > mean).
```

pyrenew/observation/noise.py
```diff
@@ -0,0 +1,372 @@
...
+import numpyro
+import numpyro.distributions as dist
+from jax.typing import ArrayLike
+
+from pyrenew.metaclass import RandomVariable
+
+_EPSILON = 1e-10
+
+
+class CountNoise(ABC):
```
dylanhmorris: Why are `CountNoise` and `MeasurementNoise` separate abstract base classes and not subclasses of an overall abstract `Noise` class

pyrenew/observation/noise.py
dylanhmorris: - Any reason not to make `noise` its own module rather than a sub-module of observation, even if in practice we'll main use it for observations? @damonbayer thoughts/preferences?

pyrenew/observation/count_observations.py
```diff
@@ -0,0 +1,348 @@
...
+            The string "subpop".
+        """
+        return "subpop"
+
+    def sample(
+        self,
+        infections: ArrayLike,
+        subpop_indices: ArrayLike,
+        times: ArrayLike,
+        counts: ArrayLike | None = None,
```
dylanhmorris: think we should name this `obs` throughout.

pyrenew/observation/base.py
```diff
@@ -0,0 +1,334 @@
...
+        - Wastewater: shedding convolution -> genome scaling -> dilution -> log
+
+        See Also
+        --------
+        sample : Uses this method then applies noise model
+        """
+        pass  # pragma: no cover
+
+    @abstractmethod
+    def sample(self, **kwargs) -> ArrayLike:
```
dylanhmorris: Should have an `obs` kwarg specified to standardize the interface (per examples below).

pyrenew/randomvariable/hierarchical.py
```diff
@@ -0,0 +1,336 @@
...
+        with numpyro.plate(f"n_{self.name}", n_groups):
+            raw_sd = numpyro.sample(
+                f"{self.name}_raw",
+                dist.Gamma(concentration, rate),
+            )
+
+            group_sd = numpyro.deterministic(
+                self.name,
+                jnp.maximum(raw_sd, self.sd_min),
+            )
```
dylanhmorris: Probably simpler and cleaner to Numpyro's truncated distribution API https://num.pyro.ai/en/stable/tutorials/truncated_distributions.html

pyrenew/observation/noise.py
```diff
@@ -0,0 +1,372 @@
...
+            Normal distributed measurements with hierarchical sensor effects.
+            Shape: (n_obs,)
+
+        Raises
+        ------
+        ValueError
+            If sensor_sd samples non-positive values.
+        """
+        sensor_mode = self.sensor_mode_prior_rv.sample(n_groups=n_sensors)
+        sensor_sd = self.sensor_sd_prior_rv.sample(n_groups=n_sensors)
```
dylanhmorris: Why not plate here? Then `sensor_mode_prior_rv` and `sensor_sd_prior_rv` don't need the n_groups interface.

dylanhmorris (changes requested):
Thanks @cdc-mitzimorris! A high-level suggested changes before re-review

- Change "expected" --> "predicted" throughout, as nothing in the design requires or enforces that the prediction be the expected value, though in many concrete cases it is, and that's desirable.
- Throughout, avoid custom names for the observed quantity, if any, to the `sample()` call. Instead use `obs`, for API mirroring of `numpyro.sample()`.
- Revise the measurements tutorial to be stricter about what is specific to the wastewater example and what is general to delay-ascertained measurements.
- Remove the provided helper classes for sampling priors (i.e. `pyrenew/randomvariable/hierarchical.py` from this PR. That will keep it more narrowly focused on observation processes. 


RE: the helper classes for sampling priors: I am open to the possibility that we could want those helper classes in Pyrenew, but as I mention in my line-by-line comments, I think it requires a broader rethink of how we do distributional variables, etc. Would also appreciate thoughts from @damonbayer on that point.

pyrenew/observation/noise.py
damonbayer: I don't have a strong opinion, but lean toward making `noise` its own module.

cdc-mitzimorris:
ready for re-review - high-level concerns addressed:

- changed `expected` to `predicted`
- standardized calls to `sample` w/ arg `obs`
- revised measurement observations tutorial
- removed helper class for hierarchical models; use NumPyro's TruncatedNormalDistribution directly.

docs/tutorials/observation_processes_counts.qmd
```diff
@@ -0,0 +1,667 @@
...
+        y="Hospital Admissions",
+        title="Poisson Noise Model (Variance = Mean)",
+    )
+    + p9.theme_grey()
+    + p9.ylim(0, 105)
+)
+```
+</details>
+
+To see the reduction in noise, it is ncessary to keep the y-axis on the same scale as in the previous plot.
```
damonbayer:
```suggestion
To see the reduction in noise, it is necessary to keep the y-axis on the same scale as in the previous plot.
```

docs/tutorials/observation_processes_counts.qmd
```diff
@@ -63,47 +83,64 @@ In this example, we use fixed parameter values for illustration; in practice, th
 The delay distribution specifies the probability that an infected person is hospitalized $d$ days after infection, conditional on the infection leading to a hospitalization.
 For example, if `hosp_delay_pmf[5] = 0.2`, then 20% of infections that result in hospitalization will appear as hospital admissions 5 days after infection.
 
-We load a delay distribution from PyRenew's datasets:
+We load a delay distribution from PyRenew's datasets which peaks around day 8-9 post-infection, compute summary statistics, and plot it.
```
dylanhmorris:
```suggestion
We load a delay distribution from PyRenew's example datasets, compute summary statistics, and plot it. The distribution peaks around day 8-9 post-infection.
```

docs/tutorials/observation_processes_measurements.qmd
```diff
@@ -14,58 +23,174 @@ import jax.numpy as jnp
...
-| Subpop structure | Optional (`CountsBySubpop`) | Inherent (hierarchical effects) |
+| Output type | Discrete counts | Continuous values |
+| Output space | Linear (expected counts) | Often log-transformed |
+| Noise model | Poisson or Negative Binomial | Normal (often on log scale) |
+| Scaling | Ascertainment rate $\alpha \in [0,1]$ | Domain-specific |
+| Subpop structure | Optional (`CountsBySubpop`) | Inherent (sensor/site effects) |
+
+### The noise model
+
+Measurement data typically exhibits **sensor-level variability**: different instruments, labs, or sampling locations have systematic biases and different precision levels.
```
dylanhmorris:
```suggestion
Measurement data typically exhibits **sensor-level variability**: different instruments, labs, or sampling locations may have systematic biases and different precision levels.
```

pyrenew/observation/noise.py
```diff
@@ -0,0 +1,372 @@
...
+from abc import ABC, abstractmethod
+
+import jax.numpy as jnp
+import numpyro
+import numpyro.distributions as dist
+from jax.typing import ArrayLike
+
+from pyrenew.metaclass import RandomVariable
+
+_EPSILON = 1e-10
```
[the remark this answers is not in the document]
dylanhmorris: Still need to do this.

pyrenew/observation/measurements.py
```diff
@@ -0,0 +1,144 @@
...
+
+        Returns
+        -------
+        ObservationSample
+            Named tuple with `observed` (sampled/conditioned measurements) and
+            `predicted` (predicted values before noise, shape: n_days x n_subpops).
+        """
+        predicted_values = self._predicted_obs(infections)
+
+        self._deterministic("predicted_log_conc", predicted_values)
```
dylanhmorris: Site name should be configurable and/or dynamically generated based on a higher-level configurable name.

dylanhmorris (changes requested):
Thanks @cdc-mitzimorris! A few things to address and then I can re-review.

pyrenew/observation/base.py
```diff
@@ -262,14 +259,36 @@ def _deterministic(self, name: str, value: ArrayLike) -> None:
...
+        suffix : str
+            Suffix for the site name (e.g., "obs").
+
+        Returns
+        -------
+        str
+            Full site name with ``self.name`` prefix.
+            For example, suffix="obs" returns "{name}_obs".
+        """
+        return f"{self.name}_{suffix}"
```
dylanhmorris: Consider using https://num.pyro.ai/en/latest/handlers.html#scope to accomplish this (your call on whether to do now or in a separate PR

test/test_observation_measurements.py
```diff
@@ -284,7 +288,7 @@ def test_predicted_obs_stored(self):
                 )
             ).get_trace()
 
-        assert "predicted_log_conc" in trace
+        assert "test_predicted" in trace
```
dylanhmorris: good to have a test of the programmatic naming. I believe we want one for obs too, no?

dylanhmorris (commented):
Thanks @cdc-mitzimorris! Just a couple remaining questions.

pyrenew/observation/base.py
```diff
@@ -0,0 +1,356 @@
...
+        Parameters
+        ----------
+        suffix : str
+            Suffix for the site name. Will be prefixed with ``self.name``.
+            For example, suffix="predicted" becomes "{name}_predicted".
+        value : ArrayLike
+            Value to track. Can be any shape.
+        """
+        site_name = f"{self.name}_{suffix}"
+        numpyro.deterministic(site_name, value)
```
dylanhmorris: Use the helper function
```suggestion
        numpyro.deterministic(self._sample_site_name(suffix), value)
```

pyrenew/observation/base.py
```diff
@@ -0,0 +1,356 @@
...
+            raise ValueError(f"{param_name} must return non-empty array")
+
+        pmf_sum = jnp.sum(pmf)
+        if not jnp.isclose(pmf_sum, 1.0, atol=atol):
+            raise ValueError(
+                f"{param_name} must sum to 1.0 (±{atol}), got {float(pmf_sum):.6f}"
+            )
+
+        if jnp.any(pmf < 0):
+            raise ValueError(f"{param_name} must have non-negative values")
```
dylanhmorris: Should land this PR and then address, but this duplicates `distutil.validate_discrete_dist_vector`.

That said, I prefer the implementation here, so this should replace that.

We could also (in #645) consider having discrete PMFs as a strong abstraction. We could then require certain RVs to be of class `PMF` and deferring the validation to those RVs' constructors. Given how fundamental discrete PMFs are to discrete-time renewal processes I favor this.

pyrenew/observation/base.py
```diff
@@ -0,0 +1,356 @@
...
+        output. This method returns the index of the first valid day.
+
+        Returns
+        -------
+        int
+            Day index (0-based) of first valid observation.
+            Equal to ``len(pmf) - 1``.
+        """
+        pmf = self.temporal_pmf_rv()
+        return int(len(pmf) - 1)
```
dylanhmorris: Not used anywhere in the class. Would be better to implement in terms of a PMF abstraction (could then also apply to computing the offset in `compute_delay_ascertained_incidence`. I think refrain from implementing for now.
```suggestion
```

pyrenew/observation/count_observations.py
```diff
@@ -0,0 +1,314 @@
...
+        is_1d = infections.ndim == 1
+        if is_1d:
+            infections = infections[:, jnp.newaxis]
+
+        def convolve_col(col):  # numpydoc ignore=GL08
+            return self._convolve_with_alignment(col, delay_pmf, ascertainment_rate)[0]
+
+        predicted_counts = jax.vmap(convolve_col, in_axes=1, out_axes=1)(infections)
+
+        return predicted_counts[:, 0] if is_1d else predicted_counts
```
dylanhmorris: If there's going to be a switch for 1d input, I think it's cleaner and clearer to `vmap` only in the >2d case, and not pad the input only to unpad the output
```suggestion
        is_1d = infections.ndim == 1
        if is_1d:
            predicted_counts = self._convolve_with_alignment(infections, delay_pmf, ascertainment_rate)[0]
        else: 
            predicted_counts = jax.vmap(lambda col: self._convolve_with_alignment(col, delay_pmf, ascertainment_rate)[0], in_axes=1, out_axes=1)(infections)
        return predicted_counts
```

pyrenew/observation/base.py
cdc-mitzimorris: noted.  marking as resolved.

pyrenew/observation/count_observations.py
```diff
@@ -0,0 +1,314 @@
...
+        Parameters
+        ----------
+        infections : ArrayLike
+            Subpopulation-level infections from the infection process.
+            Shape: (n_days, n_subpops)
+        subpop_indices : ArrayLike
+            Subpopulation index for each observation (0-indexed).
+            Shape: (n_obs,)
+        times : ArrayLike
+            Day index for each observation (0-indexed).
```
dylanhmorris:
```suggestion
            Day index for each observation (0-indexed) relative to the infections vector.
```

test/test_observation_counts.py
```diff
@@ -0,0 +1,564 @@
...
+
+        with numpyro.handlers.seed(rng_seed=42):
+            result = process.sample(
+                infections=infections,
+                obs=None,
+            )
+
+        # Timeline alignment: output length equals input length
+        assert result.observed.shape[0] == len(infections)
+        # First len(delay_pmf)-1 days are NaN (appear as -1 after NegativeBinomial sampling)
```
dylanhmorris:
```suggestion
```

test/test_observation_counts.py
dylanhmorris: Not for this PR: a test of `Counts` sampling with `obs=None` where observations are dense and the delay PMF is of length > 1 would have revealed the issue with NaN padding and -1s. Should add.

dylanhmorris (approved):
Thanks @cdc-mitzimorris! Left suggestions that should fix the tutorial rendering. Apologies for failing to make those along with the test changes in my last review. Please merge when ready!

PR #650, opened 2026-01-15
Model builder, latent infection process components and tutorials

cdc-mitzimorris:
Adds infrastructure for modeling infections across multiple subpopulations with hierarchical Rt structure, plus a builder pattern for composing multi-signal renewal models.                               
                                                                                                                                                                                                             
  Key additions:                                                                                                                                                                                             
  - HierarchicalInfections: Latent process with shared baseline Rt and subpopulation-specific deviations                                                                                                     
  - TemporalProcess protocol with AR1, DifferencedAR1, RandomWalk implementations                                                                                                                            
  - ModelBuilder + MultiSignalModel for composing latent + observation processes                                                                                                                             
  - Hierarchical priors (HierarchicalNormalPrior, GammaGroupSdPrior)                                                                                                                                         
  - Sample datasets (NHSN hospital admissions, fake NWSS wastewater)                                                                                                                                         
  - Two tutorials demonstrating the API                                                                                                                                                                      
                                                                                                                                                                                                             
  ---                                                                                                                                                                                                        
  Architecture                                                                                                                                                                                               
                                                                                                                                                                                                             
  Latent layer:                                                                                                                                                                                              
  - TemporalProcess (protocol)                                                                                                                                                                               
    - AR1, DifferencedAR1, RandomWalk (implementations)                                                                                                                                                      
    - Used by HierarchicalInfections for baseline Rt and subpopulation deviations                                                                                                                            
  - HierarchicalNormalPrior, GammaGroupSdPrior                                                                                                                                                               
    - Used by HierarchicalInfections for I0 and other group-level effects                                                                                                                                    
  - BaseLatentInfectionProcess → HierarchicalInfections                                                                                                                                                      
    - Outputs LatentSample: (aggregate, all_subpops, observed, unobserved)                                                                                                                                   
                                                                                                                                                                                                             
  Model layer:                                                                                                                                                                                               
  - ModelBuilder                                                                                                                                                                                             
    - Configures latent process + observation processes                                                                                                                                                      
    - Computes n_initialization_points from all delay PMFs                                                                                                                                                   
    - Builds MultiSignalModel                                                                                                                                                                                
  - MultiSignalModel                                                                                                                                                                                         
    - Routes infections to observations based on infection_resolution()                                                                                                                                      
    - "aggregate" → summed infections                                                                                                                                                                        
    - "subpop" → per-subpopulation infections                                                                                                                                                                
                                                                                                                                                                                                             
  ---                                                                                                                                                                                                        
  Suggested Review Order                                                                                                                                                                                     
                                                                                                                                                                                                             
  1. latent/temporal_processes.py — TemporalProcess protocol and implementations (AR1, DifferencedAR1, RandomWalk). Module docstring explains relationship to pyrenew.process.                               
  2. latent/base.py — BaseLatentInfectionProcess base class defining the LatentSample output structure                                                                                                       
  3. latent/hierarchical_infections.py — HierarchicalInfections implementation                                                                                                                               
  4. model/multisignal_model.py → model/model_builder.py — Composition layer                                                                                                                                 
  5. Tutorials — Demonstrate the API in action:                                                                                                                                                              
    - latent_hierarchical_infections.qmd: temporal process choices                                                                                                                                           
    - building_multisignal_models.qmd: full inference workflow                                                                                                                                               
  6. Tests and datasets — Can skim; mostly mechanical                                                                                                                                                        
                                                                                                                                                                                                             
  ---                                                                                                                                                                                                        
  Review Scope                                                                                                                                                                                               
                                                                                                                                                                                                             
  Feedback requested on:                                                                                                                                                                                     
  - API design: Are the interfaces intuitive? Would you change any method signatures?                                                                                                                        
  - Correctness: Does the hierarchical Rt decomposition and sum-to-zero constraint make epidemiological sense?                                                                                                                                                                                                                
                                                                                                                                                                                                             
  Deferred to follow-up PRs:                                                                                                                                                                                 
  - Module organization (e.g., whether temporal_processes.py belongs in latent/ vs process/) — tracked in #656, makes sense to address both locations consistently                                           
  - Performance optimization                                                                                                                                                                                 
  - Additional temporal process types (periodic effects, etc.)                                                                                                                                               
                                                                                                                                                                                                             
  ---                                                                                                                                                                                                        
  Known Limitations / Future Work                                                                                                                                                                            
                                                                                                                                                                                                             
  - TemporalProcess implementations don't yet support periodic effects (available in pyrenew.process)                                                                                                        
  - Tested with synthetic data; real-data validation ongoing                                                                                                                                                 
  - ModelBuilder API may evolve based on user feedback

### DIFF
--- a/docs/tutorials/building_multisignal_models.qmd
+++ b/docs/tutorials/building_multisignal_models.qmd
@@ -20,8 +20,8 @@ jupyter:
 ---
 
 ```{python}
-#| label: setup
-#| output: false
+# | label: setup
+# | output: false
 import numpyro
 
 # to run samplers in parallel you must run `set_host_device_count` before importing jax
@@ -30,7 +30,7 @@ numpyro.enable_x64()
 ```
 
 ```{python}
-#| label: imports-base
+# | label: imports-base
 import arviz as az
 import jax
 import jax.numpy as jnp
@@ -52,7 +52,7 @@ def make_rng_key():
 ```
 
 ```{python}
-#| label: imports-pyrenew
+# | label: imports-pyrenew
 from jax.typing import ArrayLike
 
 from pyrenew import datasets
@@ -187,7 +187,7 @@ Latent infection processes implement the renewal equation to generate infection 
 The generation interval PMF specifies the probability that a secondary infection occurs $d$ days after the primary infection.
 
 ```{python}
-#| label: gen-interval
+# | label: gen-interval
 covid_gen_int = [0.16, 0.32, 0.25, 0.14, 0.07, 0.04, 0.02]
 gen_int_pmf = jnp.array(covid_gen_int)
 gen_int_rv = DeterministicPMF("gen_int", gen_int_pmf)
@@ -202,7 +202,7 @@ print(f"Generation interval length: {len(gen_int_pmf)} days")
 The initial infections RV `I0_rv` specifies the **proportion of the population infected** at the first observation time. This must be a value in the interval (0, 1]. We use a Beta prior centered near a small value:
 
 ```{python}
-#| label: initial-infections
+# | label: initial-infections
 I0_rv = DistributionalVariable("I0", dist.Beta(1, 100))
 ```
 
@@ -211,8 +211,10 @@ I0_rv = DistributionalVariable("I0", dist.Beta(1, 100))
 We place a prior on the initial log(Rt), centered at 0.0 (Rt = 1.0) with moderate uncertainty:
 
 ```{python}
-#| label: initial-log-rt
-initial_log_rt_rv = DistributionalVariable("initial_log_rt", dist.Normal(0.0, 0.5))
+# | label: initial-log-rt
+initial_log_rt_rv = DistributionalVariable(
+    "initial_log_rt", dist.Normal(0.0, 0.5)
+)
 ```
 
 ### Temporal Processes for $R(t)$
@@ -225,7 +227,7 @@ We configure two temporal processes:
 The RandomWalk allows flexible evolution of subpopulation-specific transmission without mean reversion.
 
 ```{python}
-#| label: temporal-processes
+# | label: temporal-processes
 # AR1 provides mean-reverting behavior for baseline Rt
 baseline_rt_process = AR1(autoreg=0.9, innovation_sd=0.05)
 
@@ -261,7 +263,7 @@ In this example we use a dataset consisting of hospital admissions for COVID-19 
 for the first 10 months of 2023 (as reported to the CDC).
 
 ```{python}
-#| label: load-hospital-data
+# | label: load-hospital-data
 # Load daily hospital admissions for California
 ca_hosp_data = datasets.load_hospital_data_for_state("CA", "2023-11-06.csv")
 
@@ -281,7 +283,7 @@ print(
 The hospital admissions data is aggregated at the jurisdiction level, therefore we specify a `Counts` observation process.
 
 ```{python}
-#| label: hospital-obs-process
+# | label: hospital-obs-process
 # Infection-to-hospitalization delay (COVID-19, from literature)
 inf_to_hosp_pmf = jnp.array(
     [
@@ -360,7 +362,7 @@ implementations subclass it and implement `_predicted_obs()` to transform infect
 into predicted values. See `observation_processes_measurements.qmd` for a detailed tutorial.
 
 ```{python}
-#| label: wastewater-class
+# | label: wastewater-class
 class Wastewater(Measurements):
     """
     Wastewater viral concentration observation process.
@@ -417,7 +419,7 @@ For the wastewater data, we use a simulated dataset for California with realisti
 that covers the same time period.
 
 ```{python}
-#| label: load-wastewater-data
+# | label: load-wastewater-data
 # Load wastewater data for California
 ca_ww_data = datasets.load_wastewater_data_for_state("CA", "fake_nwss.csv")
 
@@ -447,7 +449,7 @@ Wastewater observations are site-level: each measurement is associated with a sp
 Here we specify HierarchicalNormalPrior for the site-level mode and GammaGroupSdPrior for the standard deviation.
 
 ```{python}
-#| label: wastewater-obs-process
+# | label: wastewater-obs-process
 # Viral shedding kinetics PMF (days post-infection)
 shedding_pmf = jnp.array(
     [
@@ -503,7 +505,7 @@ print(f"  Shedding PMF length: {len(shedding_pmf)} days")
 We instantiate a `PyrenewBuilder` object which handles the composition of the latent infection process and the observation process.
 
 ```{python}
-#| label: model-builder-init
+# | label: model-builder-init
 # Build the multi-signal model
 builder = PyrenewBuilder()
 ```
@@ -521,7 +523,7 @@ Methods `configure_latent` and `add_observation` can be called in any order. Met
 We use `configure_latent` to specify the **model structure**: generation interval, initial infections, and temporal dynamics.
 
 ```{python}
-#| label: configure-latent
+# | label: configure-latent
 print("Latent process configuration:")
 print(f"  Generation interval length: {len(gen_int_rv())} days")
 
@@ -540,7 +542,7 @@ builder.configure_latent(
 Each observation process's `name` attribute becomes the keyword used to pass that observation's data to `model.run()` (e.g., `hospital={...}`, `wastewater={...}`).
 
 ```{python}
-#| label: add-observations-build
+# | label: add-observations-build
 builder.add_observation(hosp_obs)  # Uses hosp_obs.name = "hospital"
 builder.add_observation(ww_obs)  # Uses ww_obs.name = "wastewater"
 model = builder.build()
@@ -662,7 +664,7 @@ samples = model.mcmc.get_samples()
 First we declare the population structure. We have 6 subpopulations, where 5 have wastewater monitoring and 1 does not. The subpopulations with wastewater monitoring need not be contiguous indices—they could be any subset of {0, 1, ..., n_subpops-1}.
 
 ```{python}
-#| label: population-structure
+# | label: population-structure
 # All 6 subpopulations with their population fractions
 subpop_fractions = jnp.array([0.10, 0.14, 0.21, 0.22, 0.07, 0.26])
 
@@ -688,7 +690,7 @@ print(f"Total population: {float(jnp.sum(subpop_fractions)):.0%}")
 We define a function to prepare observation data using the model's helper methods to align with the shared time axis.
 
 ```{python}
-#| label: prepare-observation-data
+# | label: prepare-observation-data
 def prepare_observation_data(
     model, n_days_fit, hosp_admits, ww_data, ww_monitored_subpops
 ):
@@ -755,7 +757,7 @@ Putting this altogether, we align the data with the model time and call `model.r
 We run 4 sampler chains.
 
 ```{python}
-#| label: fit-90-days
+# | label: fit-90-days
 # Clear JAX caches to avoid interference from earlier cells
 jax.clear_caches()
 
@@ -787,7 +789,7 @@ print(f"Elapsed time: {elapsed_90:.1f} seconds")
 ```
 
 ```{python}
-#| label: posterior-shape-90
+# | label: posterior-shape-90
 lat_inf_90 = samples_90["latent_infections"]
 print(f"Posterior samples shape: {lat_inf_90.shape}")
 print(f"Mean latent infections: {np.mean(lat_inf_90):,.0f}")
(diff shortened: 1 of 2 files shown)


dylanhmorris (changes requested):
A couple comments on the handling of #662 while I work on the overall review. @cdc-mitzimorris. All comments apply throughout; I've only listed specific examples.

pyrenew/latent/temporal_processes.py
```diff
@@ -0,0 +1,611 @@
...
+        trajectory = self.process(
+            n=n_timepoints + 1,
+            init_vals=jnp.array([[initial_value]]),
+            autoreg=jnp.array([[self.autoreg]]),
+            noise_sd=innovation_sd,
+            fundamental_process_init_vals=jnp.array([[init_rate_of_change]]),
+            noise_name=f"{name_prefix}_noise",
+        )
+
+        return jnp.squeeze(trajectory[:n_timepoints])
```
dylanhmorris: - Do the inputs definitely need to be 2D (yielding a 2D output that we then squeeze)? I don't think `DifferencedProcess` should require that and if it does it's a bug.
- Why the `n=n_timepoints + 1` followed by downsampling with `trajcetory[:n_timepoints]`?

pyrenew/latent/temporal_processes.py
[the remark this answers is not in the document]
cdc-mitzimorris: see above comment:  https://github.com/CDCgov/PyRenew/pull/650#discussion_r2765805734

baseline Rt is semantically "one trajectory for the jurisdiction," and this duplication is the result of that distinction.  Combining them would either:
  1. Reintroduce expand-then-squeeze (conditional squeeze at the end)
  2. Require always returning 2D, changing the conceptual model

happy to implement whatever would make the most sense to you and @damonbayer

dylanhmorris (changes requested):
Throughout: `numpyro.util.not_jax_tracer()` in place of try/catch validation on traced arrays.

pyrenew/datasets/wastewater_nwss_data/README.md
```diff
@@ -0,0 +1,53 @@
+# Wastewater NWSS Data
+
+## Provenance
+
+- **Source Repository**: [CDC cfa-forecast-renewal-ww](https://github.com/cdcgov/cfa-forecast-renewal-ww)
```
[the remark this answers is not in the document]
cdc-mitzimorris: fixed both

pyrenew/latent/temporal_processes.py
```diff
@@ -0,0 +1,413 @@
...
+            f"{name_prefix}_increments",
+            (increments_raw * self.innovation_sd).T,
+        )
+
+        cumulative = jnp.cumsum(increments, axis=0)
+
+        return jnp.concatenate(
+            [initial_value[jnp.newaxis, :], initial_value + cumulative],
+            axis=0,
+        )
```
dylanhmorris: Use existing `DifferencedProcess` and `IIDRandomSequence` logic internally. To enforce NCP, can use `LocScaleReparam(0)` https://github.com/CDCgov/PyRenew/blob/main/pyrenew/process/randomwalk.py

pyrenew/latent/hierarchical_infections.py
```diff
@@ -0,0 +1,274 @@
...
+        numpyro.deterministic("log_rt_baseline", log_rt_baseline)
+        numpyro.deterministic("rt_baseline", jnp.exp(log_rt_baseline))
+        numpyro.deterministic("rt_subpop", rt_subpop)
+        numpyro.deterministic("subpop_deviations", deviations)
+        numpyro.deterministic("infections_aggregate", infections_aggregate)
+
+        return LatentSample(
+            aggregate=infections_aggregate,
+            all_subpops=infections_all,
+        )
```
dylanhmorris: Return R(t) as well?

pyrenew/model/multisignal_model.py
```diff
@@ -0,0 +1,424 @@
...
+        *,
+        subpop_fractions=None,
+        num_warmup: int = 500,
+        num_samples: int = 500,
+        num_chains: int = 1,
+        rng_key: random.PRNGKey = None,
+        reparam_config: dict[str, Any] | None = None,
+        progress_bar: bool = True,
+        **observation_data,
+    ):
```
dylanhmorris: Given that this inherits from `Model`, just use `Model.run()` rather than creating a `fit` function just for the class. If you have ideas for changes to that API, I think propose them in a separate PR.

pyrenew/observation/noise.py
```diff
@@ -167,17 +180,27 @@ def sample(
...
+                name,
+                dist.Poisson(rate=predicted + _EPSILON),
+                obs=obs,
+            )
+        with numpyro.handlers.mask(mask=mask):
+            return numpyro.sample(
+                name,
+                dist.Poisson(rate=predicted + _EPSILON),
+                obs=obs,
+            )
```
dylanhmorris:
```suggestion
        with numpyro.handlers.mask(mask=True if mask is None else mask):
            return numpyro.sample(
                name,
                dist.Poisson(rate=predicted + _EPSILON),
                obs=obs,
            )
```

pyrenew/observation/measurements.py
dylanhmorris: Not for this PR: consider whether to switch to and/or also use support dense `Measurements` (similar to #662 for counts)

pyrenew/model/multisignal_model.py
```diff
@@ -0,0 +1,424 @@
...
+                if jnp.any(subpop_indices < 0):
+                    raise ValueError(
+                        f"Observation '{name}': subpop_indices cannot be negative"
+                    )
+                max_idx = jnp.max(subpop_indices)
+                if max_idx >= pop.K:
+                    raise ValueError(
+                        f"Observation '{name}': subpop_indices contains "
+                        f"{int(max_idx)} >= {pop.K} (K)"
+                    )
```
dylanhmorris: Not for this PR: might be a nicer pattern to have observation processes provide their own data validation logic and have models call them in this function.

pyrenew/model/pyrenew_builder.py
```diff
@@ -0,0 +1,254 @@
...
+        # Add lookback from each observation process
+        for name, obs_process in self.observations.items():
+            try:
+                lookback = obs_process.lookback_days()
+                lookbacks.append(lookback)
+            except (NotImplementedError, AttributeError) as e:
+                raise ValueError(
+                    f"Observation process '{name}' must implement "
+                    f"lookback_days(). Error: {e}"
+                ) from e
```
dylanhmorris: I prefer an explicit check and error to a try-catch.

pyrenew/model/pyrenew_builder.py
```diff
@@ -0,0 +1,254 @@
...
+        """
+        if self.latent_class is None:
+            raise ValueError("Must call configure_latent() before build()")
+
+        # Compute n_initialization_points
+        n_init = self.compute_n_initialization_points()
+
+        # Construct latent process with computed n_initialization_points
+        latent_params = self.latent_params.copy()
+        latent_params["n_initialization_points"] = n_init
```
dylanhmorris:
```suggestion
        latent_params = {**self.latent_params, "n_initialization_points": n_init}
```

pyrenew/model/pyrenew_builder.py
```diff
@@ -0,0 +1,254 @@
...
+
+        try:
+            latent_process = self.latent_class(**latent_params)
+        except TypeError as e:
```
dylanhmorris: Why specifically a `TypeError`?

dylanhmorris (changes requested):
Thanks @cdc-mitzimorris. Have now reviewed everything except the tutorials and tests. Flagged a bunch of small things and tried to be clear where changes are optional. My main big questions (open to pushback on both):

- Not sure specific distributional priors are needed / in scope. Currently prefer to show the user how to create a needed prior via `VectorizedRV` in the tutorials.
- TemporalProcess `RandomWalk` should wrap the process module's random walk rather than reimplementing.

docs/tutorials/building_multisignal_models.qmd
```diff
@@ -0,0 +1,1126 @@
...
+```
+
+### Initial Log Rt
+
+We set the initial log(Rt) to 0.0, meaning Rt starts at 1.0.
+
+```{python}
+# | label: initial-log-rt
+initial_log_rt_rv = DeterministicVariable("initial_log_rt", 0.0)
+```
```
dylanhmorris: Since we're fitting to data, imo there's no reason for these to be deterministic (and it will prevent `NaN` $\hat{R}$ values in the diagnostic table, which could trip up the reader even though they're valid and not a problem in this case).

dylanhmorris (commented):
Have now reviewed everything but tests. @cdc-mitzimorris I propose you make and/or push back all proposed changes and then I will re-review. I'd like to wait to review tests since some changes could require them to be reworked or all some to be eliminated (e.g. the hierarchical priors).

pyrenew/latent/hierarchical_priors.py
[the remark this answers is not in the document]
cdc-mitzimorris: These are thin wrappers, but I think they are needed.   Our target users aren't necessarily familiar with numpyro internals — getting plate naming, parameterization (e.g., Gamma mean/concentration → rate), and lower bounds right is exactly the kind of thing that's easy to mess up. Having tested, documented convenience classes for the most common hierarchical prior patterns reduces that surface area for mistakes.                                                                                                                                                                    

On the `VectorizedRV` point: `HierarchicalNormalPrior` and `StudentTGroupModePrior` could plausibly subclass it, but `GammaGroupSdPrior` also has the mean/concentration → rate conversion and the sd_min floor, so they don't all fit that abstraction cleanly.

pyrenew/model/multisignal_model.py
```diff
@@ -0,0 +1,424 @@
...
+        ValueError
+            If an observation doesn't implement infection_resolution()
+        """
+        for name, obs in self.observations.items():
+            try:
+                obs.infection_resolution()
+            except (NotImplementedError, AttributeError):
+                raise ValueError(
+                    f"Observation '{name}' must implement infection_resolution()"
+                )
```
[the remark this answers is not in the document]
cdc-mitzimorris: Changed to hasattr lookup and added early validation that the returned resolution is in the supported set ("aggregate", "subpop"), so invalid observations fail at construction time rather than sample time.

pyrenew/latent/hierarchical_infections.py
```diff
@@ -0,0 +1,274 @@
...
+    automatically from all observation processes.
+    """
+
+    def __init__(
+        self,
+        *,
+        gen_int_rv: RandomVariable,
+        I0_rv: RandomVariable,
+        baseline_temporal: TemporalProcess,
+        subpop_temporal: TemporalProcess,
```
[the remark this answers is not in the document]
cdc-mitzimorris: These are `TemporalProcess` instances, not `RandomVariable`. The type annotation on both is TemporalProcess, and the actual values passed are things like AR1(...),  DifferencedAR1(...), RandomWalk(...).

Renamed as `baseline_rt_process` and `subpop_rt_deviation_process` to highlight Rt.

pyrenew/latent/hierarchical_infections.py
cdc-mitzimorris: Rt values are already recorded as numpyro.deterministic sites (latent_infections/rt_baseline, latent_infections/rt_subpop), so they're available in posterior samples and Predictive output. Adding them to `LatentSample` would change the return type from a 2-tuple to a 3-tuple, breaking the `inf_agg, inf_all = process.sample(...)` unpacking pattern used throughout the codebase and tutorials.

pyrenew/model/pyrenew_builder.py
```diff
@@ -0,0 +1,254 @@
...
+        """
+        name = obs_process.name
+        if name in self.observations:
+            raise ValueError(
+                f"Observation '{name}' already added. "
+                f"Each observation must have a unique name."
+            )
+
+        self.observations[name] = obs_process
+        return self
```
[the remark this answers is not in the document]
cdc-mitzimorris: noted.

pyrenew/latent/hierarchical_infections.py
```diff
@@ -192,16 +190,16 @@ def sample(
...
             name_prefix="log_rt_baseline",
         )
 
-        deviations_raw = self.subpop_temporal.sample(
+        deviations_raw = self.subpop_rt_deviation_process.sample(
             n_timepoints=n_total_days,
-            n_processes=pop.K,
-            initial_value=jnp.zeros(pop.K),
+            n_processes=pop.n_subpops,
+            initial_value=jnp.zeros(pop.n_subpops),
```
dylanhmorris: To fix (not in this PR): initial deviations should be confiurable, not fixed.

dylanhmorris (approved):
Thanks, @cdc-mitzimorris!

cdc-mitzimorris:
thank you @dylanhmorris for your careful and patient review!

Issue #702, opened 2026-02-11
Handling right-truncated observation process count data


cdc-mitzimorris:
### Principle

Right-truncation is a **multiplicative adjustment to predicted counts**, applied after the delay convolution and before the noise model. It belongs in the same pipeline as ascertainment and delay:

```
infections → [delay convolution] → delayed_incidence
delayed_incidence × ascertainment × prop_already_reported → expected_counts
expected_counts → [noise] → observations
```

Convolution must come first. Ascertainment and right-truncation are both element-wise multipliers on the convolved signal, so their relative order is interchangeable. The net effect is to modify the *expected value* of the likelihood for recent time points.

### Interface Change to `Counts`

Add one optional parameter to the `Counts` constructor:

```python
class Counts(_CountBase):
    def __init__(
        self,
        name: str,
        ascertainment_rate_rv: RandomVariable,
        delay_distribution_rv: RandomVariable,
        noise: CountNoise,
        right_truncation_rv: RandomVariable | None = None,  # NEW
    ) -> None: ...
```

`right_truncation_rv` is any `RandomVariable` that samples to a 1D array summing to 1.0 — a reporting delay PMF. When `None`, no truncation adjustment is applied (the current default behavior).

### Data Interface

When right-truncation is configured, `sample()` accepts an optional `right_truncation_offset`:

```python
# At model.run() time:
ed_visits={
    "obs": model.pad_observations(ed_counts),
    "right_truncation_offset": 3,  # data was pulled 3 days after last observation
}
```

The offset is **data, not a model parameter** — it depends on when the data snapshot was taken relative to the reporting period. Setting it to `None` (or omitting it) disables the adjustment (e.g., for forecasting).

### Implementation

The core logic is a pure function:

```python
def compute_prop_already_reported(
    reporting_delay_pmf: ArrayLike,
    n_timepoints: int,
    right_truncation_offset: int,
) -> ArrayLike:
    """
    Compute the proportion of events already reported at each time point.

    Given a reporting delay PMF and the number of days between the last
    observation and the data-pull date, returns a vector where entry t
    is the proportion of events from day t that have been reported.

    Parameters
    ----------
    reporting_delay_pmf : ArrayLike
        PMF of reporting delay. Entry k = P(reported exactly k days
        after event). Must sum to 1.0.
    n_timepoints : int
        Length of the time series.
    right_truncation_offset : int
        Days between the last observation date and the data-pull date.
        Offset 0 means data was pulled on the same day as the last
[...]

PR #705, opened 2026-02-13
Counts observation processes provide right-truncation delay effect

cdc-mitzimorris:
This PR implements the design proposed in issue #702 and adds an tutorial illustrating its use and effect.

### DIFF
--- a/pyrenew/observation/count_observations.py
+++ b/pyrenew/observation/count_observations.py
@@ -168,10 +168,15 @@ class _CountBase(BaseObservationProcess):
         -------
         ArrayLike
             Adjusted predicted counts, same shape as input.
+
+        Notes
+        -----
+        Assumes a single truncation PMF shared across all subpopulations.
+        The 1D proportion array is broadcast to match 2D predicted counts.
         """
-        rt_pmf = self.right_truncation_rv()
+        trunc_pmf = self.right_truncation_rv()
         n_timepoints = predicted.shape[0]
-        delay_support = rt_pmf.shape[0] - right_truncation_offset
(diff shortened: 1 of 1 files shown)


docs/tutorials/right_truncation.qmd
```diff
@@ -0,0 +1,311 @@
...
+
+days_pmf = np.arange(len(reporting_delay_pmf))
+cdf = np.cumsum(np.array(reporting_delay_pmf))
+mean_delay = float(np.sum(days_pmf * reporting_delay_pmf))
+print(f"Reporting delay PMF support: 0 to {len(reporting_delay_pmf) - 1} days")
+print(f"Mean reporting delay: {mean_delay:.1f} days")
+print(f"CDF: {np.round(cdf, 2)}")
+```
+
+```{python}
```
damonbayer: text on these plots is very small

cdc-mitzimorris: translated ggplot2 cowplot theme to plotnine.   matplotlib handles margins differently than base R graphics, so we need to specify margins for all text elements as well.   this `cowplot_theme` definition is quite involved - we should make it its own file and include it in all tutorials.    for now, just updating the right truncation tutorial.

damonbayer: Strict translation isn't necessary, but would be cool! My main point in referencing cowplot is that it uses larger than typical labels, per the author's book, which has an entire chapter called "Use larger axis labels"
https://clauswilke.com/dataviz/small-axis-labels.html

cdc-mitzimorris: book makes some good points.    I have tweaked the theme and filed a separate issue.

PR #733, opened 2026-03-05
Pyrenew class for Infections with susceptible depletion

sbidari:
Adds: 

- `InfectionsWithSusceptibleDepletion` class
- `compute_infections_with_susceptible_depletion()` function called by `InfectionsWithSusceptibleDepletion` class to calculate infections and Rt
- Adds tests for new functions


damonbayer:
I think this contribution needs to be made to the multisignal repo, so that we can test it without rewriting the hew model.

PR #792, opened 2026-04-21
Use Panache for formatting md and qmd

damonbayer:
What do you think of this @CDCgov/cfa-predict-stf ?


damonbayer:
Appears to not work with pre-commit.ci :(

Issue #789, opened 2026-04-20
Handle temporal aggregation in observational data


cdc-mitzimorris:
# End-to-end support for coarser-than-daily count signals

Replaces the original scope of issue #789, which addressed only the observation side. That scope leaves the model incoherent for weekly data (e.g., NHSN hospital admissions) because R(t) remains daily-parametrized while observations are weekly. This revision covers both sides in one PR.

## Problem statement

Each observed count must correspond to exactly one estimated R(t) parameter, so that parameter count tracks signal information.  When a signal is reported on a coarser grid than daily (e.g., weekly MMWR hospital admissions), independently estimating one R(t) per day produces 7x as many free parameters as informative observations and is unidentified.  The latent infection trajectory must be driven by an R(t) that is piecewise-constant at the observation cadence.

Under a daily R(t) prior, weekly observations alone cannot distinguish between 7-tuples of R(t) values that sum to the same weekly total.  The prior does the work of smoothing, and the posterior is sensitive to prior tightness rather than data.  Matching the R(t) parametrization cadence to the observation cadence removes this pathology: information and parameters are in 1:1 correspondence.

## Target use case

Weekly NHSN hospital admissions and daily NSSP ED visits at population level (future: county level) in a single multi-signal model.
The end-to-end flow for the weekly signal is:

- Weekly R(t) sampled from a temporal process (one parameter per observation week), broadcast to daily by repetition.
- Daily renewal equation consumes the daily R(t) and produces a daily infection trajectory.
- Daily predicted admissions computed via ascertainment and delay convolution.
- Daily predicted admissions summed to the weekly observation grid.
- Likelihood scored at weekly scale against weekly observations.

Aggregation must sit inside the likelihood path, not in downstream processing.  It is part of the computational graph that numpyro traces for gradient-based inference, and the noise model needs predicted and observed values on the same aggregated scale.

The goal is to express this flow through the public PyRenew builder API with a clean separation of concerns, so that scale of parametrization, daily renewal mechanics, and aggregation of predictions are independent building blocks.

## Implications for the three components

**Observation processing (`CountObservation`):** Daily predictions from the renewal equation must be summed to the observation cadence before entering the likelihood.

[...]

Issue #223, opened 2024-06-27
Consider using `chex` in validation and/or testing


dylanhmorris:
https://github.com/google-deepmind/chex

Could be helpful for improve, e.g. assertions about shapes:

https://github.com/CDCgov/multisignal-epi-inference/blob/4f203361a75be79c06f990b781194bece48d1f26/model/src/test/test_first_difference_ar.py#L29-L30

AFg6K7h4fhy2:
Beyond its formality and conciseness, it looks like it might be slightly faster, but I don't know if boost is trivial, or even if it's meaningful, given that this will be used in tests. Anyway, nice find.

PR #794, opened 2026-04-21
End-to-end support for coarser-than-daily count signals

cdc-mitzimorris:
## Overview

Adds support for count observations aggregated to a weekly grid while the renewal equation continues to be evaluated daily. Two design pieces:

1. **Weekly observation likelihood path.** Daily predicted counts are summed to the weekly observation grid before the likelihood is scored. Aggregation lives inside the numpyro-traced graph so the noise model and the likelihood operate on the same scale.
2. **Independent parameter cadence.** The temporal process for $\mathcal{R}(t)$ may be parameterized daily or weekly (or any stepwise cadence) and is broadcast to the daily axis. The renewal equation and all observation transforms run daily either way.

## End-to-end flow for a weekly signal

- $\mathcal{R}(t)$ sampled by a temporal process at the chosen cadence and broadcast to the daily model axis if needed.
- Daily renewal equation consumes daily $\mathcal{R}(t)$ and produces the daily infection trajectory.
- Daily predicted counts computed via ascertainment + delay convolution.
- Daily predicted counts summed to the weekly observation grid via `pyrenew.time.daily_to_weekly`.
- Likelihood scored at weekly scale against weekly observations.

## Relationship to pyrenew-hew

The production pyrenew-hew model parameterizes $\mathcal{R}(t)$ weekly and aggregates daily predicted hospital admissions to a weekly grid. This PR brings the same capability into PyRenew while making parameter cadence a user choice rather than a fixed coupling:

- **Match pyrenew-hew today**: wrap the inner temporal process in `StepwiseTemporalProcess(step_size=7, alignment="calendar_week", week_start_dow=...)`.
- **Daily parameterization for the same signals**: drop the wrapper and use the inner process directly.
- **Mixed cadences**: weekly hospital admissions and daily ED visits in one model regardless of the Rt parameter cadence.

Observation cadence and parameter cadence are independent design choices; the builder no longer enforces a pairing rule between them.

## Reviewer guide

Review bottom-up through the dependency chain. Each unit's changes are self-contained.

**1. Synthetic data refresh (120 → 126 days)**
- Files: `datagen_he_CA_126.py`, `synthetic_CA_126/*.csv`, `synthetic_data.py`, `test_datagen_he_CA_126.py`, `test_datasets_synthetic.py`.
- Six extra days yields 18 full weeks. No API change.
- Verify `true_parameters.json` matches the generating process.

**2. Observation base validators — `pyrenew/observation/base.py`**
- New: `_validate_aggregation_params`, `_compute_period_offset`, `_validate_period_end_times`.
- Generalized `_validate_shapes_match` replaces `_validate_obs_times_shape`.
- `_validate_dow_effect` now uses the shared `require_shape` helper.
- Focus: offset arithmetic `(period_end_dow + 1 - first_day_dow) % 7` and period-boundary alignment.

**3. Latent: temporal processes — `pyrenew/latent/temporal_processes.py`**
- New class `StepwiseTemporalProcess`; `step_size` attr added to the `TemporalProcess` Protocol and to AR1 / DifferencedAR1 / RandomWalk.
- Alignment options: `"model_index"` (default) starts blocks at model index 0; `"calendar_week"` aligns weekly blocks to a declared `week_start_dow`.
- Calendar-week broadcast delegates to `pyrenew.time.weekly_to_daily`; coarse trajectory recorded as `{name_prefix}_coarse`.
- `first_day_dow` threaded through the Protocol so calendar-aligned wrappers can use the model-axis day-of-week; standard processes ignore it.

**4. Latent: shape contracts — `pyrenew/latent/{population,subpopulation}_infections.py`**
- Both `sample()` methods accept `first_day_dow` and forward it to their temporal processes.
- Output shape of every temporal process is validated via the shared `pyrenew.arrayutils.require_shape` helper.

**5. Count observation path — `pyrenew/observation/count_observations.py`**
- New ctor params `aggregation_period`, `reporting_schedule`, `period_end_dow`.
- New `_aggregate` wraps `pyrenew.time.daily_to_weekly`.
- Branching in `validate_data` / `sample` for regular vs. irregular schedules.
- Confirm all in-tree `SubpopulationCounts` call sites updated; `times` → `period_end_times` rename consistent.

**6. Measurement observations — `pyrenew/observation/measurement_observations.py`**
- One-line refactor to use the renamed shape validator.

**7. Builder + model coherence — `pyrenew/model/{pyrenew_builder,multisignal_model}.py`**
- `MultiSignalModel.sample()` accepts `first_day_dow` and forwards it to the latent process.
- `_validate_coherence` enforces calendar-anchor and structural coherence:
  - All observations sharing an `aggregation_period > 1` must agree on `period_end_dow`.
  - Every temporal-process `step_size` must be a positive integer.
  - Calendar-week-aligned temporal processes must have `week_start_dow` consistent with weekly `period_end_dow`: `period_end_dow == (week_start_dow + 6) % 7`.
- `first_day_dow` required at `validate_data` when any obs has `aggregation_period > 1`.
- The previous "step_size ≤ finest aggregation_period" rule was deliberately removed — see the in-body comment.

**8. Integration test — `test/integration/test_population_infections_he_weekly.py`**
- End-to-end MCMC: weekly admissions + daily ED visits, calendar-week-aligned weekly Rt.
- Prior-predictive shape check and posterior recovery for I0 + baseline R(t).
- Uses `numpyro.enable_x64()` + `set_host_device_count(4)`.

**9. Test fixtures + config — `test/conftest.py`, `pyproject.toml`, `_typos.toml`**
- 196-line conftest promotes repeated setup to fixtures, including reusable temporal-process stubs (`WrongShapeTemporalProcess`, `ConstantTemporalProcess`, `InvalidStepSizeTemporalProcess`).
- Pytest `integration` marker added so `-m "not integration"` skips MCMC tests.

**10. Tutorial — `docs/tutorials/building_multisignal_models.qmd`**
- New section explaining the three independent choices: parameter cadence, model time axis, observation cadence.
- Worked example pairing weekly Rt (`alignment="calendar_week"`, `week_start_dow=6`) with weekly observations (`period_end_dow=5`).

## Where to focus review attention

- Offset math in `_compute_period_offset` and the period-boundary check in `_validate_period_end_times` — these govern correctness of weekly alignment.
- `StepwiseTemporalProcess` calendar-week broadcasting — `weekly_to_daily` is reused; sanity check with `n_timepoints=17`, `first_day_dow=3`, `week_start_dow=6` produces 3 coarse samples broadcast to `[c0×3, c1×7, c2×7]`.
- The three coherence rules in `_validate_coherence` — each has pass + distinct-failure tests in `test_pyrenew_builder.py`.
- `CountObservation._aggregate` runs inside the numpyro-traced graph — likelihood scoring is at weekly scale, not post-hoc.

## Incidental fixes (not directly tied to #789)

Found while implementing aggregation; small enough that separating them would create churn.

- `_validate_index_array` empty-array guard (`observation/base.py`) — `jnp.any(indices < 0)` on empty arrays returned `False`; now returns early on `size == 0`.
- `_validate_index_array` / `_validate_obs_dense` ndim check — previously accepted non-1D arrays and relied on silent broadcasting; now rejects with a clear error.
- AR(p) stationary-SD bound in `test_ar_process_asymptotics` — the bound `|long_ts[-1]| < 3 * noise_sd` was too tight (stationary SD strictly > innovation SD when autoregressive coefficients are non-zero) and latently flaky; replaced with closed-form stationary SD per order.
- `_validate_obs_times_shape` → `_validate_shapes_match` rename — prerequisite refactor; same shape-match logic needed for `(obs, period_end_times)` pairs.
- `_typos.toml`: allow `dows`. `pyproject.toml`: register `integration` pytest marker.


cdc-mitzimorris:
@damonbayer and @dylanhmorris -  this is ready for code review.   I have added a reviewers guide section to the PR description.

cdc-mitzimorris:
@damonbayer and @dylanhmorris - to match pyrenew-hew this needs one more layer of flexibility. When signals have mixed cadences, for example weekly hospital admissions and daily ED visits, users should be able to choose whether the renewal process is parameterized daily or weekly, while still evaluating the renewal equation on the daily latent time axis.

This version is very close.    Two further changes are required:

- The builder should distinguish parameter cadence from evaluation cadence. The renewal equation and observation transforms are still evaluated daily, but the Rt temporal process may be parameterized daily or coarsely and then broadcast to daily inside the NumPyro graph.

- Stepwise temporal processes should have explicit alignment semantics: either model-index aligned or calendar-week aligned using the shared model axis day-of-week.

cdc-mitzimorris:
@dylanhmorris and @damonbayer - this is ready for review.
there's a lot of changes to look at via the github interface - here's an analysis of how much of this is code changes

### PR Analysis: `mem_789_temporal_aggregation` vs `main`

Excluding `pyrenew/datasets/`. "other" = blank lines and `#` comments.

#### Top-level totals

| area                          | doc+ | doc- | code+ | code- | other+ | other- |
|-------------------------------|-----:|-----:|------:|------:|-------:|-------:|
| **pyrenew/** (sans datasets)  |  644 |   90 |   487 |    67 |    113 |     13 |
| **test/**                     |  713 |   22 | 1,943 |    94 |    473 |      2 |

#### pyrenew/ per file (sorted by code churn)

| file                                              | doc+ | doc- | code+ | code- | other+ | other- |
|---------------------------------------------------|-----:|-----:|------:|------:|-------:|-------:|
| `observation/count_observations.py`               |  283 |   51 |   218 |    52 |     35 |     13 |
| `latent/temporal_processes.py`                    |  114 |    0 |    85 |     1 |     23 |      0 |
| `observation/base.py`                             |   89 |   12 |    76 |    10 |     18 |      0 |
| `model/multisignal_model.py`                      |   46 |   15 |    39 |     3 |      7 |      0 |
| `model/pyrenew_builder.py`                        |   20 |    5 |    37 |     0 |      7 |      0 |
| `time.py`                                         |   54 |    3 |    11 |     0 |     19 |      0 |
| `latent/subpopulation_infections.py`              |   10 |    2 |    10 |     0 |      0 |      0 |
| `latent/population_infections.py`                 |    9 |    2 |     4 |     0 |      0 |      0 |
| `arrayutils.py`                                   |   15 |    0 |     3 |     0 |      4 |      0 |
| `observation/measurement_observations.py`         |    4 |    0 |     2 |     1 |      0 |      0 |
| `latent/__init__.py`                              |    0 |    0 |     2 |     0 |      0 |      0 |

#### test/ per file (sorted by code churn)

| file                                                        | doc+ | doc- | code+ | code- | other+ | other- |
|-------------------------------------------------------------|-----:|-----:|------:|------:|-------:|-------:|
| `test_observation_counts.py`                                |   63 |    0 |   482 |     8 |     87 |      0 |
| `test_pyrenew_builder.py`                                   |   67 |    1 |   312 |    10 |     45 |      0 |
| `test_observation_validation.py`                            |   53 |    9 |   221 |    30 |     54 |      2 |
| `integration/test_population_infections_he_weekly.py`       |  165 |    0 |   223 |     0 |     70 |      0 |
| `integration/test_population_infections_he_weekly_rt.py`    |  128 |    0 |   193 |     0 |     62 |      0 |
| `test_temporal_processes.py`                                |   22 |    0 |   153 |     1 |     33 |      0 |
| `conftest.py`                                               |  142 |    0 |   100 |     0 |     71 |      0 |
| `integration/conftest.py`                                   |   47 |    0 |    93 |     3 |     21 |      0 |
| `test_datagen_he_CA_126.py`                                 |    8 |    6 |    32 |    24 |      5 |      0 |
| `test_population_infections.py`                             |    2 |    0 |    37 |     1 |      6 |      0 |
| `test_subpopulation_infections.py`                          |    2 |    0 |    38 |     0 |      4 |      0 |
| `test_time.py`                                              |    8 |    0 |    32 |     0 |      9 |      0 |
| `integration/test_population_infections_he.py`              |    1 |    1 |     7 |     9 |      1 |      0 |
| `test_ar_process.py`                                        |    0 |    0 |    13 |     1 |      5 |      0 |
| `test_datasets_synthetic.py`                                |    5 |    5 |     7 |     7 |      0 |      0 |

#### Notes on the split

- The code-vs-docstring split uses the Python AST of each file version to identify `Module`/`ClassDef`/`FunctionDef` leading string-literal bodies; every `+`/`-` line is classified by its line number in the respective revision. Triple-quoted strings that aren't docstrings (e.g., multiline literals inside function bodies) count as code.
- Code churn is dominated by `count_observations.py` (the irregular/weekly aggregation landing for #789) and `temporal_processes.py`. `observation/base.py` and `multisignal_model.py` carry the dispatch plumbing for `first_day_dow` / `obs_start_date`. `time.py` and `arrayutils.py` are mostly docstring (new module-level prose around small helpers).

docs/tutorials/building_multisignal_models.qmd
```diff
@@ -237,6 +240,54 @@ baseline_rt_process = AR1(autoreg=0.9, innovation_sd=0.05)
...
 ```
 
+### Choosing the Rt Parameter Cadence
+
+The renewal equation is evaluated on the model's daily time axis, but the temporal process for $\mathcal{R}(t)$ does not have to sample a new parameter every day.
+This separates three model choices:
+
+-   **Parameter cadence**: how often the $\mathcal{R}(t)$ temporal process samples a new latent value
+-   **Model time axis**: the daily axis used by the renewal equation and delay convolutions
+-   **Observation cadence**: the scale of the likelihood for each signal, such as daily ED visits or weekly hospital admissions
```
dylanhmorris:
```suggestion
-   **Observation cadence**: the temporal granularity each signal, such as daily ED visits or weekly hospital admissions
```

docs/tutorials/building_multisignal_models.qmd
```diff
@@ -237,6 +240,54 @@ baseline_rt_process = AR1(autoreg=0.9, innovation_sd=0.05)
...
+
+The AR(1) process above samples one value per model day:
+
+```python
+baseline_rt_process = AR1(autoreg=0.9, innovation_sd=0.05)
+```
+
+To parameterize $\mathcal{R}(t)$ weekly while still running the renewal equation daily, wrap the temporal process in `StepwiseTemporalProcess`.
```
dylanhmorris:
```suggestion
To use a weekly $\mathcal{R}(t)$ while still running the renewal equation daily, wrap the temporal process in `StepwiseTemporalProcess`.
```

docs/tutorials/building_multisignal_models.qmd
```diff
@@ -237,6 +240,54 @@ baseline_rt_process = AR1(autoreg=0.9, innovation_sd=0.05)
...
+weekly_baseline_rt_process = StepwiseTemporalProcess(
+    inner=AR1(autoreg=0.9, innovation_sd=0.05),
+    step_size=7,
+    alignment="calendar_week",
+    week=MMWR_WEEK,
+)
+```
+
+Use `alignment="calendar_week"` when the weekly Rt blocks should align to a calendar week.
+Pass the same `WeekCycle` to every weekly component that must agree on the calendar: the `StepwiseTemporalProcess` above and any weekly `PopulationCounts` or `SubpopulationCounts` observation.
```
dylanhmorris: This is the first time `WeekCycle` is mentioned in the tutorial. Needs to be introduced a bit sooner.

docs/tutorials/building_multisignal_models.qmd
```diff
@@ -237,6 +240,54 @@ baseline_rt_process = AR1(autoreg=0.9, innovation_sd=0.05)
...
+
+```python
+model.run(
+    num_warmup=500,
+    num_samples=1000,
+    rng_key=random.PRNGKey(42),
+    n_days_post_init=n_days_post_init,
+    population_size=population_size,
+    obs_start_date=date(2024, 1, 7),
+
```
dylanhmorris: No actual observations were passed here. Also extra line.

docs/tutorials/building_multisignal_models.qmd
```diff
@@ -658,18 +709,14 @@ model.run(
...
-        "sensor_indices": ww_sensor_indices,
-        "n_sensors": n_ww_sensors,
-    },
+    **obs_data,
 )
 samples = model.mcmc.get_samples()
 ```
 
+where `**obs_data` is a data dictionary which supplies the observation start date as `obs_start_date`  and a data dictionary for each set of signal data, where the name of the data dictionary corresponds to the signal name registered on the builder.
+
```
dylanhmorris, line 721: Write out an should write out the `obs_data` for this example.

dylanhmorris: (now that we no longer have the examples from the keyword `hospital=` and `wastewater=` pattern

pyrenew/latent/base.py
```diff
@@ -338,6 +338,33 @@ def get_required_lookback(self) -> int:
...
+        additional calendar-aligned components.
+
+        Returns
+        -------
+        bool
+            ``True`` if the caller of :meth:`sample` must supply a
+            ``first_day_dow`` (derived from ``obs_start_date`` at the
+            model entry point); ``False`` otherwise.
+        """
+        from pyrenew.latent.temporal_processes import TemporalProcess
```
dylanhmorris: There's no circular import problem here that I can see, so I think this should be at the module level, not within the function.

pyrenew/latent/temporal_processes.py
```diff
@@ -39,14 +39,18 @@
   while stabilizing the growth rate. Wraps [pyrenew.process.DifferencedProcess][].
 - ``RandomWalk``: No mean reversion. Rt can drift without bound.
   Wraps [pyrenew.process.RandomWalk][].
+- ``StepwiseTemporalProcess``: Wrapper that parameterizes any inner
+  ``TemporalProcess`` at a coarser cadence and broadcasts to the per-timepoint
+  scale by repetition. Use to match R(t) parametrization to the coarsest
+  observation cadence.
```
dylanhmorris: Not the only use case, so can be omitted.

```suggestion
  scale by repetition.
```

pyrenew/arrayutils.py
```diff
@@ -24,6 +24,28 @@ def __repr__(self) -> str:
         return f"PeriodicProcessSample(value={self})"
 
 
+def require_shape(arr: ArrayLike, expected: tuple[int, ...], label: str) -> None:
```
dylanhmorris, line 27: If we need this, add a `chex` dep (see #223) and use https://chex.readthedocs.io/en/latest/api.html#chex.assert_shape?

pyrenew/latent/temporal_processes.py
```diff
@@ -66,14 +71,27 @@ class TemporalProcess(Protocol):
...
+    ----------
+    step_size : int
+        Number of consecutive timepoints that share the same sampled value.
+        Defaults to ``1`` for the standard processes (one independent sample
+        per timepoint). Wrapper processes like ``StepwiseTemporalProcess``
+        expose a larger value so that model builders can enforce coherence
+        between R(t) parametrization cadence and observation aggregation.
     """
 
+    step_size: int
```
dylanhmorris, line 87: Given docs above about default. Consider setting to 1 in the base class and then overriding in inherited classes when appropriate? Open to the more explicit approach as well though.

dylanhmorris: @damonbayer thoughts?

pyrenew/latent/temporal_processes.py
```diff
@@ -120,6 +142,8 @@ class AR1(TemporalProcess):
         more volatile trajectories; smaller values produce smoother ones.
     """
 
+    step_size: int = 1
```
dylanhmorris, line 147: See above about potential to avoid with inheritance.

pyrenew/latent/temporal_processes.py
```diff
@@ -399,3 +439,184 @@ def sample(
...
+        arguments are inconsistent.
+    """
+
+    _SUPPORTED_ALIGNMENTS = {"model_index", "calendar_week"}
+
+    def __init__(
+        self,
+        inner: TemporalProcess,
+        step_size: int,
+        alignment: Literal["model_index", "calendar_week"] = "model_index",
```
dylanhmorris: Given the amount of logic below on supported alignments, validation, etc., would alignments benefit from having their own class? @cdc-mitzimorris @damonbayer Not necessarily for this PR.

pyrenew/latent/temporal_processes.py
```diff
@@ -399,3 +439,184 @@ def sample(
...
+            Number of coarse samples required to cover ``n_timepoints`` under
+            the configured alignment.
+        """
+        if self.alignment == "model_index":
+            return (n_timepoints + self.step_size - 1) // self.step_size
+        if first_day_dow is None:
+            raise ValueError(
+                "first_day_dow is required at sample time when "
+                "alignment='calendar_week'"
+            )
```
dylanhmorris, line 624: This feels like another place that deferring logic to an alignment class could simplify code and prevent casework and potentially odd error messages. I realize that right now you only reach this point if `alignment='calendar_week'` but if future changes meant that was not true, you'd end up with an inappropriate error message

docs/tutorials/building_multisignal_models.qmd
cdc-mitzimorris: removing code snippet - arguments to `run` are shown later in the tutorial.

docs/tutorials/building_multisignal_models.qmd
cdc-mitzimorris: introduced.

docs/tutorials/building_multisignal_models.qmd
cdc-mitzimorris: wrote out dict structure, keys only.

pyrenew/arrayutils.py
cdc-mitzimorris: do in another PR?
there are a few things to be ironed out:
- this check should be always on - chex assertions can be disabled.
-  inconsistencies that would be introduced is this one check returned an `AssertionError` where other checks return `ValueError`
-  the function as written provides exact user-facing strings identifying the component that's misconfigured.

pyrenew/model/pyrenew_builder.py
```diff
@@ -195,15 +196,70 @@ def compute_n_initialization_points(self) -> int:
...
+            if (
+                getattr(process, "alignment", None) == "calendar_week"
+                and obs_week is not None
+            ):
+                proc_week = getattr(process, "week", None)
+                if proc_week != obs_week:
+                    raise ValueError(
+                        f"Temporal process '{param_name}' has week={proc_week!r}, "
+                        f"which disagrees with the weekly observation "
+                        f"week={obs_week!r}"
```
dylanhmorris: I do not think we need to or should enforce this.

dylanhmorris:
> All observations sharing an aggregation_period > 1 must agree on period_end_dow.
> Every temporal-process step_size must be a positive integer.
> Calendar-week-aligned temporal processes must have week_start_dow consistent with weekly period_end_dow: period_end_dow == (week_start_dow + 6) % 7.

High-level comment: do we definitely need to enforce all weekly quantities sharing the _same_ week?

I agree that in many cases a user will want this, but it's not a given. For example, you could imagine two weekly aggregate observables, one reported in MMWR epiweeks, the other in isoweeks. Similarly, while matching temporal process weeks to observation weeks makes sense to me as a default, I don't think we should enforce it.
https://github.com/CDCgov/PyRenew/pull/794/changes#r3140086415

pyrenew/latent/temporal_processes.py
cdc-mitzimorris: agreed we can do this either way, but there are a few reasons to keep it as is:

- TemporalProcess is a typing.Protocol, which is meant to describe a structural shape, not to provide default values.
- Yes, the current code has 3 lines of boilerplate but they're in the right place - reading AR1 on its own tells you its step_size is 1.  design precept:  explicit > implicit (PEP 20). 
- If someone wants to add another TemporalProcess, seeing that the existing subclasses declare `step_size` explicitly is a prompt to consider what behavoir applies; otherwise, they might overlook an inappropriate default.

pyrenew/latent/temporal_processes.py
cdc-mitzimorris: agreed, this could be useful at some future point.   keeping for now.

pyrenew/model/pyrenew_builder.py
cdc-mitzimorris: good catch.   agree and gone.

pyrenew/latent/temporal_processes.py
cdc-mitzimorris: rule of 3 - at this point there are 2 alignments, but if we need more, then yes.   keeping for now.

test/test_ar_process.py
```diff
@@ -243,6 +243,23 @@ def test_ar_process_asymptotics(ar_inits, autoreg, noise_sd, n):
         :n
     ]
 
+    # Closed-form stationary standard deviation for the AR(p) process.
+    # The previous assertion used `3 * noise_sd`, which is only correct when
```
damonbayer: What previous assertion?

pyrenew/latent/temporal_processes.py
```diff
@@ -399,3 +438,184 @@ def sample(
...
+    a coarser cadence while downstream model components still need one value
+    per evaluation timepoint. For example, a weekly-parameterized R(t) process
+    can return daily R(t) values for a daily deterministic renewal equation.
+
+    Parameters
+    ----------
+    inner
+        Inner ``TemporalProcess`` that generates the coarse-scale
+        trajectory. Must satisfy the ``TemporalProcess`` protocol.
+    step_size
```
damonbayer, line 461: A few comments here.
1. The interdependencies here feel bad.

`alignment = "calendar_week"` => `step_size = 7` and `week != None`
`alignment != "calendar_week"` => `week = None`

Is there a better way to structure this?

2. Can / should we use "week" instead of "calendar week"?

3. Do we really need non-`None` `week` for `alignment = "calendar_week"`, or could we use the data to inform what the `week` should be?

pyrenew/latent/base.py
```diff
@@ -338,6 +339,31 @@ def get_required_lookback(self) -> int:
...
+            ``first_day_dow`` (derived from ``obs_start_date`` at the
+            model entry point); ``False`` otherwise.
+        """
+        for value in vars(self).values():
+            if (
+                isinstance(value, TemporalProcess)
+                and getattr(value, "alignment", None) == "calendar_week"
+            ):
+                return True
+        return False
```
damonbayer: Cleaner to use a generator expression?
```suggestion
        return any(
            isinstance(value, TemporalProcess)
            and getattr(value, "alignment", None) == "calendar_week"
            for value in vars(self).values()
        )
```

pyrenew/latent/population_infections.py
```diff
@@ -162,6 +168,9 @@ def sample(
         subpop_fractions
             Population fractions. Defaults to ``[1.0]`` (single population).
             Must be ``[1.0]`` if provided.
+        first_day_dow
```
damonbayer, line 171: When does the `first_day` occur in terms of model time? Can we be more clear about that?

test/conftest.py
```diff
@@ -3,8 +3,22 @@
...
+these variables at import time, so they must be set before the first
+``import jax`` anywhere in the test process; placing them at the top of
+this file (loaded by pytest before any test module) satisfies that
+requirement. The ``setdefault`` form respects any value the caller already
+set at the shell level (e.g., a CI with different configuration).
 """
 
+import os
+
+os.environ.setdefault("JAX_ENABLE_X64", "true")
```
damonbayer, line 19: Why?

test/integration/test_population_infections_he_weekly_rt.py
```diff
@@ -0,0 +1,383 @@
...
+        n_compare = min(len(true_rt), len(rt_q05))
+        covered = (true_rt[:n_compare] >= rt_q05[:n_compare]) & (
+            true_rt[:n_compare] <= rt_q95[:n_compare]
+        )
+        coverage = float(np.mean(covered))
+        assert coverage >= 0.80, (
+            f"R(t) 90% CI coverage was {coverage:.1%}, expected >= 80%"
+        )
+
+    def test_ascertainment_rates_recover_order_of_magnitude(
```
damonbayer, line 344: How did you decide which parameters to assess and why do you use different assessments? (95% coverage vs within a factor of 5)

test/integration/test_population_infections_he_weekly_rt.py
```diff
@@ -0,0 +1,383 @@
...
+                pass through unchanged.
+            """
+            if "time" in ds.dims:
+                ds = ds.isel(time=slice(n_init, None))
+                ds = ds.assign_coords(time=range(ds.sizes["time"]))
+            return ds
+
+        return dt.map_over_datasets(trim_init)
+
+    def test_mcmc_convergence(
```
damonbayer, line 267: Seems out of scope for this PR?

pyrenew/latent/temporal_processes.py
cdc-mitzimorris: > Is there a better way to structure this?

yes, you're absolutely right - this is tangled up and should be split into two classes:
- WeeklyTemporalProcess which knows about weeks and week alignment
- StepwiseTemporalProcess which knows about step size and doesn't care about week alignment.

refactored accordingly.

pyrenew/latent/base.py
cdc-mitzimorris: implemented.

test/integration/test_population_infections_he_weekly_rt.py
cdc-mitzimorris: R(t), ihr, and iedr are the main inferred quantities that exercise the weekly-Rt H+E model end to end.
The delay PMFs and ED day-of-week effects are deterministic fixtures in this test, so there is nothing to recover there.
The concentration/noise parameters are nuisance  parameters, so MCMC diagnostics are sufficient.

The assessments differ because the quantities behave differently.
R(t) is a time-varying latent trajectory with many comparable time points, so interval coverage over the trajectory is a useful integration-level check.
The threshold is tolerant because this is just a synthetic dataset and a finite MCMC run, not a simulation-based calibration study.

ihr and iedr are scalar ascertainment rates and are more weakly identified because they trade off against latent infection scale, initialization, delay/noise, and each other.
The factor-of-5 check is a coarse test that the observation scale is in the right range, not a test of precise recovery.

pyrenew/latent/population_infections.py
cdc-mitzimorris: added clarifying comment:

            Day of week for element 0 of the full latent infection time axis,
            including initialization days. When this latent process is used
            inside MultiSignalModel, the caller passes obs_start_date to the
            model and it converts the first observation day to this axis-origin
            day of week by subtracting n_initialization_points

agree that we need to be clearer about temporal indexing.  reviewing language throughout the code base.

cdc-mitzimorris:
@damonbayer and @dylanhmorris - I have addressed all comments.
substantive changes:

- weekly quantities do not need to share the same weekly alignment
- split "StepwiseTemporalProcesses" into "WeeklyTemporalProcesses" and "StepwiseTemporalProcesses"  - the former handles weekly process logic including start dow;  the latter simply handles signals observed at some number of timesteps > 1.

ready for re-review.

pyrenew/latent/temporal_processes.py
```diff
@@ -399,3 +441,240 @@ def sample(
...
+        Must satisfy the ``TemporalProcess`` protocol.
+    week
+        Calendar-week cycle used to align repeated blocks (e.g.,
+        :data:`pyrenew.time.MMWR_WEEK`, :data:`pyrenew.time.ISO_WEEK`).
+    """
+
+    step_size = 7
+    requires_calendar_anchor = True
+
+    def __init__(self, inner: TemporalProcess, week: WeekCycle) -> None:
```
damonbayer: Since initializing `WeekCycle` only takes one arg, should we just take the `start_dow` as an arg in the `WeeklyTemporalProcess` constructor and create the `WeekCycle` internally?

damonbayer (approved):
One small suggestion. Thanks @cdc-mitzimorris!

pyrenew/latent/temporal_processes.py
cdc-mitzimorris: excellent suggestion.   it turns out that with WeeklyTemporalProcess, the WeeklyCycle object isn't doing any real work.  Instead, added integer constants `MMWR_WEEK` and `ISO_WEEK` which can be used to specify `start_dow` - this should help make code more readable.  removed WeeklyCycle object and updated all code and unit tests.

PR #802, opened 2026-04-28
Add Ascertainment model for JointAscertainment

cdc-mitzimorris:
## Summary

This PR adds model-level ascertainment support so multiple observation signals can share structured ascertainment rates instead of each observation independently sampling its own rate.

The main motivation is to support H+E-style multisignal models where hospital admissions and ED visits have distinct but related observation probabilities. The new abstraction lets those probabilities be sampled once at the model level, then passed into observation processes through signal-specific accessors without creating duplicate NumPyro sample sites.

## Additions / Changes

- Added a new `pyrenew.ascertainment` package with:
     + `AscertainmentModel`, the base interface for shared ascertainment models.
     + `AscertainmentSignal`, a signal-specific accessor used by observation processes.
     + `JointAscertainment`, which samples correlated scalar ascertainment rates across signals on the logit scale.
     + A context layer that makes sampled ascertainment values available to observation processes during model execution.

- Extended `PyrenewBuilder` with `add_ascertainment(...)`, allowing shared ascertainment models to be registered alongside latent and observation components.

- Extended `MultiSignalModel` so registered ascertainment models are sampled once before observation processes run. Observation processes then retrieve their signal-specific values from the active ascertainment context.

- Added unit coverage for validation, sampling behavior, context safety, duplicate-site avoidance, builder registration, and direct `MultiSignalModel` validation.

- Added integration coverage for mixed-cadence hospital + ED models:
     + weekly hospital admissions plus daily ED visits with `JointAscertainment`

## Suggested Review Order

1. Start with `pyrenew/ascertainment/base.py` and `pyrenew/ascertainment/context.py` to understand the core abstraction and how accessors retrieve already-sampled values.
2. Review `pyrenew/ascertainment/joint.py`
3. Review the integration points in `pyrenew/model/pyrenew_builder.py` and `pyrenew/model/multisignal_model.py`.
4. Review tutorial `docs/tutorials/ascertainment.py`
5. Review `test/test_ascertainment.py` and the new builder tests in `test/test_pyrenew_builder.py`.
6. Finish with the integration fixtures and tests in `test/integration/conftest.py` and the two new H+E integration test files, since those show the intended end-to-end usage.


pyrenew/ascertainment/joint.py
```diff
@@ -0,0 +1,182 @@
...
+                "Exactly one of scale_tril, covariance_matrix, or "
+                "precision_matrix must be provided."
+            )
+
+        matrix_shape = (n_signals, n_signals)
+        for param_name, param in (
+            ("scale_tril", self.scale_tril),
+            ("covariance_matrix", self.covariance_matrix),
+            ("precision_matrix", self.precision_matrix),
+        ):
```
damonbayer: could be less verbose with

```python
for name in ("scale_tril", "covariance_matrix", "precision_matrix"):
    param = getattr(self, name)
```

pyrenew/ascertainment/joint.py
```diff
@@ -0,0 +1,182 @@
...
+            if param is not None and param.shape != matrix_shape:
+                raise ValueError(
+                    f"{param_name} must have shape {matrix_shape}, "
+                    f"got shape {param.shape}."
+                )
+
+    def _distribution(self) -> dist.MultivariateNormal:
```
damonbayer: Can be more compact:

```python
def _distribution(self) -> dist.MultivariateNormal:
    for name in ("scale_tril", "covariance_matrix", "precision_matrix"):
        value = getattr(self, name)
        if value is not None:
            return dist.MultivariateNormal(loc=self.loc, **{name: value})
    raise ValueError("No valid covariance parameter provided.")
```

If you adopt my other suggestion, we would be re-using`("scale_tril", "covariance_matrix", "precision_matrix")` and could save that list as a variable somewhere.

pyrenew/ascertainment/joint.py
```diff
@@ -0,0 +1,182 @@
...
+            f"{self.name}_eta",
+            self._distribution(),
+        )
+        rates = jnn.sigmoid(eta)
+
+        result = {}
+        for i, signal in enumerate(self.signals):
+            rate = rates[i]
+            numpyro.deterministic(f"{self.name}_{signal}", rate)
+            result[signal] = rate
```
damonbayer: use `zip` instead

```suggestion
        for signal, rate in zip(self.signals, rates):
            numpyro.deterministic(f"{self.name}_{signal}", rate)
            result[signal] = rate
```

pyrenew/ascertainment/timevarying.py
```diff
@@ -0,0 +1,170 @@
...
+        ----------
+        name
+            Name of the ascertainment model.
+        processes
+            Mapping from signal name to temporal process. Each process is
+            sampled on the logit scale and must satisfy the ``TemporalProcess``
+            protocol. Wrappers such as ``WeeklyTemporalProcess`` can be used
+            when the ascertainment rate should be estimated at a coarser
+            cadence and broadcast to the model time axis.
+        locs
```
damonbayer: I'm confused by the use of `loc` throughout this file. Maybe a better name would involve the idea of "signal".

damonbayer: or better yet, refer to logit-scale ascertainment rates.

It would be easier to work with if it accepted natural-scale ascertainment rates, I think. Logit-scale is too inflexible an assumption. The user should be required to pass something on the natural scale.

pyrenew/ascertainment/timevarying.py
```diff
@@ -0,0 +1,170 @@
...
+            Optional mapping from signal name to scalar logit-scale location.
+            Signals not present in ``locs`` use ``0.0``, corresponding to a
+            baseline rate of 0.5 before temporal deviations.
```
damonbayer: I don't think it makes sense to have any default value for this.

damonbayer (changes requested):
I think we should discuss the interface for the time-varying ascertainment rate.

pyrenew/ascertainment/joint.py
cdc-mitzimorris: Added _COVARIANCE_PARAMETER_NAMES and refactored validation plus _distribution() to reuse it. 

Error message is explicit about all accepted parameterizations, not just covariance_matrix.

pyrenew/ascertainment/timevarying.py
cdc-mitzimorris: Agreed. `locs` exposes the logit-scale implementation and hides an unreasonable default.
We could change as follows:
- rename `locs` to `baseline_rates`
- require one natural-scale rate per signal, validate values are in (0, 1), 
- convert to logits internally before adding the temporal deviations.

pyrenew/ascertainment/timevarying.py
cdc-mitzimorris: Agreed - no default baseline. 
Will change to:
- require baseline rate and exact key match with processes
- validate that each baseline rate is a scalar probability on the natural scale.

cdc-mitzimorris:
@damonbayer 

> I think we should discuss the interface for the time-varying ascertainment rate.

Agreed. 

I made the other changes you suggested, but I after further consideration,  I think that `TimeVaryingAscertainment` is problematic in that it gives the model too much freedom.  More importantly,  the current production HE model is doing something different, therefore we need to do something more or different.

I propose the following additional `AscertainmentModel` class:  `JointTimeVaryingAscertainment` which corresponds to the HE model use case:

- hospital and ED baseline ascertainment rates should share information;
- ED ascertainment may vary over time;
- hospital ascertainment can remain scalar;
- hospital IHR should not be tied to the first realized ED ascertainment value.

This preserves the useful scale-sharing idea in the production HEW model while
avoiding the stronger coupling:

```text
IHR = IEDR(0) * ratio
```

Instead, we should: sample correlated baseline ascertainment logits:

```text
mu ~ MultivariateNormal(...)
```

For each signal:

```text
if signal has no temporal process:
    alpha_j = sigmoid(mu_j)

if signal has a temporal process:
    alpha_j(t) = sigmoid(mu_j + z_j(t))
```

For HEW this gives:

```text
IHR = sigmoid(mu_h)

z_ed(t) ~ weekly AR(1)
IEDR(t) = sigmoid(mu_ed + z_ed(t))
```

So hospital and ED baselines are related, but ED temporal deviations do not
force hospital ascertainment to move.


#### Interface Sketch

```python
JointTimeVaryingAscertainment(
    name="he_ascertainment",
    signals=("hospital", "ed_visits"),
    baseline_loc=jnp.array([logit(0.01), logit(0.02)]),
    baseline_scale_tril=jnp.array(
        [
            [0.7, 0.0],
            [0.35, 0.606],
        ]
    ),
    processes={
        "ed_visits": WeeklyTemporalProcess(
            AR1(autoreg=0.8, innovation_sd=0.25),
            start_dow=MMWR_WEEK,
        ),
    },
)
```

Return value:

```python
{
    "hospital": scalar_ihr,
    "ed_visits": iedr_trajectory,
}
```

Signals listed in `signals` but absent from `processes` return scalar rates.
Signals present in `processes` return trajectories on the model time axis.

#### Relationship To Existing Components

```text
JointAscertainment
    Correlated scalar rates only.

TimeVaryingAscertainment
    Independent fixed baselines plus independent temporal deviations.

JointTimeVaryingAscertainment
    Correlated sampled baselines plus optional signal-specific temporal
    deviations.
```

#### Validation

The component should validate:

- `signals` is a non-empty tuple of unique signal names;
- `baseline_loc` has shape `(n_signals,)`;
- exactly one baseline covariance parameter is supplied;
- baseline covariance parameter has shape `(n_signals, n_signals)`;
- every key in `processes` is one of `signals`;
- every process satisfies the `TemporalProcess` protocol.

#### Note

This component keeps PyRenew's current count-observation semantics:
time-varying ascertainment is applied before delay convolution, so trajectories
are interpreted as infection-time ascertainment probabilities.

damonbayer:
> JointAscertainment
>     Correlated scalar rates only.
> 
> TimeVaryingAscertainment
>     Independent fixed baselines plus independent temporal deviations.
> 
> JointTimeVaryingAscertainment
>     Correlated sampled baselines plus optional signal-specific temporal
>     deviations.

Can you be more precise with these descriptions?

What about `optional signal-specific temporal deviations` is optional? Is there a conceptual difference between `independent temporal deviations` and `signal-specific temporal deviations`?

My instinct is that even the `JointTimeVaryingAscertainment` ties the parameters together too weakly, but I'm not totally sure.

> ```
> IHR = sigmoid(mu_h)
> 
> z_ed(t) ~ weekly AR(1)
> IEDR(t) = sigmoid(mu_ed + z_ed(t))
> ```
> 
> 
>     
>   
> 
> So hospital and ED baselines are related, but ED temporal deviations do not
> force hospital ascertainment to move.

I don't see how "hospital and ED baselines are related". Is it because `mu_h` and `mu_ed` have a bivariate normal distribution (not written)?

cdc-mitzimorris:
@damonbayer asks good questions and I think that this PR should drop the TimeVaryingAscertainment component and just do scalar/joint ascertainment.

The current TimeVaryingAscertainment API makes a free temporal process the main abstraction:
```text
logit(rate_j(t)) = logit(baseline_rate_j) + z_j(t)
```
But a free temporal ascertainment process is identifiability-expensive in a renewal model: observed counts identify something close to latent_incidence_t * ascertainment_t, so a flexible z_j(t) can compete directly with latent infections / Rt unless it is strongly motivated and constrained. We also already model day-of-week effects in the observation process, so time-varying ascertainment should not be used to absorb reporting periodicity.

If the use case is for ED visits, it needs justification - do we have any?  Plausible changes in IEDR are not slow latent trends; they are known calendar/event/regime effects such as holidays, disasters, data-feed disruptions, coding changes, or policy changes.   Those should probably be explicit covariates or regime effects, not inferred by a generic AR process that can compete with latent infections/Rt.

cdc-mitzimorris:
@damonbayer - removed TimeVaryingAscertainment (for now).

docs/tutorials/ascertainment.qmd
```diff
@@ -0,0 +1,264 @@
...
+from pyrenew.model import PyrenewBuilder
+from pyrenew.observation import NegativeBinomialNoise, PopulationCounts
+from pyrenew.randomvariable import DistributionalVariable
+from pyrenew.time import MMWR_WEEK
+
+warnings.filterwarnings("ignore", category=PlotnineWarning)
+```
+
+Ascertainment is the probability that a latent infection appears in an observed signal, e.g., a hospitalization or visit to the emergency department.
+For hospital admissions this probability is often called an infection-hospitalization ratio (IHR).
```
dylanhmorris:
```suggestion
For hospital admissions this probability is often called an infection-to-hospitalization rate (IHR).
```

docs/tutorials/ascertainment.qmd
```diff
@@ -0,0 +1,264 @@
...
+For emergency department visits it might be called an infection-ED-visit ratio (IEDR).
```
dylanhmorris:
```suggestion
For emergency department visits it might be called an infection-to-emergency-department rate (IEDR).
```

docs/tutorials/ascertainment.qmd
```diff
@@ -0,0 +1,264 @@
...
+
+```{python}
+# | label: independent-prior-draws
+n_draws = 1500
+key_ihr, key_iedr = random.split(random.PRNGKey(1))
+
+independent_ihr = dist.Beta(2, 198).sample(key_ihr, (n_draws,))
+independent_iedr = dist.Beta(2, 98).sample(key_iedr, (n_draws,))
+
+independent_df = pd.DataFrame(
```
dylanhmorris: Add a comment indicating this will be used for visualization later?

Otherwise a bit unclear what it's doing here.

docs/tutorials/ascertainment.qmd
```diff
@@ -0,0 +1,264 @@
...
+)
+```
+
+## Model-level ascertainment
+
+To share ascertainment structure across count signals, define an ascertainment model once and register it with the builder.
+Each observation process receives the appropriate signal-specific accessor from `for_signal()`.
+
+```python
+builder = PyrenewBuilder()
```
dylanhmorris, line 105: Add an example creation of the object `ascertainment`?

docs/tutorials/ascertainment.qmd
```diff
@@ -0,0 +1,264 @@
...
+)
+builder.add_observation(ed_obs)
+
+model = builder.build()
+```
+
+Signal names in `for_signal()` must match the names used when the ascertainment model was created.
+They do not have to match observation names, but matching them usually makes model code and posterior outputs easier to read.
+
+When running a model that uses weekly observations or day-of-week effects, pass `obs_start_date` so PyRenew can align the model axis to the calendar.
```
dylanhmorris: Not sure we need this line.
```suggestion
```

docs/tutorials/ascertainment.qmd
```diff
@@ -0,0 +1,264 @@
...
+
+## Summary
+
+- Use ordinary `RandomVariable` objects for independent scalar ascertainment rates.
+- Use `JointAscertainment` for correlated scalar rates across signals.
+- Ascertainment and latent infection scale are weakly identified without informative priors or external information.
```
dylanhmorris: This is true but we didn't really demonstrate it above. Remove or justify/discuss more above.

pyrenew/ascertainment/base.py
```diff
@@ -0,0 +1,187 @@
...
+        """
+        if not isinstance(ascertainment_name, str) or len(ascertainment_name) == 0:
+            raise ValueError(
+                "ascertainment_name must be a non-empty string. "
+                f"Got {type(ascertainment_name).__name__}: {ascertainment_name!r}"
+            )
+        if not isinstance(signal_name, str) or len(signal_name) == 0:
+            raise ValueError(
+                "signal_name must be a non-empty string. "
+                f"Got {type(signal_name).__name__}: {signal_name!r}"
```
dylanhmorris, line 52: Separate PR: add a check helper for non-empty string names

pyrenew/ascertainment/joint.py
```diff
@@ -0,0 +1,187 @@
...
+    """
+    Joint prior for scalar ascertainment rates across multiple signals.
+
+    This model is useful when multiple observation streams have distinct but
+    related probabilities of observing latent incidence. For example, hospital
+    admissions and emergency department visits may have different
+    infection-to-observation ratios, while still being correlated because both
+    depend on care-seeking behavior, testing practices, or reporting systems.
+
+    The model samples one multivariate normal vector around natural-scale
```
dylanhmorris:
```suggestion
    The model samples one multivariate logit-normal vector given natural-scale
```

pyrenew/ascertainment/joint.py
```diff
@@ -0,0 +1,187 @@
...
+    baseline ascertainment rates, which are converted to the logit scale
```
dylanhmorris:
```suggestion
    baseline ascertainment rates:
```

pyrenew/ascertainment/joint.py
```diff
@@ -0,0 +1,187 @@
...
+        """
+        super().__init__(name=name, signals=signals)
+        self.baseline_rates = jnp.asarray(baseline_rates)
+        self.scale_tril = self._optional_array(scale_tril)
+        self.covariance_matrix = self._optional_array(covariance_matrix)
+        self.precision_matrix = self._optional_array(precision_matrix)
+        self._validate_parameters()
+        self.baseline_logits = jnp.log(self.baseline_rates) - jnp.log1p(
+            -self.baseline_rates
+        )
```
dylanhmorris:
```suggestion
        self.baseline_logits = logit(self.baseline_rates)
```

pyrenew/ascertainment/joint.py
```diff
@@ -0,0 +1,187 @@
...
+"""
+Joint ascertainment models.
+"""
+
+from __future__ import annotations
+
+from collections.abc import Mapping
+
+import jax.nn as jnn
+import jax.numpy as jnp
```
dylanhmorris, line 10:
```suggestion
import jax.numpy as jnp
from jax.scipy.special import logit, expit
```

pyrenew/ascertainment/joint.py
```diff
@@ -0,0 +1,187 @@
...
+        Returns
+        -------
+        Mapping[str, ArrayLike]
+            Mapping from signal name to sampled scalar ascertainment rate.
+        """
+        eta = numpyro.sample(
+            f"{self.name}_eta",
+            self._distribution(),
+        )
+        rates = jnn.sigmoid(eta)
```
dylanhmorris:
```suggestion
        rates = expit(eta)
```

pyrenew/ascertainment/joint.py
```diff
@@ -0,0 +1,187 @@
+# numpydoc ignore=GL08
+"""
+Joint ascertainment models.
+"""
+
+from __future__ import annotations
+
+from collections.abc import Mapping
+
+import jax.nn as jnn
```
dylanhmorris:
```suggestion
```

pyrenew/ascertainment/joint.py
```diff
@@ -0,0 +1,187 @@
...
+            Exactly one covariance parameter must be supplied.
+        precision_matrix
+            Precision matrix for the multivariate normal on the logit scale.
+            Exactly one covariance parameter must be supplied.
+        """
+        super().__init__(name=name, signals=signals)
+        self.baseline_rates = jnp.asarray(baseline_rates)
+        self.scale_tril = self._optional_array(scale_tril)
+        self.covariance_matrix = self._optional_array(covariance_matrix)
+        self.precision_matrix = self._optional_array(precision_matrix)
```
dylanhmorris: Why not just duck-type this?

pyrenew/ascertainment/joint.py
```diff
@@ -0,0 +1,187 @@
...
+        self._validate_parameters()
+        self.baseline_logits = jnp.log(self.baseline_rates) - jnp.log1p(
+            -self.baseline_rates
+        )
+
+    @staticmethod
+    def _optional_array(value: ArrayLike | None) -> ArrayLike | None:
```
dylanhmorris: See above question about duck typing.

pyrenew/ascertainment/joint.py
```diff
@@ -0,0 +1,187 @@
...
+
+        matrix_shape = (n_signals, n_signals)
+        for name in _COVARIANCE_PARAMETER_NAMES:
+            param = getattr(self, name)
+            if param is not None and param.shape != matrix_shape:
+                raise ValueError(
+                    f"{name} must have shape {matrix_shape}, got shape {param.shape}."
+                )
+
+    def _distribution(self) -> dist.MultivariateNormal:
```
dylanhmorris: Why not instantiate this at construction, without the wrapper?

dylanhmorris: Wouldn't this also enable deferring validation of the various possible covariance args to the `MultivariateNormal` constructor?

dylanhmorris: Notably, it raises exactly the error raised here, e.g.:

```python
ValueError: Exactly one of ['covariance_matrix', 'precision_matrix', 'scale_tril'] must be specified; got ['precision_matrix', 'scale_tril'].
```

pyrenew/ascertainment/base.py
dylanhmorris: Note that there is one already in `context.py`; could be shared throughout codebase.

docs/tutorials/ascertainment.qmd
mitzimorris: moved to where this is being used.

dylanhmorris:
Thanks @cdc-mitzimorris. Before I review further, could you explain the decision to use a context layer here versus any alternative approaches you considered?

docs/tutorials/ascertainment.qmd
mitzimorris: removed.

docs/tutorials/ascertainment.qmd
mitzimorris: removed.

pyrenew/ascertainment/base.py
cdc-mitzimorris: noted.

pyrenew/ascertainment/joint.py
cdc-mitzimorris: I used jnp.asarray to accept duck-typed ArrayLike inputs at the public boundary while normalizing them internally before validation. That lets callers pass lists/NumPy arrays/JAX arrays, but keeps the rest of the class working with predictable JAX arrays.  This conversion makes .shape, validation, and downstream JAX ops consistent.

pyrenew/ascertainment/joint.py
cdc-mitzimorris: Yes, we can improve this.  Changed to keep the public inputs as `ArrayLike` so callers can pass lists, NumPy arrays, or JAX arrays.  Internally we normalize to jax.Array before validation, so updating internal functions and attributes to use `jax.Array | None`.

cdc-mitzimorris:
> Thanks @cdc-mitzimorris. Before I review further, could you explain the decision to use a context layer here versus any alternative approaches you considered?

The main design constraint was that joint ascertainment needs to create one shared NumPyro sample site, while existing observation processes expect signal-specific `RandomVariable`-like objects.

I considered passing the sampled mapping explicitly through the model stack, but that would require broader plumbing/API changes. I also wanted to avoid having each signal accessor sample independently, since that would either duplicate sites or lose the intended joint dependence. Storing sampled values directly on the model object also seemed fragile for NumPyro execution, since model calls/traces should not depend on mutable object state.

The context layer is intended to keep the public observation API unchanged: the joint model samples once, then the signal accessors retrieve their already-sampled values from execution-scoped context without creating additional NumPyro sites. This keeps ownership of the shared sample site centralized in the ascertainment model, while keeping observation processes focused on consuming  signal-specific rates.

pyrenew/ascertainment/context.py
```diff
@@ -0,0 +1,129 @@
...
+            "values were sampled."
+        )
+
+    try:
+        return values[ascertainment_name][signal_name]
+    except KeyError as exc:
+        raise RuntimeError(
+            f"Ascertainment signal {signal_name!r} from model "
+            f"{ascertainment_name!r} is not available in the current context."
+        ) from exc
```
dylanhmorris: Rather than a try/catch, suggest sequential lookups

```suggestion
   ascertainment_model_values = values.get(ascertainment_name, None):
   if ascertainment_model_values is None:
       raise ValueError(
            f"Values for ascertainment model {ascertainment_name!r} are not available in the current context.")
   signal_value = ascertainment_model_values.get(signal_name, None)
   if signal_value is None:
             raise ValueError(
            f"Ascertainment signal {signal_name!r} from model "
            f"{ascertainment_name!r} is not available in the current context.")
   return signal_value
```

pyrenew/ascertainment/joint.py
```diff
@@ -0,0 +1,146 @@
...
+            Exactly one covariance parameter must be supplied.
+        """
+        super().__init__(name=name, signals=signals)
+        self.baseline_rates: Array = jnp.asarray(baseline_rates)
+        self.scale_tril: Array | None = self._optional_array(scale_tril)
+        self.covariance_matrix: Array | None = self._optional_array(covariance_matrix)
+        self.precision_matrix: Array | None = self._optional_array(precision_matrix)
+        self._validate_parameters()
+        self.baseline_logits: Array = logit(self.baseline_rates)
+        self.distribution: dist.MultivariateNormal = dist.MultivariateNormal(
```
dylanhmorris: Why hint here? Not opposed necessarily, just curious.

pyrenew/ascertainment/joint.py
```diff
@@ -0,0 +1,146 @@
...
+        precision_matrix
+            Precision matrix for the multivariate normal on the logit scale.
+            Exactly one covariance parameter must be supplied.
+        """
+        super().__init__(name=name, signals=signals)
+        self.baseline_rates: Array = jnp.asarray(baseline_rates)
+        self.scale_tril: Array | None = self._optional_array(scale_tril)
+        self.covariance_matrix: Array | None = self._optional_array(covariance_matrix)
+        self.precision_matrix: Array | None = self._optional_array(precision_matrix)
+        self._validate_parameters()
```
dylanhmorris: I think I prefer just to pass these straight to the `dist.MultivariateNormal` constructor, and not retain them as class attributes. If you really need them, could always do something like:

```python
@property
def covariance_matrix(self):
    return self.distribution.covariance_matrix

Even for `baseline_rates` I think I'd prefer the following for single-source-of-truth reasons.

```python
@property
def baseline_rates(self):
    return expit(self.distribution.loc)
```

test/test_ascertainment.py
```diff
@@ -0,0 +1,338 @@
...
+        with pytest.raises(RuntimeError, match="boom"):
+            with ascertainment_context(
+                {"he_ascertainment": {"hospital": jnp.array(0.1)}}
+            ):
+                raise RuntimeError("boom")
+
+        with pytest.raises(RuntimeError, match="before ascertainment values"):
+            get_ascertainment_value("he_ascertainment", "hospital")
+
+    def test_missing_context_value_raises_clear_error(self):
```
dylanhmorris: Move this test case above the previous for logical order/readability?

test/test_pyrenew_builder.py
```diff
@@ -305,6 +343,163 @@ def test_prior_predictive_multi_signal(self, simple_builder):
...
+        )
+
+        assert prior_samples["he_ascertainment_eta"].shape == (3, 2)
+        assert prior_samples["he_ascertainment_hospital"].shape == (3,)
+        assert prior_samples["he_ascertainment_ed"].shape == (3,)
+        assert "hospital_predicted" in prior_samples
+        assert "ed_predicted" in prior_samples
+
+        with pytest.raises(RuntimeError, match="before ascertainment values"):
+            ascertainment.for_signal("hospital")()
```
dylanhmorris: Why do we need this check here?

test/test_pyrenew_builder.py
```diff
@@ -501,6 +696,44 @@ def test_validate_method_calls_internal_validate(self, simple_builder):
         # Should not raise
         model.validate()
 
+    def test_sample_rejects_observation_resolution_that_changes_after_validation(
```
dylanhmorris: This test should be simplified or cut I think. Enough moving parts that it's not clear it's successfully simulating the real failure mode (adjusting the observation infection_resolution post-hoc). Also somewhat out of scope for this PR. Make a new issue if you want to add this test?

dylanhmorris (approved):
A few minor suggestions. Happy for you to merge without re-review from me if @damonbayer is satisfied that his requested changes have been made. Thanks, @cdc-mitzimorris!

pyrenew/ascertainment/context.py
cdc-mitzimorris: Separated the checks to return distinct RuntimeError messages for missing model and missing signal cases.

pyrenew/ascertainment/joint.py
cdc-mitzimorris: They document the intended instance attribute types after normalization in `__init__` .  This improves code maintainability, especially for the optional covariance parameterization fields.

pyrenew/ascertainment/joint.py
cdc-mitzimorris: refactored - keep values local, only store `self.distribution`, and expose `baseline_rates` as a derived read-only property from `expit(self.distribution.loc)`.

test/test_pyrenew_builder.py
cdc-mitzimorris: good catch - already tested elsewhere - removed.

test/test_ascertainment.py
cdc-mitzimorris: reordered.

test/test_pyrenew_builder.py
cdc-mitzimorris: removed.

pyrenew/ascertainment/joint.py
```diff
@@ -74,85 +74,62 @@ def __init__(
...
-            self.covariance_matrix,
-            self.precision_matrix,
-        )
-        n_covariance_params = sum(param is not None for param in covariance_params)
-        if n_covariance_params != 1:
+        if jnp.any(baseline_rates <= 0) or jnp.any(baseline_rates >= 1):
             raise ValueError(
-                "Exactly one of scale_tril, covariance_matrix, or "
-                "precision_matrix must be provided."
+                "baseline_rates must contain probabilities in (0, 1), "
```
damonbayer:
```suggestion
                "baseline_rates must contain probabilities in [0, 1], "
```

damonbayer (changes requested):
I would like to review the tutorial, once it can be rendered.

cdc-mitzimorris:
@damonbayer - tutorial renders.

docs/tutorials/ascertainment.qmd
```diff
@@ -0,0 +1,270 @@
...
+    + p9.geom_point(alpha=0.2, size=1.0, color="steelblue")
+    + p9.geom_abline(intercept=0, slope=2, linetype="dashed", color="gray")
+    + p9.facet_wrap("~prior", nrow=1)
+    + p9.labs(
+        x="Hospital ascertainment rate (IHR)",
+        y="ED ascertainment rate (IEDR)",
+        title="Independent vs. joint scalar ascertainment",
+    )
+    + theme_tutorial
+)
```
damonbayer, line 224:
```suggestion
(
    p9.ggplot(compare_df, p9.aes(x="hospital_ihr", y="ed_iedr"))
    + p9.geom_point(alpha=0.2, size=1.0, color="steelblue")
    + p9.facet_wrap("~prior", nrow=1)
    + p9.labs(
        x="Hospital ascertainment rate (IHR)",
        y="ED ascertainment rate (IEDR)",
        title="Independent vs. joint scalar ascertainment",
    )
    + p9.geom_abline(slope=1, intercept=jnp.log10(2), linetype="dashed", color="gray")
    + p9.scale_y_log10()
    + p9.scale_x_log10()
    + theme_tutorial
)
```

Not totally precise, but this does better job at illustrating the difference

Image

damonbayer: Original for reference

Image

damonbayer (approved):
One recommendation

damonbayer:
Expect panache to reformat the tutorial doc.

docs/tutorials/ascertainment.qmd
damonbayer: p9 doesn't work automatically with the logit scale transform :(

cdc-mitzimorris: thanks - good to know!

Issue #787, opened 2026-04-15
Tutorial "Building Multisignal Models" missing discussion of MCMC `model.run()`


cdc-mitzimorris:
Section "Fitting the Model to Data: model.run()" (line 568) states that model.run() takes two types of inputs:

- Observation data
- Population structure

However, the example model.run() calls at lines 653, 784, and 949 also pass:

- num_warmup, num_samples, rng_key — top-level keyword arguments
- mcmc_args={"num_chains": 4, "progress_bar": False} — a dict forwarded to numpyro.infer.MCMC

None of these are documented in the tutorial. A reader who is not already familiar with NumPyro's MCMC API cannot tell:

1. Which keyword arguments to model.run() are PyRenew-defined vs. passed through to NumPyro.
2. What mcmc_args accepts, or why num_chains lives inside that dict while num_warmup does not.
3. Where to look for the full list of valid mcmc_args keys (e.g. chain_method, progress_bar, jit_model_args).

Proposed change

Extend the introductory list in Section "Fitting the Model to Data: model.run()" to a third category, MCMC controls, with a short paragraph covering:

- num_warmup, num_samples, rng_key — top-level arguments consumed directly by model.run().
- mcmc_args — a dict forwarded verbatim to numpyro.infer.MCMC. Common keys: num_chains, chain_method, progress_bar.
- A link to the NumPyro MCMC reference: https://num.pyro.ai/en/stable/mcmc.html

This keeps the tutorial self-contained without duplicating NumPyro's documentation, and removes the implicit assumption that readers already know the NumPyro MCMC API.

dylanhmorris:
Closed by #812

PR #812, opened 2026-05-05
Mem 787 tutorial model run

cdc-mitzimorris:
Revised tutorial `Building Multisignal Models` tutorial to address issue #787.

The section on model fitting is now written from the user perspective.  It explains how the observation data is aligned on the model time axis.   It glosses over the details of the alignment process in order to keep the narrative flow on track, preserving only what's needed in order to follow the call to `model.run()`.


cdc-mitzimorris:
@damonbayer - this should be easy to review - just one file.

cdc-mitzimorris:
@sbidari - this PR is just changing the tutorial on model building - your fresh eyes would be much appreciated!

sbidari:
Thanks for the flag @cdc-mitzimorris, I will review shortly!

docs/tutorials/building_multisignal_models.qmd
```diff
@@ -731,30 +645,35 @@ model.run(
...
+On the model time axis, indices `0` through `n_init - 1` are initialization days.
+The first observed day is model index `n_init`, and the last observed day is model index `n_init + n_days - 1`.
+
+The model provides helper methods to convert observation data onto this model time axis:
+
+- `model.pad_observations(obs)` prepends `n_init` NaN values to dense daily observation vectors.
+- `model.shift_times(times)` adds `n_init` to sparse observation time indices.
+
+The hospital admissions signal is dense: it has one value per observation day, so we pad it.
+The wastewater signal is sparse: each row has an observation day, so we shift its time index.
```
sbidari, line 668: On first read I assumed the model handles this automatically, and it took me a moment to realize the helper `prepare_observation_data` was doing padding/shifting. I think doing this data prep inline instead of a helper function is more illustrative

docs/tutorials/building_multisignal_models.qmd
```diff
@@ -779,9 +698,14 @@ print(
 print(f"Total population: {float(jnp.sum(subpop_fractions)):.0%}")
 ```
 
+The `subpop_fractions` array defines all subpopulations in the latent infection process, and its entries must sum to 1.0.
+Which subpopulations each observation process sees is determined by the observation data.
+For wastewater, `subpop_indices` links each observation to the appropriate subpopulation, while `sensor_indices` selects the sensor-specific noise parameters.
```
sbidari: Can we make this point about `subpop_indices` and `sensor_indices` more prominent, maybe bullets?

docs/tutorials/building_multisignal_models.qmd
```diff
@@ -851,10 +775,50 @@ def prepare_observation_data(
...
+    "obs_start_date": ...,
+    "hospital": {
+        "obs": ...
+    },
+    "wastewater": {
+        "obs": ...
+        "times": ...
+        "subpop_indices": ...
+        "sensor_indices": ...
+        "n_sensors": ...
```
sbidari, line 775: The tutorial does not mention why `n_sensors` is required in the wastewater data dictionary

sbidari (commented):
This tutorial is comprehensive! A main comment I have is around the helper function for data prep. I would prefer inline data prep and model fit to only one dataset (either 90 days or 180 days). Other than that, looks great to me.

docs/tutorials/building_multisignal_models.qmd
cdc-mitzimorris: The helper functions are part of the `MultiSignalModel` class - I could link these to the documentation - would that help?

The goal is to avoid making the user do their own time indexing because it's too easy to mess it up.

sbidari: I see. Then is the reason for creating `prepare_observation_data()` as opposed to something like below to DRY data prep for two n_fit_days?
```
obs_data = {
        "obs_start_date": obs_start_date,
        "hospital": {
            "obs": model.pad_observations(hosp_admits[:n_days_fit]),
        },
        "wastewater": {
            "obs":  ww_data["observed_conc"][ww_mask],
            "times": model.shift_times(ww_data["time_indices"][ww_mask]),
            "subpop_indices": ww_subpop_indices,
            "sensor_indices": ww_data["site_ids"][ww_mask],
            "n_sensors": ww_data["n_sites"],
        },
    }

Comment (sbidari) [reply docs/tutorials/building_multisignal_models.qmd:668]: I am also curious about the reasoning to include/compare 90-day vs 180-day fits in the tutorial for building multi-signal model.

Comment (cdc-mitzimorris) [reply docs/tutorials/building_multisignal_models.qmd:668]: good question.   when I  wrote this tutorial I was wondering why in production the models were run on 90 or 180 days worth of data, so I put this into the tutorial to see what would happen.
you're quite correct that this doesn't have anything to do with building the multisignal model, and it makes the tutorial rather slow to render. 
I could take out the 180 day fit and just leave the 90 day fit if you think its appropriate.

Comment (sbidari) [reply docs/tutorials/building_multisignal_models.qmd:668]: Yes, I like that idea! Just fit to 90 days to keep it more focused

Comment (cdc-mitzimorris) [reply docs/tutorials/building_multisignal_models.qmd:668]: got rid of 180 day fit.  added more informative plot to illustrate uncertainty at the edge of the observation window.

Comment (cdc-mitzimorris) [reply docs/tutorials/building_multisignal_models.qmd:668]: I also revised the data prep section and removed the now-unneeded `prepare_observation_data` and added more discussion of `pad_observations` and `shift_times`.

File: docs/tutorials/building_multisignal_models.qmd (hunk 2 above)
Comment (cdc-mitzimorris) [reply docs/tutorials/building_multisignal_models.qmd]: improved discussion, bulleted `subpop_indices` and `sensor_indices`

File: docs/tutorials/building_multisignal_models.qmd (hunk 3 above)
Comment (cdc-mitzimorris) [reply docs/tutorials/building_multisignal_models.qmd:775]: added to discussion of the wastewater measurement observation paragraph starting on line 709.

**Comment by cdc-mitzimorris** [comment]
@sbidari - ready for re-review.

File: docs/tutorials/building_multisignal_models.qmd [hunk 4]
```diff
@@ -1039,217 +970,95 @@ plot_df_90["signal"] = pd.Categorical(
...
-    )
-    + p9.facet_wrap("~signal", ncol=1, scales="free_y")
-    + p9.scale_y_log10()
     + p9.labs(
         x="Day",
-        y="Count",
-        title="Posterior Latent Infections vs Observed Hospitalizations (180 days)",
+        y="90% interval width",
+        color="Period",
+        title="Posterior uncertainty over the 90-day fit",
```
sbidari:
```suggestion
        title="Posterior uncertainty in latent infections over the 90-day fit",
```

docs/tutorials/building_multisignal_models.qmd
```diff
@@ -1039,217 +970,95 @@ plot_df_90["signal"] = pd.Categorical(
 )
 ```
```
sbidari, line 966: What is the reason for showing 14 days Moving Average of Hospital Admissions instead of daily values?

docs/tutorials/building_multisignal_models.qmd
dylanhmorris: The comment isn't accurate. The print call just prints the length of the PMF. Seems worth fixing here since it's a quick fix. I thnk just remove any output, or just print the generation interval distribution itself b/c it's short.

dylanhmorris:
```suggestion
- **Initial infections ($(I)(0)$)**: Starting condition for the renewal process
```

dylanhmorris, line 221:
```suggestion
### Log $\mathcal{R}(t)$ at time $0$
```

dylanhmorris:
```suggestion
We place a prior on the $\log(\mathcal{R}(t))$ at time $t = 0$, centered at 0.0 ($\mathcal{R}(t=0) = 1$) with moderate uncertainty:
```

dylanhmorris:
```suggestion
### Choosing the \mathcal{R}(t)$ Parameter Cadence
```

docs/tutorials/building_multisignal_models.qmd
```diff
@@ -271,12 +268,12 @@ This separates three model choices:
 - **Model time axis**: the daily axis used by the renewal equation and delay convolutions
 - **Observation cadence**: the temporal granularity for each signal, such as daily ED visits or weekly hospital admissions
 
-The AR(1) process above samples one value per model day:
+The DifferencedAR(1) process above samples one value per model day:
```
dylanhmorris:
```suggestion
The `DifferencedAR1` process above samples one value per model day:
```

dylanhmorris: either  `DifferencedAR1` (the class) or "AR(1) process on the first differences" the concept. I think the former is fine.

docs/tutorials/building_multisignal_models.qmd
dylanhmorris, line 309:
```suggestion
- **NumPyro sites**: Prefixes all sample and deterministic sites (e.g., `hospital_obs`, `hospital_predicted`)
```

dylanhmorris:
```suggestion
The `Wastewater` observation process uses `LogNormalNoise`, which takes hierarchical priors for the site-level mode and standard deviation parameters.
```

dylanhmorris:
```suggestion
Here we specify `HierarchicalNormalPrior` for the site-level mode and `GammaGroupSdPrior` for the standard deviation.
```

dylanhmorris: Apologies for not noticing this sooner. I don't think we need (or want) the key to vary depending on when the tutorial is run, and wrapping this in a function in a collapsed codeblock also makes it harder for an inexperienced user to learn how to pass a key to `model.run()`. I think remove and replace with explicitly specified keys.

docs/tutorials/building_multisignal_models.qmd
```diff
@@ -764,128 +784,39 @@ For this example, such a data dictionary would have the following structure:
 }
 ```
 
-## Running the Model
+At call time, this dictionary is unpacked with `**`, forwarding each entry as a keyword argument to `model.run()`.
```
dylanhmorris:
```suggestion
At call time, we unpack this dictionary with `**`, forwarding each entry as a keyword argument to `model.run()`. Note that we could also have manually passed each dictionary entry to `model.run()`. We chose to create (and then unpack) an `obs_data` dictionary to keep our data organized in one place. 
```

docs/tutorials/building_multisignal_models.qmd
```diff
@@ -1039,217 +970,95 @@ plot_df_90["signal"] = pd.Categorical(
...
+        data=period_summary_df,
+        mapping=p9.aes(
+            x="start",
+            xend="end",
+            y="mean_ci_width",
+            yend="mean_ci_width",
+            color="period",
+        ),
+        size=1.6,
     )
```
dylanhmorris, line 1017: This plot should also have a log scale y axis (consistent with the incidence plot above.

```suggestion
    )
   + p9.scale_y_log10()
```

dylanhmorris (changes requested):
Thanks @cdc-mitzimorris. Read through the rendered version and found some minor editing stuff that isn't actually in the diff, but I think can and should be addressed here. 

One critical change requested: given that the motivation for this PR is to teach users to run models, I think we should remove the `make_rng_key()` function and explicitly create rng keys every time we call `model.run()`.

docs/tutorials/building_multisignal_models.qmd
cdc-mitzimorris: to render  .qmd to html you need a space before a bulleted list.   maybe not for .qmd to .md, but I see no harm in leaving it as is.

docs/tutorials/building_multisignal_models.qmd
cdc-mitzimorris: smoothing.  the daily values are plotted as points - the data is very noisy.  you want to see some kind of trend, and see that the trend lags the infections.

cdc-mitzimorris:
all changes made.

dylanhmorris (approved):
Thanks, @cdc-mitzimorris!

PR #825, opened 2026-05-18
fig-cap as paragraph

damonbayer:
Re: https://github.com/CDCgov/PyRenew/pull/819#discussion_r3250753418

This duplicates the alt-text as a paragraph below each figure. Perhaps it should delete the alt-text, as the fig-cap is not really intended to server the purpose of alt-text.


cdc-mitzimorris (approved):
lgtm

PR #827, opened 2026-05-18
Day-of-week effects applied on observation time axis (i.e., not before time 0)

cdc-mitzimorris:
The renewal process samples infections on a shared daily time axis that begins n_init days before the first observed report. Observations occupy the suffix of this axis; the prefix is initialization padding used by the convolution but never reported. The bug arises because the day-of-week reporting effect, and the masked-likelihood machinery that handles unobserved days, both leak information from the unobserved prefix into the gradient computation that drives MCMC. Two distinct paths are involved, and both must be closed.

Change 1 — `safe_obs` placeholder for masked count observations

In `_score_masked`, replace `safe_obs = jnp.where(jnp.isnan(obs), safe_predicted, obs)` with `safe_obs = jnp.where(jnp.isnan(obs), 0.0, obs)`.

On a padded MultiSignalModel, the observation array is dense across the full time axis, with NaN entries marking unobserved days (the entire initialization prefix, plus any gaps inside the observation window). The likelihood masks those positions so they don't contribute to the model's log-density. But every position still has a log_prob evaluated against the noise distribution; the masked ones are then zeroed out before summing.

The old placeholder used the model's own predicted mean as the dummy observation at masked positions. Predicted means are fractional (e.g., 7.42 reported visits), while count distributions like NegativeBinomial2 are only defined on whole-number counts. Feeding fractional values into a count log_prob returns -inf, which is then zeroed by the mask in the forward direction — but the gradient computation that NUTS needs in order to take its next step propagates that pathology backward, and the parameters feeding the predicted mean (notably any stochastic day-of-week prior) receive undefined gradients. NUTS interprets the resulting trajectory as divergent on every iteration.

`0.0` is a valid whole-number count for every distribution we support, so the masked-position log_prob is now a finite, smoothly-differentiable number. It still doesn't contribute to the likelihood (the mask still zeroes it), but the gradient flowing backward through the mask is now a well-defined zero rather than something undefined.

Change 2 — Apply day-of-week effect to the observation suffix only

`_apply_day_of_week` now multiplies the day-of-week vector against `predicted[first_observed_idx:]`, leaving the initialization prefix untouched. MultiSignalModel computes `first_observed_idx = n_init` and `first_observed_dow = obs_start_date.weekday()` from the user-supplied calendar anchor and threads them through every observation process.

Conceptually, the day-of-week effect is a reporting effect: it scales counts on actual reporting days.  The initialization prefix has no reports, so applying the effect there has no physical meaning.  As implemented, the effect was applied across the entire model time axis, aligned to the day-of-week of element 0 of the padded axis — which works numerically when the effect is a fixed deterministic vector, but causes failures when doing inference.

The mechanism: ascertainment × delay convolution leaves the first few prefix positions of predicted undefined (the convolution has no infection history to draw from yet). Those undefined positions then got multiplied by day-of-week values, contaminating the multiplier itself with the same kind of undefined-gradient pathology described above. Once the multiplier carries a hidden defect, the gradient with respect to the day-of-week prior is undefined, and again NUTS diverges. Restricting the multiplication to the suffix means undefined prefix values never meet the day-of-week parameters, so the gradient stays well-defined.


pyrenew/model/multisignal_model.py
```diff
@@ -186,6 +186,35 @@ def _resolve_first_day_dow(
         n_init = self.latent.n_initialization_points
         return (convert_date(obs_start_date).weekday() - n_init) % 7
```
dylanhmorris, line 187: Can call the new helper function here for DRY
```suggestion
        return (self._resolve_first_observed_dow(obs_start_date) - n_init) % 7
```

pyrenew/observation/count_observations.py
dylanhmorris, line 467:
```suggestion
        MASKED_PLACEHOLDER = 1
        valid_pred = ~jnp.isnan(predicted)
```

pyrenew/observation/count_observations.py
```diff
@@ -462,7 +480,7 @@ def _score_masked(
         safe_predicted = jnp.where(jnp.isnan(predicted), 1.0, predicted)
```
dylanhmorris, line 473:
```suggestion
        safe_predicted = jnp.where(jnp.isnan(predicted), MASKED_PLACEHOLDER, predicted)
```

dylanhmorris: If I understand the issue correctly, we just need to make sure the placeholder observations are all ints (so that we don't get `-inf` log probs, when `NegativeBinomial2(predicted, validate_args=True).log_prob(obs)` is called. Given that, I think we can and should just use the same placeholder (either 0 or 1) for both predictions and observation, and highlight with a variable name that it will get masked out.

pyrenew/observation/count_observations.py
```diff
@@ -301,13 +307,13 @@ def _apply_day_of_week(
...
-            get_sequential_day_of_week_indices(first_day_dow, n_timepoints)
+        n_obs_days = predicted.shape[0] - first_observed_idx
+        obs_effect = dow_effect[
+            get_sequential_day_of_week_indices(first_observed_dow, n_obs_days)
         ]
         if predicted.ndim == 2:
-            daily_effect = daily_effect[:, None]
-        return predicted * daily_effect
+            obs_effect = obs_effect[:, None]
+        return predicted.at[first_observed_idx:].multiply(obs_effect)
```
dylanhmorris: If I'm understanding the bug and the fix correct, the goal here is to avoid multiplying `nan` predictions by an `obs_effect` array. I wonder whether something like this could accomplish the same without needing `first_observed_idx` and `first_observed_dow`. 

```python
        obs_effect = dow_effect[
            get_sequential_day_of_week_indices(first_day_dow, n_timepoints)
        ]
      non_nan_pred = ~jnp.isnan(predicted)
      return predicted.at[non_nan_pred].multiply(obs_effect[non_nan_pred])
```
Could also use `jnp.where` for the last line. Not sure which is more readable/performant/safe.
```Python
        obs_effect = dow_effect[
            get_sequential_day_of_week_indices(first_day_dow, n_timepoints)
        ]
      non_nan_pred = ~jnp.isnan(predicted)
      safe_pred = jnp.where(non_nan_pred, predicted, 0) # "double where trick" # https://docs.jax.dev/en/latest/faq.html#gradients-contain-nan-where-using-where
      return jnp.where(non_nan_pred, safe_pred * obs_effect, predicted)
```

This approach (if it works) would mean that we wouldn't need `_aggregate` and `sample` to take either a `first_day_dow` or a `first_observed{idx, dow}` pair and then check whether the dow effect was defined.

Would it work?

dylanhmorris (changes requested):
Thanks @cdc-mitzimorris. One potential simplification (if it works) that could avoid some branching. LMK what you think. A couple other optional suggestions.

pyrenew/observation/count_observations.py
cdc-mitzimorris: it has to be zero.  if it's 1, that gets multiplied by some fractional pre-observation latent infection which is not an integer and the negative binomial goes boom.

pyrenew/observation/count_observations.py
cdc-mitzimorris: it has to be zero. if it's 1, that gets multiplied by some fractional pre-observation latent infection which is not an integer and the negative binomial goes boom.

pyrenew/observation/count_observations.py
cdc-mitzimorris: yes, this works and is a whole lot simpler.  implemented.

cdc-mitzimorris:
@dylanhmorris  your suggestion is simpler - fixed accordingly.  thanks!

dylanhmorris (approved):
Thanks, @cdc-mitzimorris!

PR #828, opened 2026-05-19
State-Centered Temporal Processes

cdc-mitzimorris:
Added state-centered parameterizations for all three temporal-process
classes in `pyrenew.latent`:

- `AR1` — stationary AR(1) on log-Rt levels
- `DifferencedAR1` — AR(1) on first differences of log-Rt (the production
  process)
- `RandomWalk` — unconstrained drift on log-Rt

Each class now takes a constructor argument `parameterization: Literal["innovation", "state"]`, defaulting to `"innovation"` to preserve current behavior. Setting `"state"` switches the internal sampling from standardized increments to the latent state path directly.

All three state-centered variants are custom `numpyro.distributions.Distribution` subclasses in `pyrenew/latent/state_centered_distributions.py`:

- `StateRandomWalk`: samples the post-initial path as accumulated Normal increments from the deterministic initial value.
- `StateAR1` and `StateDifferencedAR1`: the AR(1) process on the level and on first differences, respectively.

Each has a vectorized log_prob (slice arithmetic, no scan) that carries the MCMC gradient, and a sample method (cumsum for the walk, lax.scan for the AR(1) variants) used only for forward simulation such as prior and posterior predictive, not on the gradient path.

Both parameterizations encode the same prior over the state path; they differ only in sampler geometry, that is, which latent variables HMC operates on.

Code added

| File | Type | Purpose |
|---|---|---|
| `pyrenew/latent/state_centered_distributions.py` | new | `StateAR1`, `StateDifferencedAR1` |
| `pyrenew/latent/temporal_processes.py` | modified | `parameterization` flag on all three classes; `_prepare_initial_value` helper |
| `test/test_temporal_processes.py` | modified | +31 unit tests (parameterization flag, state-centered shape/site/prior-equivalence) |
| `test/test_helpers.py` | modified | `fixed_ar1_state`, `fixed_differenced_ar1_state` factories |
| `test/integration/conftest.py` | modified | `he_model_state_centered`, `he_weekly_rt_model_state_centered`, `he_weekly_model_state_centered` fixtures |
| `test/integration/test_population_infections_he_state_centered.py` | new | 5 end-to-end tests, daily Rt |
| `test/integration/test_population_infections_he_weekly_rt_state_centered.py` | new | 5 end-to-end tests, weekly Rt via `WeeklyTemporalProcess` |
| `_typos.toml` | modified | Whitelist `reparametrized_params` (NumPyro upstream attribute name) |


cdc-mitzimorris:
Benchmark tests on the models written for https://github.com/CDCgov/PyRenew/pull/819  show a 2x to 4x speedup.
Working on benchmark scripts for more systematic testing.

cdc-mitzimorris:
ran the benchmarks on my machine - here are the results:

```
time python -m benchmarks.suites.rt_params --candidate he --prior both --repeats 3
rt_params suite: 4 candidate(s) x 2 prior(s) x 3 repeat(s) = 24 fits
>> fitting he_daily_innovation@sd=0.01,ar=0.9 (repeat 1/3) ...
   done he_daily_innovation@sd=0.01,ar=0.9 (repeat 1/3): 62.9s, divergences=0, min ESS/s=0.15
>> fitting he_daily_innovation@sd=0.01,ar=0.9 (repeat 2/3) ...
   done he_daily_innovation@sd=0.01,ar=0.9 (repeat 2/3): 66.4s, divergences=0, min ESS/s=0.25
>> fitting he_daily_innovation@sd=0.01,ar=0.9 (repeat 3/3) ...
   done he_daily_innovation@sd=0.01,ar=0.9 (repeat 3/3): 68.4s, divergences=0, min ESS/s=0.14
>> fitting he_daily_state@sd=0.01,ar=0.9 (repeat 1/3) ...
   done he_daily_state@sd=0.01,ar=0.9 (repeat 1/3): 63.5s, divergences=0, min ESS/s=5.79
>> fitting he_daily_state@sd=0.01,ar=0.9 (repeat 2/3) ...
   done he_daily_state@sd=0.01,ar=0.9 (repeat 2/3): 62.7s, divergences=0, min ESS/s=5.59
>> fitting he_daily_state@sd=0.01,ar=0.9 (repeat 3/3) ...
   done he_daily_state@sd=0.01,ar=0.9 (repeat 3/3): 63.4s, divergences=0, min ESS/s=6.61
>> fitting he_weekly_innovation@sd=0.01,ar=0.9 (repeat 1/3) ...
   done he_weekly_innovation@sd=0.01,ar=0.9 (repeat 1/3): 68.9s, divergences=0, min ESS/s=1.05
>> fitting he_weekly_innovation@sd=0.01,ar=0.9 (repeat 2/3) ...
   done he_weekly_innovation@sd=0.01,ar=0.9 (repeat 2/3): 69.3s, divergences=0, min ESS/s=0.12
>> fitting he_weekly_innovation@sd=0.01,ar=0.9 (repeat 3/3) ...
   done he_weekly_innovation@sd=0.01,ar=0.9 (repeat 3/3): 70.7s, divergences=0, min ESS/s=0.47
>> fitting he_weekly_state@sd=0.01,ar=0.9 (repeat 1/3) ...
   done he_weekly_state@sd=0.01,ar=0.9 (repeat 1/3): 17.9s, divergences=0, min ESS/s=28.92
>> fitting he_weekly_state@sd=0.01,ar=0.9 (repeat 2/3) ...
   done he_weekly_state@sd=0.01,ar=0.9 (repeat 2/3): 16.6s, divergences=0, min ESS/s=30.93
>> fitting he_weekly_state@sd=0.01,ar=0.9 (repeat 3/3) ...
   done he_weekly_state@sd=0.01,ar=0.9 (repeat 3/3): 16.8s, divergences=0, min ESS/s=32.74
>> fitting he_daily_innovation@sd=0.1,ar=0.5 (repeat 1/3) ...
   done he_daily_innovation@sd=0.1,ar=0.5 (repeat 1/3): 79.7s, divergences=0, min ESS/s=0.03
>> fitting he_daily_innovation@sd=0.1,ar=0.5 (repeat 2/3) ...
   done he_daily_innovation@sd=0.1,ar=0.5 (repeat 2/3): 79.4s, divergences=0, min ESS/s=0.03
>> fitting he_daily_innovation@sd=0.1,ar=0.5 (repeat 3/3) ...
   done he_daily_innovation@sd=0.1,ar=0.5 (repeat 3/3): 80.2s, divergences=0, min ESS/s=0.03
>> fitting he_daily_state@sd=0.1,ar=0.5 (repeat 1/3) ...
   done he_daily_state@sd=0.1,ar=0.5 (repeat 1/3): 30.0s, divergences=0, min ESS/s=10.49
>> fitting he_daily_state@sd=0.1,ar=0.5 (repeat 2/3) ...
   done he_daily_state@sd=0.1,ar=0.5 (repeat 2/3): 31.4s, divergences=0, min ESS/s=9.56
>> fitting he_daily_state@sd=0.1,ar=0.5 (repeat 3/3) ...
   done he_daily_state@sd=0.1,ar=0.5 (repeat 3/3): 29.4s, divergences=0, min ESS/s=10.88
>> fitting he_weekly_innovation@sd=0.1,ar=0.5 (repeat 1/3) ...
   done he_weekly_innovation@sd=0.1,ar=0.5 (repeat 1/3): 72.2s, divergences=0, min ESS/s=0.03
>> fitting he_weekly_innovation@sd=0.1,ar=0.5 (repeat 2/3) ...
   done he_weekly_innovation@sd=0.1,ar=0.5 (repeat 2/3): 72.8s, divergences=0, min ESS/s=0.04
>> fitting he_weekly_innovation@sd=0.1,ar=0.5 (repeat 3/3) ...
   done he_weekly_innovation@sd=0.1,ar=0.5 (repeat 3/3): 73.8s, divergences=0, min ESS/s=0.04
>> fitting he_weekly_state@sd=0.1,ar=0.5 (repeat 1/3) ...
   done he_weekly_state@sd=0.1,ar=0.5 (repeat 1/3): 22.6s, divergences=0, min ESS/s=42.72
>> fitting he_weekly_state@sd=0.1,ar=0.5 (repeat 2/3) ...
   done he_weekly_state@sd=0.1,ar=0.5 (repeat 2/3): 22.3s, divergences=0, min ESS/s=44.03
>> fitting he_weekly_state@sd=0.1,ar=0.5 (repeat 3/3) ...
   done he_weekly_state@sd=0.1,ar=0.5 (repeat 3/3): 22.3s, divergences=0, min ESS/s=52.17

--- synthetic_he_weekly_hospital | cadence=daily | innovation_sd=0.01 ---
metric                   innovation        state  state/innov
--------------------------------------------------------------
Wall time (s)                  65.9         63.2        0.96x
ESS/s Rt (median)             0.748       27.329     36.53x *
ESS/s Rt (min)                0.183        5.997     32.78x *
Divergences                       0            0          n/a
Tree depth (mean)             10.00         9.91        0.99x
Tree depth (max)                 10           10        1.00x
E-BFMI (min)                  0.888        0.943      1.06x *
R-hat Rt (max)                1.275        1.006      0.79x *

--- synthetic_he_weekly_hospital | cadence=weekly | innovation_sd=0.01 ---
metric                   innovation        state  state/innov
--------------------------------------------------------------
Wall time (s)                  69.7         17.1      0.25x *
ESS/s Rt (median)             1.856       98.404     53.02x *
ESS/s Rt (min)                0.546       30.864     56.49x *
Divergences                       0            0          n/a
Tree depth (mean)             10.00         7.17      0.72x *
Tree depth (max)                 10            9      0.90x *
E-BFMI (min)                  0.896        0.925        1.03x
R-hat Rt (max)                1.150        1.005      0.87x *

--- synthetic_he_weekly_hospital | cadence=daily | innovation_sd=0.1 ---
metric                   innovation        state  state/innov
--------------------------------------------------------------
Wall time (s)                  79.8         30.2      0.38x *
ESS/s Rt (median)             0.078       72.302    928.36x *
ESS/s Rt (min)                0.032       10.311    322.37x *
Divergences                       0            0          n/a
Tree depth (mean)             10.00         8.06      0.81x *
Tree depth (max)                 10           10        1.00x
E-BFMI (min)                  0.901        0.920        1.02x
R-hat Rt (max)                2.350        1.014      0.43x *

--- synthetic_he_weekly_hospital | cadence=weekly | innovation_sd=0.1 ---
metric                   innovation        state  state/innov
--------------------------------------------------------------
Wall time (s)                  73.0         22.4      0.31x *
ESS/s Rt (median)             0.098       75.878    772.58x *
ESS/s Rt (min)                0.038       46.309   1226.16x *
Divergences                       0            0          n/a
Tree depth (mean)             10.00         7.58      0.76x *
Tree depth (max)                 10            9      0.90x *
E-BFMI (min)                  0.980        0.941        0.96x
R-hat Rt (max)                2.165        1.004      0.46x *

(* marks an improvement over innovation; ratios are state / innovation)

Wrote results to benchmarks/results

real	21m15.997s
user	80m23.152s
sys	0m7.630s
```

cdc-mitzimorris:
@dylanhmorris @sbidari ready for code review

SamuelBrand1:
Linking https://github.com/epiforecasts/EpiNow2/pull/1396 because it seems to cover a similar reparametrisation

cdc-mitzimorris:
> Linking [epiforecasts/EpiNow2#1396](https://github.com/epiforecasts/EpiNow2/pull/1396) because it seems to cover a similar reparametrisation

thanks!  good to know there's good precedent.

cdc-mitzimorris:
To simplify this PR, I propose splitting out the `benchmarks` code, which adds  **~2,346 lines of source** plus a **718-line test file**, separately reviewable from the API change.

The code added/ modified by this PR, exclusive of benchmarks code is:

| File | Type | +/− | Purpose |
|---|---|---:|---|
| `pyrenew/latent/state_centered_distributions.py` | new | +387 | `StateAR1`, `StateDifferencedAR1` custom NumPyro distributions |
| `pyrenew/latent/temporal_processes.py` | modified | +207 / −71 | `parameterization` flag on `AR1`, `DifferencedAR1`, `RandomWalk`; `_prepare_initial_value` helper |
| `test/test_temporal_processes.py` | modified | +493 | Unit tests for parameterization flag, state-centered shape/site/prior-equivalence |
| `test/test_helpers.py` | modified | +46 | `fixed_ar1_state`, `fixed_differenced_ar1_state` factories |
| `test/integration/conftest.py` | modified | +122 / −111 | `he_model_state_centered`, `he_weekly_rt_model_state_centered`, `he_weekly_model_state_centered` fixtures |
| `test/integration/test_population_infections_he_state_centered.py` | new | +185 | 5 end-to-end tests, daily Rt |
| `test/integration/test_population_infections_he_weekly_rt_state_centered.py` | new | +227 | 5 end-to-end tests, weekly Rt via `WeeklyTemporalProcess` |
| `_typos.toml` | modified | +6 | Whitelist `reparametrized_params` (NumPyro upstream attribute) |

**API-change subtotal:** ~1,673 added / ~182 removed.

Other changes also picked up on this branch (not in the PR description)

| File | Type | +/− | Notes |
|---|---|---:|---|
| `.gitignore` | modified | +3 | |
| `pyproject.toml` | modified | +2 | |
| `docs/tutorials/_quarto.yml` | modified | +1 / −1 | |
| `docs_scripts/add_markdown_to_divs.py` | new | +32 | Replacement for the removed postprocessing script |
| `docs_scripts/postprocess_generated_markdown.py` | removed | −62 | Replaced by `add_markdown_to_divs.py` |
| `test/test_docs_postprocessing.py` | removed | −39 | Tests for the removed script |
| `test/test_distributional_rv.py` | modified | +2 / −2 | Minor tweak |

**Other-changes subtotal:** ~40 added / ~104 removed.

Benchmarking component adds roughly 3K lines of code.  (oof!)

| File | Type | Lines | Purpose |
|---|---|---:|---|
| `benchmarks/__init__.py` | new | 11 | Package docstring; entry-point conventions |
| `benchmarks/README.md` | new | 177 | How to run suites, layout, extension points |
| `benchmarks/core/__init__.py` | new | 1 | Package marker |
| `benchmarks/core/signals.py` | new | 104 | `DatasetProvider` protocol — seam between suites and data source |
| `benchmarks/core/datasets.py` | new | 101 | Synthetic `DatasetProvider` wrapping `pyrenew/datasets/` |
| `benchmarks/core/real_data.py` | new | 245 | Real-data `DatasetProvider` (CDC NHSN + NSSP feeds) |
| `benchmarks/core/reference_data.py` | new | 85 | Static US location/population table (replaces R `forecasttools` dep) |
| `benchmarks/core/priors.py` | new | 42 | Benchmark-local priors mirroring HEW production subset |
| `benchmarks/core/models.py` | new | 295 | `build_*` model builders pairing a dataset with a model family |
| `benchmarks/core/runner.py` | new | 275 | Single-fit MCMC wrapper; collects timing + diagnostic metrics |
| `benchmarks/core/reporting.py` | new | 610 | CSV/JSON/Markdown reporters; pair comparisons, candidate summaries |
| `benchmarks/suites/__init__.py` | new | 1 | Package marker |
| `benchmarks/suites/rt_params.py` | new | 399 | The `rt_params` suite — compares `innovation` vs `state` on weekly Rt |
| `test/test_benchmarks_rt_params.py` | new | 718 | Unit tests for the suite, runner, reporting, and providers |

**Totals:** 13 source files (~2,346 lines) + 718 lines of tests. The benchmark framework is self-contained under `benchmarks/`; the only PR change outside that tree is the test file above.


@dylanhmorris, @sbidari

cdc-mitzimorris:
@dylanhmorris and @sbidari - ready for review.

pyrenew/latent/state_centered_distributions.py
dylanhmorris, line 1: Put these in the `pyrenew/distributions` module?

pyrenew/latent/state_centered_distributions.py
```diff
@@ -0,0 +1,387 @@
...
+
+    Given a deterministic initial state $x_0$ = ``initial_loc``:
+
+    $$
+    x_t \sim \mathrm{Normal}(x_{t-1}, \sigma), \quad t = 1, \dots, T
+    $$
+
+    The sampled value is the post-initial path
+    $[x_1, x_2, \ldots, x_{\mathrm{num\_steps}}]$ of length ``num_steps``.
+    """
```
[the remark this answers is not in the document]
dylanhmorris: Why not leverage `dist.GaussianRandomWalk` here? AFAICT, the one issue is that `dist.GaussianRandomWalk` for our purposes is that it treats the first sampled value as coming from a necessarily zero-mean Gaussian, whereas we'd like to treat it as coming from a Gaussian with arbitrary mean (given by our `initial_loc`). But we can handle that by treating initial loc as an offset. In other words, if we have an appropriate internal `self.gaussian_random_walk_`, then for forward sampling:

```python
def sample(self, key, sample_shape):
   return self.initial_loc + self.gaussian_random_walk_.sample(key=key, sample_shape=sample_shape)
```

And for log prob evaluation

```python
def log_prob(self, value):
    return self.gaussian_random_walk_.log_prob(value - self.initial_loc)
```

We could either turn `StateRandomWalk` into a light wrapper of `GaussianRandomWalk` (as in my illustration above) or use a `dist.GaussianRandomWalk` directly within the `RandomWalk` `TemporalProcess` and handle the offsetting there.

pyrenew/latent/state_centered_distributions.py
```diff
@@ -0,0 +1,387 @@
...
+        init_with_event = jnp.expand_dims(initial_loc, -1)
+        init_bcast = jnp.broadcast_to(init_with_event, value.shape[:-1] + (1,))
+        v = jnp.concatenate([init_bcast, value], axis=-1)
+        step_probs = Normal(v[..., :-1], jnp.expand_dims(scale, -1)).log_prob(
+            v[..., 1:]
+        )
+        return jnp.sum(step_probs, axis=-1)
+
+
+class StateAR1(Distribution):
```
dylanhmorris, line 125: Avoiding the scan in log prob evaluation and using it only in forward sampling is a really nice feature of doing this as true `numpyro` `Distributions`, rather than as PyRenew `RandomVariables`.

Question: does this argue for reimplementing the innovation-based parameterizations in the same way? Do you expect performance implications?

pyrenew/latent/state_centered_distributions.py
```diff
@@ -0,0 +1,387 @@
...
+    $$
+    $$
+    x_t \sim \mathrm{Normal}(\phi \, x_{t-1}, \sigma), \quad t = 1, \dots, T-1
+    $$
+
+    where $\sigma_{\text{stat}} = \sigma / \sqrt{1 - \phi^2}$ is the
+    stationary standard deviation, $\mu_0$ is ``initial_loc``, $\phi$ is
+    ``autoreg``, and $\sigma$ is ``scale``.
+
+    The sampled value is the full path $[x_0, x_1, \ldots, x_{T-1}]$.
```
dylanhmorris: AR1 is different from the random walk in returning full path versus post init path. I understand that the temporal process version (below) does sample the initial state. But it does so unbundled in the innovation parametrization, so why not do the same in the state parametrization?

dylanhmorris (changes requested):
Thanks @cdc-mitzimorris! A couple design / organization questions and suggestions before I review further

pyrenew/latent/state_centered_distributions.py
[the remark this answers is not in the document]
cdc-mitzimorris: Good call - `StateRandomWalk` was an exact location shift of `dist.GaussianRandomWalk`: a zero-mean Gaussian walk supplies the path and initial_loc offsets it.

Changed to the light-wrapper form (offset inside the Distribution) rather than offsetting at the `RandomWalk` call site.  Keeping the offset inside means the numpyro.sample site value is the state path $[x_1, \ldots, x_T]$ rather than a zero-centered walk, which is the whole point of the "state-centered" naming and  is nicer to read in the trace.

pyrenew/latent/state_centered_distributions.py
cdc-mitzimorris: I would prefer to keep them in the module with the high-level components that use them.

pyrenew/latent/state_centered_distributions.py
cdc-mitzimorris: worth investigating, file an issue?    for our purposes, the number of timepoints is small, so it might not matter.

pyrenew/latent/state_centered_distributions.py
cdc-mitzimorris: re-parameterized so all implementations match.

cdc-mitzimorris:
> Thanks @cdc-mitzimorris! A couple design / organization questions and suggestions before I review further

thanks @dylanhmorris. questions answered, suggestion taken.   ready for more.

pyrenew/latent/state_centered_distributions.py
```diff
@@ -0,0 +1,378 @@
...
+        scale: ArrayLike,
+        initial_loc: ArrayLike = 0.0,
+        num_steps: int = 1,
+        *,
+        validate_args: bool | None = None,
+    ) -> None:
+        """Construct a state-centered random-walk distribution."""
+        if not isinstance(num_steps, int) or num_steps <= 0:
+            raise ValueError(f"num_steps must be a positive integer; got {num_steps!r}")
+        self.scale = scale
```
dylanhmorris: Keep only in one place (within `self.gaussian_random_walk_`). If we need top-level access, define a @property
```suggestion
```

pyrenew/latent/state_centered_distributions.py
```diff
@@ -0,0 +1,378 @@
...
+        self.initial_loc = initial_loc
+        self.num_steps = num_steps
```
dylanhmorris: As with scale, keep in only one place.
```suggestion
```

pyrenew/latent/state_centered_distributions.py
```diff
@@ -0,0 +1,378 @@
...
+        Forward-sample a post-initial random-walk state path.
+
+        Returns
+        -------
+        ArrayLike
+            Array of shape ``sample_shape + batch_shape + (num_steps,)``.
+        """
+        assert is_prng_key(key)
+        walk = self.gaussian_random_walk_.sample(key, sample_shape)
+        return jnp.expand_dims(jnp.asarray(self.initial_loc), -1) + walk
```
dylanhmorris, line 100: Can this not be defered this to the broadcasting rules? If it can't, should leave an explanatory comment.

```suggestion
        return self.initial_loc + walk
```

pyrenew/latent/state_centered_distributions.py
```diff
@@ -0,0 +1,378 @@
...
+            Post-initial path of shape
+            ``sample_shape + batch_shape + (num_steps,)``.
+
+        Returns
+        -------
+        ArrayLike
+            Log-density of shape ``sample_shape + batch_shape``.
+        """
+        offset = jnp.expand_dims(jnp.asarray(self.initial_loc), -1)
+        return self.gaussian_random_walk_.log_prob(value - offset)
```
dylanhmorris, line 122: Same question as above. Can this not be deferred this to the broadcasting rules? If it can't, should leave an explanatory comment.

```suggestion
        return self.gaussian_random_walk_.log_prob(value - self.initial_loc)
```

pyrenew/latent/state_centered_distributions.py
dylanhmorris: Opened #833

pyrenew/latent/state_centered_distributions.py
```diff
@@ -0,0 +1,378 @@
...
+from numpyro.distributions import constraints
+from numpyro.distributions.continuous import GaussianRandomWalk, Normal
+from numpyro.distributions.distribution import Distribution
+from numpyro.distributions.util import validate_sample
+from numpyro.util import is_prng_key
+
+
+class StateRandomWalk(Distribution):
+    r"""
+    State-centered random-walk prior on a post-initial state path.
```
dylanhmorris:
```suggestion
    State-centered random walk on a post-initial state path.
```

pyrenew/latent/state_centered_distributions.py
```diff
@@ -0,0 +1,378 @@
...
+        ArrayLike
+            Log-density of shape ``sample_shape + batch_shape``.
+        """
+        offset = jnp.expand_dims(jnp.asarray(self.initial_loc), -1)
+        return self.gaussian_random_walk_.log_prob(value - offset)
+
+
+class StateAR1(Distribution):
+    r"""
+    State-centered AR(1) prior on a post-initial state path.
```
dylanhmorris:
```suggestion
    State-centered AR(1) process for a post-initial state path.
```

pyrenew/latent/state_centered_distributions.py
dylanhmorris, line 1: Not seeing rendered API documentation for this module in the website preview. Can you check @cdc-mitzimorris?

pyrenew/latent/state_centered_distributions.py
```diff
@@ -0,0 +1,378 @@
...
+    x_t \sim \mathrm{Normal}(\phi \, x_{t-1}, \sigma), \quad t = 1, \dots, T
+    $$
+
+    where $\phi$ is ``autoreg`` and $\sigma$ is ``scale``.
+
+    The sampled value is the post-initial path
+    $[x_1, x_2, \ldots, x_{\mathrm{num\_steps}}]$ of length ``num_steps``.
+    The initial state $x_0$ is not part of the sample; it is supplied as
+    ``initial_loc`` and used to score the first transition. A random initial
+    state drawn from the stationary distribution is handled by the calling
```
dylanhmorris:
```suggestion
    state drawn from the stationary distribution can be handled by the calling
```

pyrenew/latent/state_centered_distributions.py
dylanhmorris: Some argument that we may want to generalize this to abitrary AR(p) (matching `ARProcess` in `pyrenew.process`, but I think we can defer that to #833, unless you disagree and think it's just worth doing here @cdc-mitzimorris.

pyrenew/latent/state_centered_distributions.py
dylanhmorris, line 1: Question about which axis represents time. Should these internal distributions mirror `GaussianRandomWalk`'s convention (time is the final axis) or PyRenew's (time is the first axis, consistent with the outcome of a `scan` over time and with diffrax and other libraries)

The TemporalProcess` wrappers should absolutely follow PyRenew's convention, but we should consider whether the distributions should as well. I lean toward "time-is-the-first-dimension" within the distributions classes themselves (as this mirrors PyRenew, examples with timeseries scans in the numpyro docs ([example](https://num.pyro.ai/en/stable/examples/var2.html)) also other libraries e.g. [diffrax](https://docs.kidger.site/diffrax/).

pyrenew/latent/state_centered_distributions.py
cdc-mitzimorris: changed, added property accessor

pyrenew/latent/state_centered_distributions.py
cdc-mitzimorris: Good question. 

The constraint is structural, not stylistic: a `numpyro.distributions.Distribution` is always `sample_shape + batch_shape + event_shape` with the `event` (time) axis trailing. `log_prob` reduction, `real_vector` validation, plates, and `Predictive` all rely on that. So while these stay Distribution subclasses with `event_shape = (num_steps,)`, their public `sample/log_prob` must be time-last.  Matching GRW and matching NumPyro are the same thing here.

Everything we control is already time-first: the scan math builds `path_time_first`, and the wrappers return `(n_timepoints, n_processes)`. The only time-last code is one `jnp.moveaxis(..., 0, -1)` per class, which just adapts to NumPyro's required layout.

Making the distributions time-first publicly would mean not subclassing `Distribution` (losing `numpyro.sample/Predictive/reparam`).  This would require enough work such that it would warrent filing an issue and then doing this as a separate PR.

pyrenew/latent/state_centered_distributions.py
cdc-mitzimorris: added comment.

pyrenew/latent/state_centered_distributions.py
cdc-mitzimorris: added comment.

pyrenew/latent/state_centered_distributions.py
cdc-mitzimorris: would prefer to defer.

pyrenew/latent/state_centered_distributions.py
cdc-mitzimorris: good catch.  fixed `latent/__initi__.py`

pyrenew/latent/state_centered_distributions.py
```diff
@@ -0,0 +1,406 @@
...
+    Given a deterministic initial state $x_0$ = ``initial_loc``:
+
+    $$
+    x_t \sim \mathrm{Normal}(x_{t-1}, \sigma), \quad t = 1, \dots, T
+    $$
+
+    The sampled value is the post-initial path
+    $[x_1, x_2, \ldots, x_{\mathrm{num\_steps}}]$ of length ``num_steps``.
+
+    This is a location shift of [`numpyro.distributions.GaussianRandomWalk`][]:
```
dylanhmorris:
```suggestion
    This is a location shift of [`numpyro.distributions.GaussianRandomWalk`][numpyro.distributions.continuous.GaussianRandomWalk]:
```
Cross-ref is broken. This should fix, I believe (but do check).

pyrenew/latent/state_centered_distributions.py
```diff
@@ -0,0 +1,406 @@
...
+        not enforced.
+    scale
+        Innovation standard deviation $\sigma$. Must be positive.
+    initial_loc
+        Deterministic initial state $x_0$. Used to score the first
+        transition; not itself sampled. Defaults to ``0.0``.
+    num_steps
+        Length of the post-initial path. Must be a positive integer.
+    validate_args
+        Forwarded to the base [`numpyro.distributions.Distribution`][].
```
dylanhmorris:
```suggestion
        Forwarded to the base [`Distribution`][].
```
Cross-ref is broken. See if the above works via scoped cross-referencing (since you imported `Distribution`). If not, this should work I think:

```suggestion
        Forwarded to the base [`Distribution`][numpyro.distributions.distribution.Distribution].
```

pyrenew/latent/state_centered_distributions.py
```diff
@@ -0,0 +1,406 @@
...
+        self.initial_loc = initial_loc
+        self.num_steps = num_steps
+
+        batch_shape = lax.broadcast_shapes(
+            jnp.shape(autoreg),
+            jnp.shape(scale),
+            jnp.shape(initial_loc),
+        )
+        event_shape = (num_steps,)
+        super().__init__(batch_shape, event_shape, validate_args=validate_args)
```
dylanhmorris:
```suggestion
        super().__init__(batch_shape, (num_steps,), validate_args=validate_args)
```

pyrenew/latent/state_centered_distributions.py
```diff
@@ -0,0 +1,406 @@
...
+        -------
+        ArrayLike
+            Array of shape ``sample_shape + batch_shape + (num_steps,)``.
+        """
+        assert is_prng_key(key)
+
+        per_step_shape = sample_shape + self.batch_shape
+        autoreg = jnp.broadcast_to(jnp.asarray(self.autoreg), per_step_shape)
+        scale = jnp.broadcast_to(jnp.asarray(self.scale), per_step_shape)
+        initial_loc = jnp.broadcast_to(jnp.asarray(self.initial_loc), per_step_shape)
```
dylanhmorris, line 217: Is manual broadcasting necessary here? If so, add explanatory comment.

pyrenew/latent/state_centered_distributions.py
```diff
@@ -0,0 +1,406 @@
...
+        noise = random.normal(key, shape=(self.num_steps,) + per_step_shape)
+
+        def step(
+            prev: ArrayLike, z_t: ArrayLike
+        ) -> tuple[ArrayLike, ArrayLike]:  # numpydoc ignore=GL08
+            new = autoreg * prev + scale * z_t
+            return new, new
+
+        _, xs = lax.scan(step, initial_loc, noise)
+        return jnp.moveaxis(xs, 0, -1)
```
dylanhmorris: Suggest an explanatory comment
```suggestion
        return jnp.moveaxis(xs, 0, -1) # ensure time is the trailing axis (length num_steps)
```

pyrenew/latent/state_centered_distributions.py
```diff
@@ -0,0 +1,406 @@
...
+
+    Generative form, given a deterministic initial state $x_0$ = ``initial_loc``:
+
+    $$
+    x_1 \sim \mathrm{Normal}(x_0, \sigma_{\text{stat}})
+    $$
+    $$
+    x_t \sim \mathrm{Normal}(x_{t-1} + \phi \, (x_{t-1} - x_{t-2}), \sigma),
+    \quad t \geq 2
+    $$
```
dylanhmorris: Math isn't rendering properly in docs. I suspect an extra line extra line break separating the two display equations might fix

```suggestion
    $$
    x_1 \sim \mathrm{Normal}(x_0, \sigma_{\mathrm{stat}})
    $$
    
    $$
    x_t \sim \mathrm{Normal}(x_{t-1} + \phi \, (x_{t-1} - x_{t-2}), \sigma),
    \quad t \geq 2
    $$
```

pyrenew/latent/state_centered_distributions.py
```diff
@@ -0,0 +1,406 @@
...
+
+    where $\sigma_{\text{stat}} = \sigma / \sqrt{1 - \phi^2}$, $\phi$ is
```
dylanhmorris:
```suggestion
    where $\sigma_{\mathrm{stat}} = \sigma / \sqrt{1 - \phi^2}$, $\phi$ is
```

pyrenew/latent/state_centered_distributions.py
dylanhmorris: I think the same argument applies to the `StateDifferencedAR1` and the `DifferencedAR1` temporal process. We should sample only the post-init path in the distribution implementation.

pyrenew/latent/state_centered_distributions.py
```diff
@@ -0,0 +1,406 @@
...
+        v = jnp.concatenate([init_bcast, value], axis=-1)
+
+        scale_t = jnp.expand_dims(scale, -1)
+        autoreg_t = jnp.expand_dims(autoreg, -1)
+        step_locs = autoreg_t * v[..., :-1]
+        step_probs = Normal(step_locs, scale_t).log_prob(v[..., 1:])
+        return jnp.sum(step_probs, axis=-1)
+
+
+class StateDifferencedAR1(Distribution):
```
dylanhmorris, line 266: Rewrite to only sample the post-init path? See
https://github.com/CDCgov/PyRenew/pull/828#discussion_r3416758062

pyrenew/latent/state_centered_distributions.py
```diff
@@ -0,0 +1,406 @@
...
+    r"""
+    State-centered differenced AR(1) prior on a length-``num_steps`` post-initial path.
```
dylanhmorris, line 268:
```suggestion
    State-centered differenced AR(1) process for a length-``num_steps`` post-initial path.
```

pyrenew/latent/state_centered_distributions.py
dylanhmorris: in other words, the distribution should require an initial loc and and initial diff at construction (and/or x0 and x1, with the initial diff computed)

pyrenew/latent/state_centered_distributions.py
```diff
@@ -0,0 +1,406 @@
...
+        scale = jnp.asarray(self.scale)
+        autoreg = jnp.asarray(self.autoreg)
+        initial_loc = jnp.asarray(self.initial_loc)
+
+        init_with_event = jnp.expand_dims(initial_loc, -1)
+        init_bcast = jnp.broadcast_to(init_with_event, value.shape[:-1] + (1,))
+        v = jnp.concatenate([init_bcast, value], axis=-1)
+
+        scale_t = jnp.expand_dims(scale, -1)
+        autoreg_t = jnp.expand_dims(autoreg, -1)
```
dylanhmorris: Should first broadcast the `initial_loc` to the per step shape and then add a trailing time dimension. Calling expand_dims before broadcasting to the per step shape is extraneous.

Also suggested reordering slightly and standardizing variable names

```suggestion
        per_step_shape = jnp.shape(value)[:-1])
        initial_loc = jnp.broadcast_to(initial_loc, per_step_shape)
        
        # add length-1 trailing time dimension to fixed-in-time paramaters
        initial_loc_t = jnp.expand_dims(initial_step, -1)
        scale_t = jnp.expand_dims(scale, -1)
        autoreg_t = jnp.expand_dims(autoreg, -1)
        
        v = jnp.concatenate([initial_loc_t, value], axis=-1)        
```

pyrenew/latent/temporal_processes.py
```diff
@@ -239,32 +301,50 @@ def sample(
...
+                noise_sd=innovation_sd,
+                noise_name=f"{name_prefix}_noise",
+            )
 
+        stationary_sd = innovation_sd / jnp.sqrt(1 - autoreg**2)
         with numpyro.plate(f"{name_prefix}_init_plate", n_processes):
             init_states = numpyro.sample(
                 f"{name_prefix}_init",
                 dist.Normal(initial_value, stationary_sd),
             )
```
dylanhmorris: Move init plate out of the fork
```suggestion
        stationary_sd = innovation_sd / jnp.sqrt(1 - autoreg**2)
        with numpyro.plate(f"{name_prefix}_init_plate", n_processes):
            init_states = numpyro.sample(
                f"{name_prefix}_init",
                dist.Normal(initial_value, stationary_sd),
            )
        if self.parameterization == "innovation":
            return self.ar_process(
                n=n_timepoints,
                init_vals=init_states[jnp.newaxis, :],
                autoreg=autoreg_broadcast[jnp.newaxis, :],
                noise_sd=innovation_sd,
                noise_name=f"{name_prefix}_noise",
            )

```

pyrenew/latent/temporal_processes.py
```diff
@@ -239,32 +301,50 @@ def sample(
...
+                autoreg=autoreg_broadcast,
+                scale=scale_broadcast,
+                initial_loc=init_states,
+                num_steps=n_timepoints - 1,
+            ),
         )
-
-        return trajectories
+        x = jnp.concatenate([init_states[:, jnp.newaxis], post_init], axis=-1)
+        return x.T
```
dylanhmorris:
```suggestion
        return jnp.moveaxis(xs, -1, 0) # ensure time is the leading axis (length n_timepoints)
```

pyrenew/latent/temporal_processes.py
```diff
@@ -279,6 +359,19 @@ class DifferencedAR1(TemporalProcess):
...
     a simplified, protocol-compliant interface.
 
+    The ``parameterization`` argument selects between sampling standardized
+    innovations on the differences (``"innovation"``) and sampling the state
+    path ``x[1:T]`` directly under the priors
+
+    ```
+    x[1] ~ Normal(x[0], innovation_sd / sqrt(1 - autoreg^2))
+    x[t] ~ Normal(x[t-1] + autoreg * (x[t-1] - x[t-2]), innovation_sd)   t >= 2
+    ```
```
dylanhmorris: Consider writing as math rather than as code, to be consistent with other docs.

pyrenew/latent/temporal_processes.py
```diff
@@ -367,33 +472,44 @@ def sample(
...
         autoreg_broadcast = jnp.broadcast_to(jnp.asarray(autoreg), (n_processes,))
 
-        stationary_sd = innovation_sd / jnp.sqrt(1 - autoreg**2)
-
-        with numpyro.plate(f"{name_prefix}_init_rate_plate", n_processes):
-            init_rates = numpyro.sample(
-                f"{name_prefix}_init_rate",
-                dist.Normal(0, stationary_sd),
+        if self.parameterization == "innovation":
+            stationary_sd = innovation_sd / jnp.sqrt(1 - autoreg**2)
```
dylanhmorris: If we change `StateDifferencedAR1` to be only on the post-init path, per my request above, this can be simplified to move all initial rate/ initial value sampling outside the parameterization fork (similar to AR1).

pyrenew/latent/temporal_processes.py
```diff
@@ -367,33 +472,44 @@ def sample(
...
+                autoreg=autoreg_broadcast,
+                scale=scale_broadcast,
+                initial_loc=initial_value,
+                num_steps=n_timepoints - 1,
+            ),
         )
-
-        return trajectories
+        full_path = jnp.concatenate([initial_value[:, jnp.newaxis], post_init], axis=-1)
+        return full_path.T
```
dylanhmorris:
```suggestion
        return jnp.moveaxis(full_path, -1, 0) # ensure time is the leading axis (length n_timepoints)
```

pyrenew/latent/temporal_processes.py
```diff
@@ -482,28 +616,41 @@ def sample(
...
+        post_init = numpyro.sample(
+            f"{name_prefix}_state",
+            StateRandomWalk(
+                scale=scale_broadcast,
+                initial_loc=initial_value,
+                num_steps=n_timepoints - 1,
+            ),
         )
+        x = jnp.concatenate([initial_value[:, jnp.newaxis], post_init], axis=-1)
+        return x.T
```
dylanhmorris:
```suggestion
        return jnp.moveaxis(x, -1, 0) # ensure time is the leading axis (length n_timepoints)
```

test/test_temporal_processes.py
```diff
@@ -505,6 +642,361 @@ def test_repr_uses_random_variable_argument_names(self, process, expected):
...
+    def test_repr_shows_parameterization(self, process_cls, kwargs):
+        """``__repr__`` exposes the current parameterization for diagnostics."""
+        process = process_cls(**kwargs, parameterization="state")
+        assert "parameterization='state'" in repr(process)
+
+
+class TestStateCenteredRandomWalk:
+    """State-centered RandomWalk samples the state path directly via GaussianRandomWalk."""
+
+    def test_return_shape(self):
```
dylanhmorris: Test properties like the return shape (that should be the same for all parametrizations) where the overall process class is tested (iterating over all the possible parametrizations). 

Reserve `TestStateCenteredRandomWalk` for properties that are unique to the `parameterization="state"` case (e.g.  `test_trace_has_state_site_not_step_site` below).

test/test_temporal_processes.py
dylanhmorris, line 1: Throughout, for organization and DRY, ensure tests of parameterization-agnostic `TemporalProcess` behavior are separated from tests of parameterization-specific `TemporalProcess` behavior. Example: https://github.com/CDCgov/PyRenew/pull/828/changes#r3421561443

dylanhmorris (changes requested):
Thanks @cdc-mitzimorris! Two requested changes:

- Change the `StateDifferencedAR1` `Distribution` to sample only the post-init path (https://github.com/CDCgov/PyRenew/pull/828/changes#r3421572974) and adjust the sampling logic for the downstream `StateDifferencedAR1` `TemporalProcess` accordingly
- Ensure tests of parameterization-agnostic `TemporalProcess` properties are conducted in one place for all parameterizations https://github.com/CDCgov/PyRenew/pull/828/changes#r3421572974

Also left some minor suggestions (mostly docs rendering and variable naming)

pyrenew/latent/state_centered_distributions.py
dylanhmorris: Makes sense. I think we should just accept that Distributions will have trailing time and RandomVariables will have leading time.

pyrenew/latent/state_centered_distributions.py
cdc-mitzimorris: fixed.  problem was reference itself, not sytax.

pyrenew/latent/state_centered_distributions.py
cdc-mitzimorris: fixed - problem was reference, not syntax.

pyrenew/latent/state_centered_distributions.py
cdc-mitzimorris: necessary for initial_loc, not autoreg or scale.
added comment.

pyrenew/latent/state_centered_distributions.py
cdc-mitzimorris: added

test/test_temporal_processes.py
cdc-mitzimorris: Summary of the test changes

  Distribution-level tests fixed for the refactor (TestStateCenteredDistribution*):
  - test_state_differenced_ar1_log_prob_matches_manual_transition_sum — now passes initial_diff, drops the stationary init_prob term, and prepends both x0 and x1 to score only the post-init transitions.
  - test_state_differenced_ar1_single_step_sample_matches_initial_transition — now checks the first genuine innovation step $x_2 = x_1 + \phi(x_1-x_0) + \sigma z$ given $x_0$ and $d_1$, instead of the old stationary draw of
  $x_1$.

  Reorganization (reviewer's request):
  - TestTemporalProcessVectorizedSampling now parametrizes over ["innovation", "state"], so return-shape is verified for every process across both parameterizations in one place.
  - New TestDeterministicInitialStateProcesses holds the agnostic "first row equals initial_value" property for the deterministic-init processes (RandomWalk, DifferencedAR1) across both parameterizations. This absorbed and
  replaced TestRandomWalkInitialValues (DRY).
  - The three TestStateCentered* classes are now reserved for state-only behavior (trace sites, state-site shape, prior-moment equivalence, and the n=1/n=2 edge cases).

  Updated for the new behavior:
  - DifferencedAR1 state trace test renamed/flipped: now asserts diff_init_rate and diff_state are present (init rate is now unbundled, mirroring AR1), no diff_noise.
  - test_state_site_shape for DifferencedAR1 now expects n_timepoints - 2 (state path starts at $x_2$).
  - Added test_second_row_equals_initial_value_plus_init_rate (locks in $x_1 = x_0 + d_1$ from the _init_rate site) and test_n_timepoints_two_skips_state_site (covers the new n=2 branch).

test/test_temporal_processes.py
cdc-mitzimorris: refactored.

pyrenew/latent/temporal_processes.py
cdc-mitzimorris: moved.

pyrenew/latent/temporal_processes.py
cdc-mitzimorris: changed here and across the other 3 returns.

pyrenew/latent/temporal_processes.py
cdc-mitzimorris: rewritten.

pyrenew/latent/temporal_processes.py
cdc-mitzimorris: changed.

cdc-mitzimorris:
@dylanhmorris made all changes per requests and suggestions.

test/test_temporal_processes.py
```diff
@@ -54,25 +61,161 @@ def fixed_rw_kwargs(innovation_sd=0.05):
...
+
+        distribution = StateRandomWalk(
+            scale=scale,
+            initial_loc=initial_loc,
+            num_steps=value.shape[-1],
+        )
+
+        full_path = jnp.concatenate([initial_loc[:, None], value], axis=-1)
+        expected = dist.Normal(full_path[:, :-1], scale[:, None]).log_prob(
+            full_path[:, 1:]
```
dylanhmorris:
```suggestion
            value
```

test/test_temporal_processes.py
```diff
@@ -54,25 +61,161 @@ def fixed_rw_kwargs(innovation_sd=0.05):
...
+            autoreg=autoreg,
+            scale=scale,
+            initial_loc=initial_loc,
+            num_steps=value.shape[-1],
+        )
+
+        full_path = jnp.concatenate([initial_loc[:, None], value], axis=-1)
+        transition_locs = autoreg[:, None] * full_path[:, :-1]
+        expected = dist.Normal(transition_locs, scale[:, None]).log_prob(
+            full_path[:, 1:]
```
dylanhmorris:
```suggestion
            value
```

test/test_temporal_processes.py
```diff
@@ -54,25 +61,161 @@ def fixed_rw_kwargs(innovation_sd=0.05):
...
+            initial_diff=initial_diff,
+            num_steps=value.shape[-1],
+        )
+
+        x1 = initial_loc + initial_diff
+        full_path = jnp.concatenate([initial_loc[:, None], x1[:, None], value], axis=-1)
+        previous_delta = full_path[:, 1:-1] - full_path[:, :-2]
+        transition_locs = full_path[:, 1:-1] + autoreg[:, None] * previous_delta
+        transition_probs = dist.Normal(transition_locs, scale[:, None]).log_prob(
+            full_path[:, 2:]
```
dylanhmorris, line 252:
```suggestion
            value
```

test/test_temporal_processes.py
```diff
@@ -54,25 +61,161 @@ def fixed_rw_kwargs(innovation_sd=0.05):
...
-        """Test shape and initial value handling with array initial values."""
+    @pytest.mark.parametrize("invalid_num_steps", [0, 1.5])
+    def test_num_steps_must_be_positive_integer(
+        self, distribution_cls, kwargs, invalid_num_steps
+    ):
+        """Constructors reject non-positive and non-integer step counts."""
+        with pytest.raises(ValueError, match="num_steps must be a positive integer"):
+            distribution_cls(**kwargs, num_steps=invalid_num_steps)
+
+    def test_state_differenced_ar1_single_step_sample_matches_initial_transition(self):
```
dylanhmorris: I think we can and should test the forward sampling returns the full post-initial state trajectories for all three new distributions and a variety of parameters.

test/test_temporal_processes.py
```diff
@@ -505,6 +646,362 @@ def test_repr_uses_random_variable_argument_names(self, process, expected):
...
+        """State-mode trace records ``_state``; innovation-mode ``_step`` is absent."""
+        rw = RandomWalk(**fixed_rw_kwargs(innovation_sd=0.1), parameterization="state")
+        traced = numpyro.handlers.trace(
+            numpyro.handlers.seed(rw.sample, rng_seed=0)
+        ).get_trace(n_timepoints=8, n_processes=2, name_prefix="rw")
+        assert "rw_state" in traced
+        assert "rw_step" not in traced
+
+    def test_state_site_contains_actual_post_initial_states(self):
+        """The ``_state`` site stores shifted states, not zero-origin offsets."""
```
dylanhmorris, line 895:
```suggestion
        """The ``_state`` site stores shifted states, not zero-origin offsets, and is therefore equivalent to the post initial values of the path returned by the ``TemporalProcess`"""
```

test/test_temporal_processes.py
```diff
@@ -505,6 +646,362 @@ def test_repr_uses_random_variable_argument_names(self, process, expected):
...
+    def test_state_site_shape(self):
+        """The state site holds the post-initial path of shape ``(n_processes, n_timepoints - 2)``."""
+        d = DifferencedAR1(**fixed_ar1_kwargs(), parameterization="state")
+        traced = numpyro.handlers.trace(
+            numpyro.handlers.seed(d.sample, rng_seed=0)
+        ).get_trace(n_timepoints=12, n_processes=3, name_prefix="diff")
+        assert traced["diff_state"]["value"].shape == (3, 10)
+
+    def test_second_row_equals_initial_value_plus_init_rate(self):
+        """``x[1]`` equals ``initial_value`` plus the unbundled initial difference."""
```
dylanhmorris, line 1116:
```suggestion
        """``x[1]`` equals ``initial_value`` plus the sampled initial difference."""
```

dylanhmorris (approved):
Thanks, @cdc-mitzimorris! A few suggestions for test readability and coverage but happy for you to merge without re-review.

PR #851, opened 2026-07-06
throttle dependabot PRs

cdc-mitzimorris:
- Groups all GitHub Actions bumps into one PR.
- Groups routine uv minor/patch updates into one PR.
- Leaves major updates as separate PRs, which is usually worth it because they deserve review.
- Delays fresh releases a bit with cooldown, reducing churn from rapid follow-up releases.
- Keeps pyproject.toml ranges stable with versioning-strategy: increase-if-necessary.
- Lowers the open Dependabot version-update PR cap to 3. GitHub’s default is 5

Security updates are handled differently: grouping can apply to security updates too, but cooldown does not apply to security PRs, and GitHub has a separate security PR limit.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,10 +5,34 @@ updates:
   - package-ecosystem: "github-actions"
     directory: "/"
     schedule:
-      # Check for updates to GitHub Actions every week
       interval: "weekly"
+      day: "monday"
+      time: "09:00"
+      timezone: "America/New_York"
+    groups:
+      github-actions:
+        patterns:
+          - "*"
+
   - package-ecosystem: "uv"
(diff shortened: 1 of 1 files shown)


cdc-mitzimorris:
@dylanhmorris  and @damonbayer - filing this because dependabot wanted to change pyproject.toml lower bounds as well as update uv lockfiles.
I think that the pyproject.toml  lower bounds should represent the minimum version PyRenew actually requires, not the newest version Dependabot resolved. The lockfile can move forward, but the package metadata should only tighten when our code depends on newer features/behavoirs.

dylanhmorris (approved):
Thanks @cdc-mitzimorris!

Issue #852, opened 2026-07-10
Add "RatioLinkedAscertainment" component


cdc-mitzimorris:
Add a new class "RatioLinkedAscertainment" which relates the ascertainment rate of two observation processes where the ascertainment rate for one process is the base rate and the ascertainment rate for the other is some ratio of the two signals.

This follows pyrenew-hew which parameterizes the ascertainment for the ED visits and hospitalizations as a base rate (IEDR) and a multiplier (IHR / IEDR).

This can be implemented straightforwardly using the base Ascertainment component:

```python
"""Linked ascertainment components."""

from __future__ import annotations

from collections.abc import Mapping

import jax
import numpyro
from pyrenew.ascertainment import AscertainmentModel
from pyrenew.metaclass import RandomVariable


class RatioLinkedAscertainment(AscertainmentModel):
    """Two ascertainment rates expressed as a base rate and a ratio."""

    def __init__(
        self,
        name: str,
        base_signal: str,
        linked_signal: str,
        base_rate_rv: RandomVariable,
        ratio_rv: RandomVariable,
    ) -> None:
        """Create a linked ascertainment model for two named signals."""
        super().__init__(name=name, signals=(base_signal, linked_signal))
        self.base_signal = base_signal
        self.linked_signal = linked_signal
        self.base_rate_rv = base_rate_rv
        self.ratio_rv = ratio_rv

    def sample(self, **kwargs: object) -> Mapping[str, jax.Array]:
        """Sample the base rate and ratio, then compute the linked rate."""
        base_rate = self.base_rate_rv()
        ratio = self.ratio_rv()
        linked_rate = base_rate * ratio
        numpyro.deterministic(f"{self.name}_{self.base_signal}", base_rate)
        numpyro.deterministic(f"{self.name}_{self.linked_signal}", linked_rate)
        return {self.base_signal: base_rate, self.linked_signal: linked_rate}
```

The corresponding Python code to specify a model using the PyrenewBuilder class is:

```python
iedr_rv = DistributionalVariable(name="iedr",  dist.LogNormal(0.0, jnp.log(jnp.sqrt(2.0))))

ratio_rv = DistributionalVariable(name="ihr_rel_iedr",  dist.LogNormal(0.0, jnp.log(jnp.sqrt(2.0))))

ascertainment = RatioLinkedAscertainment(
    name="he_ascertainment",
    base_signal="ed_visits",
    linked_signal="hospital",
    base_rate_rv=iedr_rv,
[...]

PR #863, opened 2026-07-15
 Component RatioLinkedAscertainment

cdc-mitzimorris:
implemented #852,  updated tutorial.


docs/tutorials/ascertainment.qmd
```diff
@@ -21,126 +21,138 @@ jupyter:
...
+
+The baseline rates $\boldsymbol{\alpha}_0$ set the prior location.
+The covariance matrix $\boldsymbol{\Sigma}$ controls the scale of variation in each logit-transformed rate and how strongly the rates vary together under the prior.
+
+For two signals, positive prior covariance means that prior draws with a higher-than-expected rate for one signal tend to have a higher-than-expected rate for the other.
+Negative covariance means that they tend to move in opposite directions.
+Zero covariance means that the multivariate normal prior does not couple their deviations from the baseline rates.
+
+The covariance structure is supplied to `JointAscertainment`; the component does not estimate an unknown covariance matrix from the observation data.
+It describes prior dependence, not a causal relationship between the signals.
```
dylanhmorris, line 147: One could infer $\Sigma$ rather than hard coding it, no? Just pass in a sampled quantity. I think simpler to say that `JointAscertainment` RVs treat $\Sigma$ as fixed; to infer $\Sigma$ you'd need to sample it upstream.

docs/tutorials/ascertainment.qmd
```diff
@@ -21,126 +21,138 @@ jupyter:
...
 ```
 
-Ascertainment is the probability that a latent infection appears in an observed signal, e.g., a hospitalization or visit to the emergency department.
-For hospital admissions this probability is often called an infection-hospitalization rate (IHR).
-For emergency department visits it might be called an infection-ED-visit rate (IEDR).
+## Overview
 
-PyRenew count observations accept an `ascertainment_rate_rv`.
-For simple models this can be any ordinary `RandomVariable`.
+An ascertainment rate is the probability that an infection results in an observed event.
```
dylanhmorris:
```suggestion
An ascertainment rate is the rate at which infections lead to observed events.
```

docs/tutorials/ascertainment.qmd
```diff
@@ -21,126 +21,138 @@ jupyter:
...
+For hospital admissions, this probability is often called the infection-hospitalization rate (IHR).
```
dylanhmorris:
```suggestion
For hospital admissions, this is often called the infection-hospitalization rate (IHR).
```

docs/tutorials/ascertainment.qmd
```diff
@@ -154,103 +166,83 @@ joint_ascertainment = JointAscertainment(
...
-    )
-    + p9.geom_abline(slope=1, intercept=jnp.log10(2), linetype="dashed", color="gray")
-    + p9.scale_y_log10()
-    + p9.scale_x_log10()
-    + theme_tutorial
-)
-```
+The ratio may be any positive value.
+Unlike `JointAscertainment`, this component does not automatically constrain the derived linked rate to the unit interval.
+Users must choose priors that make $\alpha_{\mathrm{base}}\rho > 1$ negligibly likely.
```
dylanhmorris: You can imagine certain observables with more than one ascertained event per infection, so I don't think all ascertainment rates need to be probabilities. I think we can say something like

```suggestion
If both rates should be $<1$ (for instance because both are probabilities), make sure to choose priors that make $\alpha_{\mathrm{base}}\rho > 1$ negligibly likely.
```

test/test_ascertainment.py
```diff
@@ -288,6 +292,125 @@ def test_signal_accessor_requires_active_context(self):
...
+        assert ascertainment.signals == ("ed_visits", "hospital")
+        assert ascertainment.base_signal == "ed_visits"
+        assert ascertainment.linked_signal == "hospital"
+        assert ascertainment.base_rate_rv is base_rate_rv
+        assert ascertainment.ratio_rv is ratio_rv
+
+        base_accessor = ascertainment.for_signal("ed_visits")
+        linked_accessor = ascertainment.for_signal("hospital")
+        assert base_accessor.signal_name == "ed_visits"
+        assert linked_accessor.signal_name == "hospital"
```
dylanhmorris, line 359: also assert the expected type and the expected `ascertainment_name`?

test/test_ascertainment.py
```diff
@@ -288,6 +292,125 @@ def test_signal_accessor_requires_active_context(self):
...
+        ascertainment = RatioLinkedAscertainment(
+            name="he_ascertainment",
+            base_signal="ed_visits",
+            linked_signal="hospital",
+            base_rate_rv=DeterministicVariable("iedr", jnp.array(0.2)),
+            ratio_rv=DeterministicVariable("ihr_rel_iedr", jnp.array(0.5)),
+        )
+
+        with numpyro.handlers.trace() as trace:
+            values = ascertainment.sample(unused_context_value=True)
```
dylanhmorris: why `unused_context_value=True`?

test/test_ascertainment.py
```diff
@@ -288,6 +292,125 @@ def test_signal_accessor_requires_active_context(self):
...
+    def test_sample_returns_base_rate_and_scaled_linked_rate(self):
+        """Test that the linked rate is the base rate times the ratio."""
+        ascertainment = RatioLinkedAscertainment(
+            name="he_ascertainment",
+            base_signal="ed_visits",
+            linked_signal="hospital",
+            base_rate_rv=DeterministicVariable("iedr", jnp.array(0.2)),
+            ratio_rv=DeterministicVariable("ihr_rel_iedr", jnp.array(0.5)),
+        )
+
```
dylanhmorris, line 377: Does this not need a seed? If it does, should we be concerned that the tests are passing?

test/test_ascertainment.py
```diff
@@ -288,6 +292,125 @@ def test_signal_accessor_requires_active_context(self):
...
+            values["ed_visits"],
+            trace["he_ascertainment_ed_visits"]["value"],
+        )
+        assert jnp.array_equal(
+            values["hospital"],
+            trace["he_ascertainment_hospital"]["value"],
+        )
+
+    def test_sample_calls_each_random_variable_once(self):
+        """Test that one base-rate site and one ratio site are sampled."""
```
dylanhmorris:
```suggestion
        """Test that the base-rate site and the ratio site are sampled and converted into user-exposed, signal specific deterministic sites"""
```

dylanhmorris (changes requested):
Thanks @cdc-mitzimorris. A few minor changes requested.

docs/tutorials/ascertainment.qmd
cdc-mitzimorris: good point. simplified.

test/test_ascertainment.py
cdc-mitzimorris: added.

test/test_ascertainment.py
cdc-mitzimorris: good catch.  not needed - removed.

test/test_ascertainment.py
cdc-mitzimorris: these are deterministic values - i.e., fixed - this test is for the returned signal mapping.   the next test is for stochastic sites creation.

cdc-mitzimorris:
> Thanks @cdc-mitzimorris. A few minor changes requested.

@dylanhmorris all requested changes made (and questions answered).

dylanhmorris (approved):

dylanhmorris:
Thanks, @cdc-mitzimorris!

PR #888, opened 2026-09-15
Allow posterior prediction to extend state-centered temporal processes beyond the fitted time horizon.

cdc-mitzimorris:
This PR lets users can select either parameterization and get the same forecast-horizon behavior:

- Posterior prediction at the fitted horizon replays the fitted trajectory.
- Posterior prediction at a longer horizon replays the fitted trajectory and samples the missing suffix.

Changes made:

- Converted state branches of RandomWalk, AR1, and DifferencedAR1 to NumPyro scan.
- _state sites now use (time, process) layout as innovation processes.
- Added posterior-extension and terminal-carry tests.
- Added forecasting parity tests for state and innovation parameterizations.

Files changed:

- pyrenew/latent/temporal_processes.py
- test/test_temporal_processes.py
- latent/init.py (pyrenew/latent/__init__.py)


pyrenew/latent/__init__.py
```diff
@@ -28,14 +28,16 @@
 from pyrenew.latent.infections import Infections
 from pyrenew.latent.infectionswithfeedback import InfectionsWithFeedback
 from pyrenew.latent.population_infections import PopulationInfections
+from pyrenew.latent.state_centered_distributions import (
```
dylanhmorris: If I understand the point of this PR correctly, we should not be user-exposing the distributions in place of the temporal processes. If tests would fail otherwise, need to revise.

pyrenew/latent/temporal_processes.py
dylanhmorris, line 1: I think we either
1. should get rid of the distribution objects altogether (since they're no longer used here) or 
2. should introduce forecast/posterior prediction-friendly implementations directly to the distributions' forward `sample()` methods.

I favor 2. What do you think @cdc-mitzimorris?

dylanhmorris (changes requested):
A couple key things.

pyrenew/latent/__init__.py
cdc-mitzimorris: good catch - removed.

AGENTS.md
```diff
@@ -0,0 +1,17 @@
+Functions, classes, and packages designs should be straightforward.
```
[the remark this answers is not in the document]
cdc-mitzimorris: There is a separate PR for AGENTS.md.

pyrenew/latent/temporal_processes.py
cdc-mitzimorris: they have been deleted.

cdc-mitzimorris:
> A couple key things.

@dylanhmorris  - all requested changes made.

test/test_temporal_processes.py
```diff
@@ -1186,6 +973,208 @@ def model_innov():
...
+        fitted_n_timepoints = 5
+        forecast_n_timepoints = 8
+        state_offset = 2 if process_name == "differenced_ar1" else 1
+        fitted_scanned_timepoints = fitted_n_timepoints - state_offset
+        forecast_scanned_timepoints = forecast_n_timepoints - state_offset
+        fitted_states = jnp.arange(
+            n_draws * fitted_scanned_timepoints * n_processes,
+            dtype=jnp.asarray(0.2).dtype,
+        ).reshape(n_draws, fitted_scanned_timepoints, n_processes)
+        fitted_states /= 10.0
```
dylanhmorris: Why?

test/test_temporal_processes.py
```diff
@@ -1186,6 +973,208 @@ def model_innov():
...
+            return_sites=("trajectory", "forecast_state"),
+        )
+        key = jax.random.PRNGKey(20)
+        forecast_a = predictive_a(key, forecast_n_timepoints)
+        forecast_b = predictive_b(key, forecast_n_timepoints)
+
+        observed_difference = (
+            forecast_b["forecast_state"][:, fitted_scanned_timepoints, :]
+            - forecast_a["forecast_state"][:, fitted_scanned_timepoints, :]
+        )
```
dylanhmorris, line 1098: add an assertion that they do in fact have different transition means?

dylanhmorris (approved):
Thanks @cdc-mitzimorris. Please reconcile merge conflicts with main so there's no `AGENTS.md` in this diff.

Longer term we may wish to think about a full refactor of how we handle processes (as PyRenew base RVs vs NumPyro distributions vs TemporalProcess wrappers of PyRenew base RVs. But I think we should land this to unblock you.

test/test_temporal_processes.py
cdc-mitzimorris: good catch - no reason for this - removing.

test/test_temporal_processes.py
cdc-mitzimorris: added

